PR #3, opened 2018-04-25
Vendored ccm

nckturner:
Creates aws ccm binary, with the cloud provider specific code vendored in-tree code.  It should eventually migrate to this repository, but this will help us get started with e2e testing.


gnufied:
Discussed this in meeting today. Lets merge this for now.

/lgtm

cmd/aws-cloud-controller-manager/main.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	"k8s.io/apiserver/pkg/server/healthz"
+	"k8s.io/apiserver/pkg/util/logs"
+	"k8s.io/kubernetes/cmd/cloud-controller-manager/app"
+	"k8s.io/kubernetes/cmd/cloud-controller-manager/app/options"
+	_ "k8s.io/kubernetes/pkg/client/metrics/prometheus" // for client metric registration
+	_ "k8s.io/kubernetes/pkg/cloudprovider/providers/aws"
```
andrewsykim: This is how all out-of-tree providers are doing it now, but in the near future, this import path may change depending on [this proposal](https://docs.google.com/document/d/1YqsPzeG10E59vea5StKFM6W_xZq034eTaf1n649JQvs/edit).

nckturner: Yup, this was just meant to get this repository building a binary, and keep the source of truth in-tree.  Once the shared, generic cloudprovider code moves to staging, these imports will reflect that.  And once the aws code moves into *this repository*, these imports will reflect that too.

justinsb:
We also discussed this in sig-aws on Friday (June 1st 2018).  Consensus was to proceed with this PR, in that it allows progress and doesn't really hurt us.  It might be possible to do more in-tree with bazel restrictions etc, but that shouldn't stop us proceeding down this path also.

jsafrane:
Do we have merge bot running here?

jsafrane:
We do :-)

PR #11, opened 2018-10-04
Copy latest version of in-tree provider

mcrute:
This PR copies the latest version of the in-tree cloud provider code for AWS into this repository. The code here is unmodified and identical to the in-tree code as of commit 1ce5f67af0.

This is the first change in a series of changes updating this repository and getting the code into a working state. I'm breaking the initial set of changes into 3 commits to make it easier to review: copy in-tree code, update dependencies, and build fixes.

**Release note**:
```release-note
NONE
```
/kind feature
/cc @dims @cheftako @d-nishi

### DIFF
--- a/pkg/cloudprovider/providers/aws/BUILD
+++ b/pkg/cloudprovider/providers/aws/BUILD
@@ -1,0 +1,105 @@
+package(default_visibility = ["//visibility:public"])
+
+load(
+    "@io_bazel_rules_go//go:def.bzl",
+    "go_library",
+    "go_test",
+)
+
+go_library(
+    name = "go_default_library",
+    srcs = [
+        "aws.go",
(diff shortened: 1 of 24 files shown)


justinsb:
So this is a different strategy than what I think @cheftako is pursuing for GCE.  That's not necessarily a bad thing - we can try multiple approaches.  My understanding is that @cheftako is planning on moving the GCE code into staging, and extracting it from there.  The act of moving it should show whether there are any lingering dependencies.  Copying it is a little easier I guess, and should give us some ideas of lingering dependencies, but this does mean all changes will have to be made in two places (until we remove the in-tree code).

mcrute:
I'm going to close these for now and work on an alternative approach. Will open other PRs later.

PR #18, opened 2018-11-15
Update package dependencies

mcrute:
This updates the vendored dependencies as a result of adding the cloud provider code from #17. This is a result of a `dep ensure -update`. The only file that might be interesting here is [Gopkg.toml](https://github.com/kubernetes/cloud-provider-aws/pull/18/files#diff-836546cc53507f6b2d581088903b1785) which was updated to newer versions and some unused packages removed.

/cc @d-nishi
/milestone v1.13
/kind feature
/sig aws
/sig cloud-provider


gyuho:
LGTM. Defer to @d-nishi. Thanks!

PR #19, opened 2018-11-15
Update cloud provider

mcrute:
This change updates the driver application so the cloud provider can function out of tree. It also adds a Makefile to build and test the code as well as the appropriate supporting scripts.

/cc @d-nishi
/milestone v1.13
/kind feature
/sig aws
/sig cloud-provider

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/aws-cloud-controller-manager
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,53 @@
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


d-nishi:
cc @gyuho

gyuho:
/lgtm

Looks great!

Non-related question. Have we considered klog? Seems like upstream is migrating from glog https://github.com/kubernetes/kubernetes/pull/70889.

mcrute:
> Non-related question. Have we considered klog? Seems like upstream is migrating from glog [kubernetes/kubernetes#70889](https://github.com/kubernetes/kubernetes/pull/70889).

I did consider that and it's something I want to clean up. I just opened #20 to address this. Thanks!

PR #31, opened 2019-02-18
Fix build

leakingtapan:
Fix the build and update readme to redirect request for EBS volume plugin

```release-note
None
```


leakingtapan:
@mcrute thx for the quick lgtm

PR #43, opened 2019-06-28
use the correct cloudprovider package

amdonov:
/kind api-change

There are two implementations of the cloudprovider API in the current code. The package registers with one and the k8s command uses the other and fails to find the aws cloud-provider. This update removes the second implementation.


nckturner:
/close
This is no longer needed because we are importing legacy (any change to aws.go should be made there, for now).

PR #45, opened 2019-07-10
WIP/Proposal - Cluster name-prefixed load balancers

Preskton:
### What type of PR is this?

/kind api-change

... since this changes the contract for how we name load balancers when created by the cloud provider.

### What this PR does / why we need it

As an operator, I want the load balancer names that are created via Kubernetes to have semi-predictable names so that I can do things like limit the cluster's service principal role IAM to only affect resources prefixed with the cluster name.

This also replaces a method that is now deprecated in the base cloud provider implementation, `DefaultLoadBalancerName`, ref: https://github.com/kubernetes/cloud-provider/blob/02ef7372f6c090f5a90430f1b178b38f1fba2b1e/cloud.go#L82.

If anything goes badly, we fall back to the prior implementation (`a` + the service's UID).

### Which issue(s) this PR fixes

No issue has been specifically raised for this. Should I do this as part of the process?

### Special notes for your reviewer

This is my first PR into `cloud-provider-aws`. I'm not even sure that I've submitted it to the right place. Happy to adjust how I've done this, test coverage, etc. - feedback much appreciated! :)

I'll likely add some additional coverage around the edge cases of:

- greater than 32 chars
- special characters - although my guess this is not possible based on validation that's already happening upstream in AWS on cluster name and k8s on service name.

### Does this PR introduce a user-facing change?

```release-note
Load Balancers: When creating a load balancer for a `Service` of type `LoadBalancer`, the name of the new load balancer will now follow the format: `$clusterName-$serviceName`, limited to 32 characters.
```

### DIFF
--- a/pkg/cloudprovider/providers/aws/aws.go
+++ b/pkg/cloudprovider/providers/aws/aws.go
@@ -3666,18 +3666,10 @@ func (c *Cloud) GetLoadBalancer(ctx context.Context, clusterName string, service
 func (c *Cloud) GetLoadBalancerName(ctx context.Context, clusterName string, service *v1.Service) string {
 	// Ref: https://docs.aws.amazon.com/elasticloadbalancing/2012-06-01/APIReference/API_CreateLoadBalancer.html
 	// unique for the region / max 32 chars / alphanumeric plus hyphen / can't start or end with hyphen
+	reg := regexp.MustCompile(`[^a-zA-Z0-9\-]+`)
 
-	ret := "a" + string(service.UID) // This is the old implementation from cloudbalancer.DefaultLoadBalancerName
-
-	reg, err := regexp.Compile(`[^a-zA-Z0-9\-]+`)
-
-	if err == nil {
-		ret = fmt.Sprintf("%s-%s", clusterName, string(service.UID))
-		ret = reg.ReplaceAllString(ret, "")
-	} else {
-		klog.Warning("LoadBalancerName regexp failed to compile; falling back to previous implementation")
-		ret = strings.Replace(ret, "-", "", -1)
-	}
+	ret := fmt.Sprintf("%s-%s", clusterName, string(service.UID))
+	ret = reg.ReplaceAllString(ret, "")
 
 	if len(ret) > 32 {
 		ret = ret[:32]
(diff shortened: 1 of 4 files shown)


Preskton:
The 32 character limit restriction is pretty intense: someone could pretty easily reduce the set of load balancers they could create by creating a cluster with a fairly long name.

Would this make more sense if we tagged the xLB in AWS with `kubernetes.io/cluster` with a value of the cluster name? I think this would serve my purpose of being able to restrict IAM with `iam:ResourceTag` filters.

pkg/cloudprovider/providers/aws/aws.go
```diff
@@ -45,7 +46,7 @@ import (
 	gcfg "gopkg.in/gcfg.v1"
 	"k8s.io/klog"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
jaypipes: technically, this isn't necessary. the package name is already "v1". but, it doesn't hurt anything I suppose :)

pkg/cloudprovider/providers/aws/aws.go
```diff
@@ -3663,8 +3664,18 @@ func (c *Cloud) GetLoadBalancer(ctx context.Context, clusterName string, service
 
 // GetLoadBalancerName is an implementation of LoadBalancer.GetLoadBalancerName
 func (c *Cloud) GetLoadBalancerName(ctx context.Context, clusterName string, service *v1.Service) string {
-	// TODO: replace DefaultLoadBalancerName to generate more meaningful loadbalancer names.
-	return cloudprovider.DefaultLoadBalancerName(service)
+	// Ref: https://docs.aws.amazon.com/elasticloadbalancing/2012-06-01/APIReference/API_CreateLoadBalancer.html
+	// unique for the region / max 32 chars / alphanumeric plus hyphen / can't start or end with hyphen
+	reg := regexp.MustCompile(`[^a-zA-Z0-9\-]+`)
```
jaypipes: That regex actually would allow a hyphen at the start or end of the name.

pkg/cloudprovider/providers/aws/aws.go
```diff
@@ -3663,8 +3664,18 @@ func (c *Cloud) GetLoadBalancer(ctx context.Context, clusterName string, service
...
+
+	ret := fmt.Sprintf("%s-%s", clusterName, string(service.UID))
+	ret = reg.ReplaceAllString(ret, "")
+
+	if len(ret) > 32 {
+		ret = ret[:32]
```
jaypipes: Any cluster with a name 31 characters long will result in the ELB name for all LoadBalancers in the cluster being the same. I'm pretty sure that's not what you want, right?

Preskton: Yup, unfortunately, that's the char limit for LB names. :(

pkg/cloudprovider/providers/aws/aws.go
Preskton: Must be my import settings - will clean up.

PR #56, opened 2019-09-12
Provides new service annotation for load balancer name.

### LINKED ISSUE #55: Allow load balancer name to be provided via service annotation
Would like to add an additional annotation type for the load balancer name.

**Why is this needed**:
Currently balancer can only be created with a machine-generated name making it very hard to distinguish in AWS if multiple balancers have been created within an account.

Additionally the DefaultLoadBalancerName from the cloud-provider package which is currently in use here has been deprecated and needs a replacement.

/kind feature

mrf:
**What type of PR is this?**
/kind feature

Provides a configurable load balancer name via a service annotation and replaces the [deprecated](https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L77) DefaultLoadBalancer call with provider-specific logic for creating a default name.

Fixes #55 

```release-note
New aws-load-balancer-name annotation to provide a unique name for a load balancer.
```


selslack:
@micahhausler @jsafrane this looks useful. Can we have it moving towards the release, please?

PR #59, opened 2019-09-25
add ec2:DescribeAccountAttributes needed when creating a service

janitha09:
**What type of PR is this?**
> /kind documentation

Documents a permission needed to create an elastic load balancer for a service

Fixes #

NONE
```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ For the `aws-cloud-controller-manager` to be able to communicate to AWS APIs, yo
         "autoscaling:DescribeAutoScalingGroups",
         "autoscaling:DescribeLaunchConfigurations",
         "autoscaling:DescribeTags",
+        "ec2:DescribeAccountAttributes",
         "ec2:DescribeInstances",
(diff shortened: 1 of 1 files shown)


nckturner:
@janitha09 So I had to do some digging here, this is a bit of a corner case.  This permissions is only needed if an ELB has never been created in the account before.  There are three possible ways to get passed this error: 1. add permissions to the EKS role like you suggest, 2. create an ELB in the console, or 3. create the ELB SLR explicitly (all three solutions create the ELB SLR in some way).  

I think this should be documented more thoroughly. Additionally, I found a reference online that suggested you need the `iam:CreateServiceLinkedRole` permission as well--did you test this?

randomvariable:
Do indeed need `iam:CreateServiceLinkedRole`. we added that in Cluster API here:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/commit/6b1875be6c162e1969bc5ba10d629b742ac70ef1

PR #61, opened 2019-10-02
Refactor to use legacy cloud provider

nckturner:
* This is starting with v1.15.0
* This is important because we need to start by using legacy-cloud-provider so that we don't fork this repo from in-tree before people migrate.  That approach would make migration much more difficult.

**What type of PR is this?**

> /kind cleanup

In order to start using this repo to build valid cloud-controller-manager binaries and begin adding tests, we need to migrate to the legacy-cloud-provider.  

```release-note
NONE
```


nckturner:
Followed openstack's approach, which involves defining the cobra command, and additionally redefining the `initFunc`s for each cloud controller.

cmd/aws-cloud-controller-manager/main.go
```diff
@@ -26,54 +26,67 @@ limitations under the License.
...
-for a Kubernetes cluster.`,
+		Use:  "aws-cloud-controller-manager",
+		Long: `aws-cloud-controller-manager manages AWS cloud resources for a Kubernetes cluster.`,
 		Run: func(cmd *cobra.Command, args []string) {
-			verflag.PrintAndExitIfRequested()
+
+			// Use our version instead of the Kubernetes formatted version
+			versionFlag := cmd.Flags().Lookup("version")
+			if versionFlag.Value.String() == "true" {
+				klog.V(1).Infof("%s version: %s", cmd.Name(), version)
```
andrewsykim: exit after printing version?

cmd/aws-cloud-controller-manager/main.go
```diff
@@ -87,22 +100,126 @@ for a Kubernetes cluster.`,
...
+func newControllerInitializers() map[string]initFunc {
+	controllers := map[string]initFunc{}
+	controllers["cloud-node"] = startCloudNodeController
+	controllers["cloud-node-lifecycle"] = startCloudNodeLifecycleController
+	controllers["service"] = startServiceController
+	controllers["route"] = startRouteController
+	return controllers
+}
+
+func startCloudNodeController(ctx *cloudcontrollerconfig.CompletedConfig, cloud cloudprovider.Interface, stopCh <-chan struct{}) (http.Handler, bool, error) {
```
andrewsykim: If you call [NewCloudControllerManagerCommand()](https://github.com/kubernetes/kubernetes/blob/67d928acdc351d3d4f9e20a92cee4452a26ed0c4/cmd/cloud-controller-manager/app/controllermanager.go#L58) in main.go, you should get all of this with it.

nckturner: Yeah I saw that.  It would be easier but my plan was to start with this and then eventually replace/remove the dependencies on k/k, which I think will be easier if we have one level deeper.

geckofu:
Hi, I'm a newcomer and would like to contribute to this project. Where should I work on after this migration? In [kubernetes' main repo staging area](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers) as the [legacy-cloud-provider](https://github.com/kubernetes/legacy-cloud-providers) says?

M00nF1sh:
@geckofu 
For now, it's still kubernetes's main repo's staging area(which will be synced to legacy-cloud-provider and consumed by this project).
In the future, we'll gradually migrate off legacy-cloud-provider dependency.

kapilt:
i feel like the better approach would have been to go in the opposite direction, ie have the legacy cloud in k8s main repo depend on the code in this repo. as is it just erases most of the progress made in moving the provider out of tree.

andrewsykim:
@kapilt that's a good point but there's a few reasons for not importing this repo back from kubernetes/kubernetes:
* We have less control over the dependencies that are pulled into this repo that would eventually get pulled into all the core Kubernetes components. The existing legacy repo is already contained and we ensure that dep graph doesn't grow. 
* It'll be harder to keep backwards compatibility if we are pulling in new changes here to core Kubernetes
* It's much easier to iterate this repo from what already exists because the legacy repo passes Kubernetes pre-checks in the main repo and we know have more confidence that what we are supporting out-of-tree should work for most users transitioning from in-tree.

PR #70, opened 2020-01-07
add dockerfile for containerization

mattlqx:
**What type of PR is this?**
 /kind feature

Containerizes the aws-cloud-controller-manager binary so that it can be run from Docker. Mostly a copy from [cloud-provider-vsphere](https://github.com/kubernetes/cloud-provider-vsphere/blob/master/cluster/images/controller-manager/Dockerfile).

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,46 @@
+# Copyright 2018 The Kubernetes Authors.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+################################################################################
+##                               BUILD ARGS                                   ##
+################################################################################
+# This build arg allows the specification of a custom Golang image.
+ARG GOLANG_IMAGE=golang:1.13.5
+
+################################################################################
+##                              BUILD STAGE                                   ##
+################################################################################
+# Build the manager as a statically compiled binary so it has no dependencies
+# libc, muscl, etc.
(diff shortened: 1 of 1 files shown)


mattlqx:
I signed it

Dockerfile
```diff
@@ -0,0 +1,54 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
```
leakingtapan: Let's use k8s license. Example see: https://github.com/kubernetes/cloud-provider-aws/blob/master/cmd/aws-cloud-controller-manager/main.go#L2

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+################################################################################
+##                               BUILD ARGS                                   ##
+################################################################################
+# This build arg allows the specification of a custom Golang image.
+ARG GOLANG_IMAGE=golang:1.12.10
```
leakingtapan: NIT: use golang 1.13

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+
+# The distroless image on which the CPI manager image is built.
+#
+# Please do not use "latest". Explicit tags should be used to provide
+# deterministic builds. This image doesn't have semantic version tags, but
+# the fully-qualified image can be obtained by entering
+# "gcr.io/distroless/static:latest" in a browser and then copying the
+# fully-qualified image from the web page.
+ARG DISTROLESS_IMAGE=gcr.io/distroless/static@sha256:9b60270ec0991bc4f14bda475e8cae75594d8197d0ae58576ace84694aa75d7a
```
leakingtapan: Is it possible to scratch image?

Dockerfile
mattlqx: Added.

Dockerfile
mattlqx: Done!

Dockerfile
mattlqx: Looks like it as long as there's a /tmp.

PR #71, opened 2020-01-08
change dockerfile to distroless, use make

mattlqx:
found a few issues in my earlier pull #70 .

1. theres a Makefile in this project, so we should use it to build. the version variable usage in the ld flags was taken as a literal due to single quotes. 
2. for some reason, there is an issue with the AWS credentials when using the scratch image. using the distroless image works fine. i tried creating a bunch of paths and copying files to see which one being missing was causing the issue but i was unable to figure it out.

heres the error that occurs when using scratch on startup.
```
I0108 21:51:36.763010       1 aws.go:1112] Using AWS assumed role arn:aws:iam::XXX:role/kube
I0108 21:51:36.763042       1 aws.go:1183] Building AWS cloudprovider
I0108 21:51:36.763110       1 aws.go:1149] Zone not specified in configuration file; querying AWS metadata service
F0108 21:51:37.679349       1 controllermanager.go:121] Cloud provider could not be initialized: could not init cloud provider "aws": error finding instance i-XXXXXXXXXXXX: "error listing AWS instances: \"NoCredentialProviders: no valid providers in chain. Deprecated.\\n\\tFor verbose messaging see aws.Config.CredentialsChainVerboseErrors\""
```

/kind bug

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,15 @@
 # This build arg allows the specification of a custom Golang image.
 ARG GOLANG_IMAGE=golang:1.13.5
 
+# The distroless image on which the CPI manager image is built.
+#
+# Please do not use "latest". Explicit tags should be used to provide
+# deterministic builds. This image doesn't have semantic version tags, but
+# the fully-qualified image can be obtained by entering
+# "gcr.io/distroless/static:latest" in a browser and then copying the
+# fully-qualified image from the web page.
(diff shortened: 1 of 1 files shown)


leakingtapan:
Interesting. Although it’s good to know why it failed with scratch. I’m fine to use diatribes for now.

/lgtm
/approve

Issue #42, opened 2019-06-25
Status of project and documentation


darwin67:
Following the announcement of `v1.15.0`, my understanding is that the future development of cloud providers are being moved out of the kubernetes core.

Existing core including AWS has been moved to https://github.com/kubernetes/legacy-cloud-providers.
Which pretty much breaks the link in [note][broken-link] and also seems like in-tree providers will be completely removed in the next couple releases.
We're currently on 1.14.3 and have tested 1.15.0 and can verify that the in-tree provider still works.

But the point I wanted to make is, this repo doesn't seem to have much activity compared to the other major cloud providers.
You also don't have any sample manifests for deploying the `cloud-controller-manager` and I haven't had any luck getting the aws provider to work as an external cloud provider ever since I first attempted at 1.13..
Asking on kubernetes slack also haven't work well so far, which I suspect either no one had got this working, or simply no one cares.
Doing a google search and reading docs of the CCM and other cloud providers also got me no where close to a working example of this external provider.

So these are my requests:

1. Please do clarify on the status, or any plans on work with this project.
2. Please update documentation and provide a close to working sample manifest to deploy this provider

Activity in this repo for the past 6+ months had mostly been cosmetic changes and I haven't seen anything related to feature updates, bug fixes or even documentations.

Please do forgive me if my tone is offending and I might just be paranoid for no good reason, but if you can at least clarify 1 for me, that can help me decide how to approach in the future. Either fork this repo and work on it on my own, or some other paths.

FYI:
We don't use EKS and have no plans on doing so in the future.

Thanks!

/triage support

[broken-link]: https://github.com/kubernetes/cloud-provider-aws#note

selslack:
This repo is kinda dead, yea.

particledecay:
How fitting that @selslack comments that this repo is kinda dead, and the next comment is the bot adding a stale label lol...

Is there any status on this? I've had the same trouble as @darwin67 with getting the external cloud provider working on k8s 1.17, since in-tree is deprecated now. The Azure and OpenStack cloud provider repos actually have documentation on getting those working, but nothing for this one.

Is there anyone out there that's gotten this project working as an external cloud provider in recent versions of Kubernetes?

brookssw:
[...]

PR #111, opened 2020-06-25
Vendor aws

sargun:
**What type of PR is this?**
/kind cleanup

It makes it so all the code is in one place.

This addresses some of the concerns in https://github.com/kubernetes/cloud-provider-aws/issues/42#issuecomment-635617333. 

NONE


nckturner:
I appreciate this, but we're not ready to move the cloud provider code from legacy-cloud-providers quite yet.  When we do, we will stop accepting commits there, but we only want one source of truth for now.

sargun:
@nckturner I'm curious why, and when? Depending on the answer to this, we will consider forking, or waiting for the change. Right now, it's fairly difficult for us to manage the current process.

nckturner:
@sargun That's fair.  ASAP, ideally.  Maybe we can get your changes merged upstream into k/k, or look into @andrewsykim's V2 proposal.  Thirdly, you're right, we could copy the code over here.  I'd like to have the timeline concrete if we do that, and know exactly when we are deprecating the in-tree provider.  I'll bring it up tomorrow in the provider-aws meeting.

PR #114, opened 2020-07-23
add initial experimental v2 AWS cloud provider

andrewsykim:
Signed-off-by: Andrew Sy Kim <kim.andrewsy@gmail.com>

**What type of PR is this?**
/kind feature

Adds an initial implementation of an experimental v2 provider that is exclusively out-of-tree. The actual implementation is incomplete, this just adds the initial boilerplate code for the cloud provider interface. Future PRs will incrementally implement the entire interface.

There are many requested features about the existing AWS cloud provider that are difficult to implement since we can't break behavior against existing clusters. Some of these include but are not limited to:
* allowing node names that are not the ec2 private DNS
* better naming of ELB/NLB resources

In addition, the existing provider is catered towards the deprecated in-tree architecture where it relies on the kubelet, kube-controller-manager and kube-apiserver to be AWS aware. As we approach a future of only out-of-tree providers, starting fresh on a v2 provider under these new assumptions may be the right direction. The v2 provider would be a breaking change from v1, meaning existing clusters shouldn't migrate to the v2 provider. The v2 provider will only support new clusters that enable the out-of-tree cloud provider (i.e. setting --cloud-provider=external on kube components). The v1 provider is **NOT** being deprecated and will continue to be supported in this repo. However, a future version of Kubernetes will remove the in-tree provider from the core Kubernetes release. 

```release-note
add initial experimental v2 AWS cloud provider
```

### DIFF
--- a/cmd/aws-cloud-controller-manager/main.go
+++ b/cmd/aws-cloud-controller-manager/main.go
@@ -48,14 +48,20 @@ import (
 	"k8s.io/kubernetes/cmd/cloud-controller-manager/app/options"
 	"k8s.io/kubernetes/pkg/features" // add the kubernetes feature gates
 	utilflag "k8s.io/kubernetes/pkg/util/flag"
-	"k8s.io/legacy-cloud-providers/aws"
 	netutils "k8s.io/utils/net"
 
 	cloudprovider "k8s.io/cloud-provider"
+	awsv1 "k8s.io/cloud-provider-aws/pkg/providers/v1"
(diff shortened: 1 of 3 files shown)


andrewsykim:
/hold

Requires discussion

justinsb:
We discussed this in the aws-provider biweekly ; there are fragmentation risks but I agree that they are outweighed by the unblocking nature of a v2.  Then we can see if the changes need to be breaking (and are worth the breakage!)

/lgtm

Issue #125, opened 2020-09-15
Tracking issue for v2 aws provider


andrewsykim:
**What would you like to be added**:
In https://github.com/kubernetes/cloud-provider-aws/pull/114 we added the initial boilerplate code to enable a v2 AWS cloud provider. This is a tracking issue for future PRs that add to the implementation. 

**Why is this needed**:
This project will continue to support the existing AWS provider in "out-of-tree" mode. However, there are many known issues and gaps in the current provider implementation that is not trivial to address. A new v2 implementation will try to address these limitations. The v2 implementation is a breaking change from v1 and is only supported for new clusters. Existing clusters cannot upgrade to v1. 

- [ ] Instances (@andrewsykim)  
 - [ ] Zones
- [ ] LoadBalancer (@HaibaraAi96) 
- [ ] Routes

/kind feature

andrewsykim:
Some issues we should consider as we implement v2:
https://github.com/kubernetes/cloud-provider-aws/issues/124
https://github.com/kubernetes/cloud-provider-aws/issues/63
https://github.com/kubernetes/cloud-provider-aws/issues/55

HaibaraAi96:
I'd love to work on the loadbalancer implementation 😃

andrewsykim:
Sounds good @HaibaraAi96! I updated the tracking issue to break up the work here a bit and assigned the loadbalancer implementation to you. Feel free to use the legacy implementation as a reference https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/legacy-cloud-providers/aws

rfranzke:
> The v2 implementation is a breaking change from v1 and is only supported for new clusters. Existing clusters cannot upgrade to v1.

What does this mean exactly, i.e., how will existing clusters be upgraded in the future (assuming the legacy AWS cloud provider code is being removed from k/k)?

PR #128, opened 2020-09-23
loadbalancer[1]: add initial implementation of LoadBalancer interface

nicolehanjing:
**What type of PR is this?**
/kind feature

This is an initial PR of implementing the load balancer. 
implemented several functions:
- GetLoadBalancer
- GetLoadBalancerName
- UpdateLoadBalancer
- EnsureLoadBalancerDeleted

some TODOs for follow-up PRs:
- implement `EnsureLoadBalancer` function
- handle cluster tags
- external LB/ NLB
- `GetLoadBalancerName` may need improvements

Part of #125

```release-note
Add initial implementation of loadbalancer for the v2 provider.
```


pkg/providers/v2/loadbalancer.go
```diff
@@ -0,0 +1,342 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/elb"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	v1 "k8s.io/api/core/v1"
+	cloudprovider "k8s.io/cloud-provider"
+	"k8s.io/klog/v2"
+)
+
+// newLoadBalancer returns an implementation of cloudprovider.LoadBalancer
+func newLoadBalancer(region string, creds *credentials.Credentials) (cloudprovider.LoadBalancer, error) {
```
HaibaraAi96, line 36: @andrewsykim I'm just following the pattern for implementing the LoadBalancer interface
For the functions to implement, I leave EnsureLoadBalancer() and DeleteLoadBalancer() empty and still try to gain a better understanding of the implementation
For other functions, I still have a lot of TODOs and probably need more contexts, so I might change my code a lot very soon, this just shows the basic structure of my implementation

nckturner:
A couple points for load balancer in V2.  We're planning on moving classic load balancers to a deprecated state so that new services use NLB's by default, and we currently have NLB code split between a target IP mode, which is going into the "aws-load-balancer-controller" and the existing NLB instance mode, which is in the legacy cloud provider.  We'd like to move both NLB target types to the aws-load-balancer-controller, so all new services would be delegated to that, unless they specifically requested legacy behavior.  In terms of what the legacy classic load balancer support looks like, I'm guessing what we're planning here will be fine, which from my perspective is basically the existing behavior with support for friendly names? 

/cc @M00nF1sh

nicolehanjing:
> A couple points for load balancers in V2. We're planning on eventually moving classic load balancers to a deprecated state so that new services use NLB's by default, and we currently have NLB code split between a target IP mode, which is going into the "aws-load-balancer-controller" and the existing NLB instance mode, which is in the legacy cloud provider. We'd like to move both NLB target types to the aws-load-balancer-controller, so all new services would be delegated to that, unless they specifically requested legacy behavior. In terms of what the legacy classic load balancer support looks like, I'm guessing what we're planning here will be fine, which from my perspective is basically the existing behavior with support for friendly names?
> 
> /cc @M00nF1sh

yeah most of the behavior refers to legacy implementation but we need to support a friendly name for the load balancer
I think we would want to have clusterName and serviceName included in the LB name, but also make this name unique with a maximum of 32 characters

PR #129, opened 2020-09-23
Create a new instance of aws cloud

nicolehanjing:
**What type of PR is this?**
/kind feature

This PR creates a new instance of aws cloud, which helps future PRs to commit in parallel with fewer conflicts

Part of #125

```release-note
Create a new instance of aws cloud of v2 provider.
```

### DIFF
--- a/pkg/providers/v2/cloud.go
+++ b/pkg/providers/v2/cloud.go
@@ -19,14 +19,21 @@ limitations under the License.
 package v2
 
 import (
+	"fmt"
 	"io"
+	"regexp"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
 	cloudprovider "k8s.io/cloud-provider"
 )
 
 func init() {
 	cloudprovider.RegisterCloudProvider(ProviderName, func(config io.Reader) (cloudprovider.Interface, error) {
-		return &cloud{}, nil
+		return newCloud()
 	})
 }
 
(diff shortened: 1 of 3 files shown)


pkg/providers/v2/interface.go
```diff
@@ -41,6 +48,86 @@ var _ cloudprovider.Interface = (*cloud)(nil)
 type cloud struct {
 }
 
+type awsSDKProvider struct {
+	creds *credentials.Credentials
```
andrewsykim, line 52: Not sure we need a new type like `awsSDKProvider` yet, seems premature.  Can we have `creds` and the region field be in `cloud` instead?

pkg/providers/v2/interface.go
```diff
@@ -41,6 +48,86 @@ var _ cloudprovider.Interface = (*cloud)(nil)
...
+	GetMetadata(path string) (string, error)
+}
+
+func getAvailabilityZone(metadata EC2Metadata) (string, error) {
+	return metadata.GetMetadata("placement/availability-zone")
+}
+
+// Derives the region from a valid az name.
+// Returns an error if the az is known invalid (empty)
+func azToRegion(az string) (string, error) {
```
andrewsykim, line 67: Can we add a unit test for this?

HaibaraAi96: sure, will add soon!

pkg/providers/v2/interface.go
HaibaraAi96: that sounds reasonable, will update!

pkg/providers/v2/interface.go
HaibaraAi96: @andrewsykim Updated!
I add a new file called interface_test.go and include the unit test for it
I also delete the type awsSDKProvider and call metadataClient directly in newCloud()

pkg/providers/v2/interface.go
```diff
@@ -39,6 +46,71 @@ var _ cloudprovider.Interface = (*cloud)(nil)
 
 // cloud is the AWS v2 implementation of the cloud provider interface
 type cloud struct {
```
andrewsykim, line 48: Line 36 above needs to be updated to call newCloud as part of `RegisterCloudProvider`

pkg/providers/v2/interface.go
```diff
@@ -39,6 +46,71 @@ var _ cloudprovider.Interface = (*cloud)(nil)
...
+	region := r.FindString(az)
+	if region == "" {
+		return "", fmt.Errorf("invalid AZ: %s", az)
+	}
+
+	return region, nil
+}
+
+// newCloud creates a new instance of AWSCloud.
+func newCloud() (*cloud, error) {
```
andrewsykim, line 80: This should return the interface it implements instead
```
func newCloud() (cloudprovider.Interface, error)
```

andrewsykim (commented):
/ok-to-test

pkg/providers/v2/cloud.go
```diff
@@ -39,6 +46,71 @@ var _ cloudprovider.Interface = (*cloud)(nil)
...
+}
+
+func getAvailabilityZone(metadata EC2Metadata) (string, error) {
+	return metadata.GetMetadata("placement/availability-zone")
+}
+
+// Derives the region from a valid az name.
+// Returns an error if the az is known invalid (empty)
+func azToRegion(az string) (string, error) {
+	if len(az) < 1 {
```
andrewsykim, line 66: nit: this probably reads slightly better as:
```
if len(az) == 0
```

HaibaraAi96: gotcha, updated!

pkg/providers/v2/cloud.go
```diff
@@ -19,14 +19,21 @@ limitations under the License.
...
+	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
 	cloudprovider "k8s.io/cloud-provider"
 )
 
 func init() {
 	cloudprovider.RegisterCloudProvider(ProviderName, func(config io.Reader) (cloudprovider.Interface, error) {
-		return &cloud{}, nil
+		return newCloud()
```
HaibaraAi96, line 36: @andrewsykim updated! 
Also passed the unit test, let me know if you have any other suggestions :)

andrewsykim (commented):
/approve
/lgtm

thanks @nicolehanjing

PR #127, opened 2020-09-23
[WIP] instances: add initial implementation of instancesV2 interface

andrewsykim:
Signed-off-by: Andrew Sy Kim <kim.andrewsy@gmail.com>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
This is a first pass at implementing Instances. This is still a WIP, some TODOs we should address before this PR merges:
* unit tests 
* support node naming policy other than private DNS names, we should ideally decide on what node naming policies are allowed
* implement InstanceShutdown
* others??

**Which issue(s) this PR fixes**:
Part of https://github.com/kubernetes/cloud-provider-aws/issues/125

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Add initial implementation of Instances for the v2 provider.
```

PR #131, opened 2020-09-25
instances: initial implementation of instancesV2 interface

nicolehanjing:
**What type of PR is this?**
/kind feature

Take over the initial work here: https://github.com/kubernetes/cloud-provider-aws/pull/127
This is a first pass at implementing Instances. 
Some TODOs:
- add more unit tests
- support node naming policy other than private DNS names

Part of #125

```release-note
Add initial implementation of Instances for the v2 provider.
```


pkg/providers/v2/instances_test.go
```diff
@@ -0,0 +1,177 @@
...
+	instances, err := newInstances(region, creds)
+	if err != nil {
+		t.Errorf("Error creating a new instance: %v", err)
+	}
+
+	for i := 0; i < 5; i++ {
+		nodeName := fmt.Sprintf("ip-172-21-32-%d.ec2.internal", i)
+
+		node := makeNode(nodeName, 2)
+		exists, err := instances.InstanceExists(context.TODO(), node)
```
nicolehanjing, line 140: I'm still figuring out what's the right way to trigger the unit test for those methods, will update soon

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,334 @@
...
+		return "", errors.New("instance ID was not set")
+	}
+
+	return "aws://" + aws.StringValue(instance.Placement.AvailabilityZone) + "/" + aws.StringValue(instance.InstanceId), nil
+}
+
+// parseInstanceIDFromProviderID parses the node's instance ID based on the well-known provider ID format:
+//   * aws://<availability-zone>/<instance-id>
+// This function always assumes a valid providerID format was provided.
+func parseInstanceIDFromProviderID(providerID string) (string, error) {
```
nicolehanjing, line 309: @andrewsykim is that intended to only parse well-formatted providerID? Should I take care of invalid cases?

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,334 @@
...
+// instanceShutdownByProviderID returns true if the instance is shutdown according to the cloud provider.
+func (i *instances) instanceShutdownByProviderID(ctx context.Context, providerID string) (bool, error) {
+	ec2Instance, err := i.getInstanceByProviderID(ctx, providerID)
+	if err != nil {
+		return false, err
+	}
+
+	if ec2Instance.State != nil {
+		state := aws.StringValue(ec2Instance.State.Name)
+		if state == ec2.InstanceStateNameTerminated || state == ec2.InstanceStateNameStopping || state == ec2.InstanceStateNameStopped {
```
nicolehanjing, line 234: @andrewsykim A little confused on how we define "shutdown"
I feel that states after "shutting down" (terminated, stopping, stopped) should all be considered as a "shutdown" state, is that right? 🤔

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,334 @@
...
+
+	if len(instances) == 0 {
+		return nil, cloudprovider.InstanceNotFound
+	}
+
+	if len(instances) > 1 {
+		return nil, fmt.Errorf("multiple instances found with provider ID: %s", instanceID)
+	}
+
+	state := instances[0].State.Name
```
nicolehanjing, line 176: should we consider the state of the instance?

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	return ret, err
+}
+
+func (i *instances) InstanceMetadata(ctx context.Context, node *v1.Node) (*cloudprovider.InstanceMetadata, error) {
+	var err error
+	var ec2Instance *ec2.Instance
+	if node.Spec.ProviderID == "" {
+		//  TODO: support node name policy other than private DNS names
```
nicolehanjing, line 107: not sure if I have enough contexts here, can you share more inputs? :) @andrewsykim

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,335 @@
...
+		if state == ec2.InstanceStateNameTerminated || state == ec2.InstanceStateNameStopping || state == ec2.InstanceStateNameStopped {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// nodeAddresses for Instance returns a list of v1.NodeAddress for the give instance.
+// TODO: should we support ExternalIP by default?
```
nicolehanjing, line 261: do we need to address this before merging this PR?

pkg/providers/v2/instances_test.go
```diff
@@ -0,0 +1,177 @@
...
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
nicolehanjing: gotcha!

pkg/providers/v2/instances.go
nicolehanjing: gotcha, thanks for the contexts!

pkg/providers/v2/instances.go
nicolehanjing: gotcha, thanks for the explanation!

pkg/providers/v2/instances_test.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	for _, testCase := range testCases {
+		providerID, err := getInstanceProviderID(&testCase.instance)
+		assert.NoError(t, err)
+		assert.Equal(t, testCase.providerID, providerID)
+	}
+}
+
+func TestInstanceExists(t *testing.T) {
```
nicolehanjing, line 135: @andrewsykim Updated the unit tests, let me know if you have any other suggestions :)

randomvariable:
We'll need some follow up issues to do the following, but not blocking this PR:
* Add rate limiters and metrics collection to the AWS request handlers
* Check if there's anything needed to support custom EC2 endpoints (for airgapped and Outposts setups)

andrewsykim:
> We'll need some follow up issues to do the following, but not blocking this PR:
>   Add rate limiters and metrics collection to the AWS request handlers
>   Check if there's anything needed to support custom EC2 endpoints (for airgapped and Outposts setups)

+1, we should definitely add the first one before removing the alpha gating env var:

https://github.com/kubernetes/cloud-provider-aws/blob/4eef54c572faaa38bfeeef357c59489af8781661/cmd/aws-cloud-controller-manager/main.go#L104-L108

pkg/providers/v2/cloud.go
```diff
@@ -107,10 +173,191 @@ func newCloud() (cloudprovider.Interface, error) {
...
+	awsCloud := &cloud{
+		creds:     creds,
+		instances: instances,
+		region:    region,
+		cfg:       &cfg,
+		metadata:  metadataClient,
+		ec2:       ec2Service,
+	}
+
+	selfAWSInstance, err := awsCloud.buildSelfAWSInstance()
```
andrewsykim, line 203: "selfAWSInstance" was only required in-tree because kubelet had to reference it's own instance. In the context here I don't think we need it.

pkg/providers/v2/cloud.go
```diff
@@ -107,10 +173,191 @@ func newCloud() (cloudprovider.Interface, error) {
...
+			return nil, err
+		}
+	}
+
+	return awsCloud, nil
+}
+
+// Builds the awsInstance for the EC2 instance on which we are running.
+// This is called when the AWSCloud is initialized, and should not be called otherwise (because the awsInstance for the local instance is a singleton with drive mapping state)
+func (c *cloud) buildSelfAWSInstance() (*awsInstance, error) {
```
andrewsykim, line 229: Likewise, I think we can remove this?

pkg/providers/v2/cloud.go
```diff
@@ -46,8 +55,50 @@ var _ cloudprovider.Interface = (*cloud)(nil)
...
+	creds     *credentials.Credentials
+	instances cloudprovider.InstancesV2
+	region    string
+	tagging   awsTagging
+	cfg       *CloudConfig
+	ec2       EC2
+
+	// The AWS instance that we are running on
+	// Note that we cache some state in awsInstance (mountpoints), so we must preserve the instance
+	selfAWSInstance *awsInstance
```
andrewsykim, line 67: likewise, we can remove this.

pkg/providers/v2/cloud.go
```diff
@@ -46,8 +55,50 @@ var _ cloudprovider.Interface = (*cloud)(nil)
 
 // cloud is the AWS v2 implementation of the cloud provider interface
 type cloud struct {
-	creds  *credentials.Credentials
-	region string
+	creds     *credentials.Credentials
+	instances cloudprovider.InstancesV2
+	region    string
+	tagging   awsTagging
```
andrewsykim, line 61: I think there are some implications around tagging that we may not want to introduce yet, let's add this in a follow up PR?

pkg/providers/v2/cloud.go
```diff
@@ -149,7 +396,7 @@ func (c *cloud) Routes() (cloudprovider.Routes, bool) {
 
 // HasClusterID returns true if the cluster has a clusterID
 func (c *cloud) HasClusterID() bool {
-	return false
+	return len(c.tagging.ClusterID) > 0
```
andrewsykim, line 399: I would be in favor of deferring tagging login in a follow-up PR. I _think_ we can work around any errors by setting the `--allow-untagged-cloud` flag.

nicolehanjing: gotcha!

pkg/providers/v2/cloud.go
nicolehanjing: yeah sure!

pkg/providers/v2/instances.go
randomvariable: After looking into kops and EKS in some detail, I don't think we would want "arbitrary" names, or at least if we do, it would have to come with huge caveats around node security.

In particular, EKS and kops-controller match a instance ID attestation with the privateDNSName, via an STS based authentication webhook in EKS via aws-iam-authenticator, or via issuance of a certificate with the privateDNSname as the node name with presentation of the AWS instance identity document in kops-controller's case (CAPI will likely also implement the latter).

In theory, this means we can securely support instanceID or privateDNSName, but beyond that, it would have to come with big "you don't want to use this" warnings.

@nckturner @justinsb

randomvariable: Also, why would we use a ConfigMap vs. say ComponentConfig for the controller?

pkg/providers/v2/instances_test.go
```diff
@@ -0,0 +1,243 @@
...
+			},
+		},
+	}
+
+	mockEC2.EXPECT().DescribeInstances(gomock.Any()).Return(expectOutput, nil).AnyTimes()
+
+	for i := 0; i < 5; i++ {
+		nodeName := fmt.Sprintf("ip-172-21-32-%d.ec2.internal", i)
+
+		node := makeNode(nodeName, 2)
```
andrewsykim, line 137: What's the reason for calling `mkaeNode` 5 times but the mocked DescribeInstance call returns only 1 instance?

Can we simplify this to only create 1 node or is there a test case we're exercising by creating mulitple?

pkg/providers/v2/instances_test.go
```diff
@@ -0,0 +1,243 @@
...
+				Instances: []*ec2.Instance{
+					&instance,
+				},
+			},
+		},
+	}
+
+	mockEC2.EXPECT().DescribeInstances(gomock.Any()).Return(expectOutput, nil).AnyTimes()
+
+	for i := 0; i < 5; i++ {
```
andrewsykim, line 171: same question here for why we're creating mulitple nodes.

pkg/providers/v2/instances_test.go
```diff
@@ -0,0 +1,243 @@
...
+		PrivateIpAddress: aws.String(privateIP),
+		PublicDnsName:    aws.String(publicDNSName),
+		PublicIpAddress:  aws.String(publicIP),
+		InstanceType:     aws.String("c3.large"),
+		Placement:        &ec2.Placement{AvailabilityZone: aws.String("us-west-1a")},
+		State: &ec2.InstanceState{
+			Name: aws.String(stateName),
+		},
+	}
+	if setNetInterface == true {
```
andrewsykim, line 49: In what scenario would set pass `setNetInterface=false` to `makeInstance`?

pkg/providers/v2/instances_test.go
```diff
@@ -0,0 +1,243 @@
...
+const TestClusterID = "clusterid.test"
+
+func makeInstance(num int, privateIP, publicIP, privateDNSName, publicDNSName string, setNetInterface bool, stateName string) ec2.Instance {
+	instance := ec2.Instance{
+		InstanceId:       aws.String(fmt.Sprintf("i-%d", num)),
+		PrivateDnsName:   aws.String(privateDNSName),
+		PrivateIpAddress: aws.String(privateIP),
+		PublicDnsName:    aws.String(publicDNSName),
+		PublicIpAddress:  aws.String(publicIP),
+		InstanceType:     aws.String("c3.large"),
```
andrewsykim, line 43: Instead of hard coding some of the test inputs, I'm wondering if we want to use test tables to input the `ec2.Instance` into each test case. A familiar example might be from https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/controllers/node/node_controller_test.go#L166-L171

andrewsykim: That way for each test case we can tweak the individual fields like instance type

pkg/providers/v2/instances.go
nckturner: EKS currently has a lot of built-in assumptions on private DNS.  That being said, instance ID is attractive because its guaranteed to be unique.  I'm ok with making it possible for other names, but we will need to be able to restrict it for EKS to stuff available to DescribeInstances, like @randomvariable said.  I don't think a ConfigMap is a good idea though.  Its seems like that suggests we want to allow on-the-fly reconfiguration, which for EKS we definitely wouldn't want.  Guessing, but if we allowed customers to make changes here then we'd probably want it protected behind an EKS API, which would be easier to do if it was configuration passed in via flags/ComponentConfig file on disk, and required a restart.

pkg/providers/v2/instances_test.go
nckturner: +1

nicolehanjing: gotcha, will update!

pkg/providers/v2/instances_test.go
nicolehanjing: I was making different nodes with different instanceID, that's why I call **makeNode** multiple times with different offset
but I think you are right, will simplify the code and update soon

pkg/providers/v2/instances_test.go
nicolehanjing: @andrewsykim I've updated the unit test to use test tables
but I think we can still keep this makeInstance() function, if we want to tweak the instanceType we can just pass it as a parameter(like stateName, privateIP, etc), I feel it's more readable and we don't need to repeat the code block of creating instances in every test, wdyt?

pkg/providers/v2/instances_test.go
nicolehanjing: I was just hoping to test some special cases, yeah I think maybe I should remove this

pkg/providers/v2/instances.go
andrewsykim: > Also, why would we use a ConfigMap vs. say ComponentConfig for the controller?

I'm not particularly tied to using ConfigMap here, but if we use a config file it should ideally be yaml/json and not INI. I'm not sure ComponentConfig is relevant here since this is a config file read by aws-cloud-controller-manager, not a config for any of its options/flags. 

> EKS currently has a lot of built-in assumptions on private DNS. That being said, instance ID is attractive because its guaranteed to be unique. I'm ok with making it possible for other names, but we will need to be able to restrict it for EKS to stuff available to DescribeInstances

Sounds like there needs to be a broader discussion on this topic for sure. @nicolehanjing since we know for sure that we still want to support private DNS, let's get this PR only working with that and make sure we have a follow-up PR to support instance ID and other naming policies.

pkg/providers/v2/instances.go
andrewsykim: I think we can defer this for a follow-up

pkg/providers/v2/instances_test.go
```diff
@@ -0,0 +1,431 @@
...
+func TestGetInstanceProviderID(t *testing.T) {
+	instance0 := makeInstance(0, "192.168.0.1", "1.2.3.4", "instance-same.ec2.internal", "instance-same.ec2.external", "running")
+	instance1 := makeInstance(1, "192.168.0.2", "", "instance-same.ec2.internal", "", "running")
+	instance2 := makeInstance(2, "192.168.0.3", "1.2.3.4", "instance-other.ec2.internal", "", "running")
+
+	testCases := []struct {
+		instance   ec2.Instance
+		providerID string
+	}{
+		{instance0, "aws:///us-west-1a/i-0"},
```
andrewsykim, line 93: would prefer to call makeInstance directly in the testcase here instead of indirectly referencing a constant

pkg/providers/v2/instances_test.go
```diff
@@ -0,0 +1,431 @@
...
+			expectedResult:  true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			expectedEc2Output := &ec2.DescribeInstancesOutput{}
+			state := aws.StringValue(test.fakeEc2Instance.State.Name)
+			if state != ec2.InstanceStateNameTerminated {
+				expectedEc2Output = &ec2.DescribeInstancesOutput{
```
andrewsykim, line 255: Wondering if the entire `ec2.DescribeInstancesOutput` should be part of the test table instead of just `ec2.Instance`. This way we can differe the output based on the state

pkg/providers/v2/cloud.go
```diff
@@ -46,8 +54,22 @@ var _ cloudprovider.Interface = (*cloud)(nil)
...
+	cfg       *CloudConfig
+	ec2       EC2
+	metadata  EC2Metadata
+}
+
+// CloudConfig wraps the settings for the AWS cloud provider.
+type CloudConfig struct {
+	Global struct {
+		// KubernetesClusterTag is the legacy cluster id we'll use to identify our cluster resources
+		KubernetesClusterTag string
```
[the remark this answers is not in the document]
andrewsykim: ack, thanks for raising that!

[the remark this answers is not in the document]
nicolehanjing: thanks for the contexts!

pkg/providers/v2/instances.go
nicolehanjing: gotcha! thanks for the info!

pkg/providers/v2/instances.go
nicolehanjing: 👌🏻

pkg/providers/v2/instances_test.go
nicolehanjing: gotcha!

pkg/providers/v2/instances_test.go
nicolehanjing: yeah I've considered that way before, that should also work, but I feel this way can save a lot more space
I think using the other way is probably more readable? Will update :)

pkg/providers/v2/instances_test.go
```diff
@@ -0,0 +1,431 @@
...
+		instance   ec2.Instance
+		providerID string
+	}{
+		{instance0, "aws:///us-west-1a/i-0"},
+		{instance1, "aws:///us-west-1a/i-1"},
+		{instance2, "aws:///us-west-1a/i-2"},
+	}
+
+	for _, testCase := range testCases {
+		providerID, err := getInstanceProviderID(&testCase.instance)
```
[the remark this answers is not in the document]
nicolehanjing: I see, good to know, updated!

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,365 @@
...
+			}
+		}
+	}
+
+	return addresses, nil
+}
+
+// getInstanceProviderID returns the provider ID of an instance which is ultimately set in the node.Spec.ProviderID field.
+// The well-known format for a node's providerID is:
+//    * aws:///<availability-zone>/<instance-id>
```
nicolehanjing, line 308: @andrewsykim I checked the master node's providerID and it's of this form:
`aws:///<availability-zone>/<instance-id>`
instead of "aws://<availability-zone>/<instance-id>"
I'm wondering if I need to handle both cases or only the one with three "/"s

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,365 @@
...
+	if node.Spec.ProviderID == "" {
+		_, err = i.getInstanceByPrivateDNSName(ctx, node.Name)
+		if err == cloudprovider.InstanceNotFound {
+			return false, nil
+		}
+
+		if err != nil {
+			return false, err
+		}
+	} else {
```
andrewsykim, line 84: I think this else is probably redundant, this can be :
```
    if node.Spec.ProviderID == "" {
		_, err = i.getInstanceByPrivateDNSName(ctx, node.Name)
		if err == cloudprovider.InstanceNotFound {
			return false, nil
		}

		if err != nil {
			return false, err
		}
 
                 return true, nil
    }

    _, err = i.getInstanceByProviderID(ctx, node.Spec.ProviderID)
   if err == cloudprovider.InstanceNotFound {
       return false, nil
   }

   if err != nil {
       return false, err
   }

   return true, nil
```

pkg/providers/v2/cloud.go
```diff
@@ -46,8 +54,22 @@ var _ cloudprovider.Interface = (*cloud)(nil)
...
+	creds     *credentials.Credentials
+	instances cloudprovider.InstancesV2
+	region    string
+	cfg       *CloudConfig
+	ec2       EC2
+	metadata  EC2Metadata
+}
+
+// CloudConfig wraps the settings for the AWS cloud provider.
+type CloudConfig struct {
```
andrewsykim, line 66: We should remove this if we're not reading it anywhere. I think the use of cluster tags needs a bit more discussion

andrewsykim (commented):
This is looking good for a first pass, left some minor comments

pkg/providers/v2/cloud.go
nicolehanjing: gotcha!

andrewsykim:
@nckturner mind giving this a review? This is just a first pass with base functionality, we still need to discuss naming policy, tagging, etc but I think we can do this iteratively in a follow-up PR.

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,361 @@
...
+	availabilityZone string
+	ec2              EC2
+	region           string
+}
+
+// InstanceExists indicates whether a given node exists according to the cloud provider
+func (i *instances) InstanceExists(ctx context.Context, node *v1.Node) (bool, error) {
+	var err error
+	if node.Spec.ProviderID == "" {
+		_, err = i.getInstanceByPrivateDNSName(ctx, node.Name)
```
nckturner, line 76: I wouldn't mind some level 4 (or maybe higher) logging here that printed a line "looking for node X by private DNS name".

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,361 @@
...
+		if err == cloudprovider.InstanceNotFound {
+			return false, nil
+		}
+
+		if err != nil {
+			return false, err
+		}
+	}
+
+	_, err = i.getInstanceByProviderID(ctx, node.Spec.ProviderID)
```
nckturner, line 86: Another log line at 4 or higher for "looking for node by provider ID".

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,361 @@
...
+	if node.Spec.ProviderID == "" {
+		_, err = i.getInstanceByPrivateDNSName(ctx, node.Name)
+		if err == cloudprovider.InstanceNotFound {
+			return false, nil
+		}
+
+		if err != nil {
+			return false, err
+		}
+	}
```
nckturner, line 84: If the instance does exist by private DNS name, and the provider ID is empty, we are still falling through this block and calling `getInstanceByProviderID()`?  Should we have an else statement for the provider ID check?

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,361 @@
...
+	region           string
+}
+
+// InstanceExists indicates whether a given node exists according to the cloud provider
+func (i *instances) InstanceExists(ctx context.Context, node *v1.Node) (bool, error) {
+	var err error
+	if node.Spec.ProviderID == "" {
+		_, err = i.getInstanceByPrivateDNSName(ctx, node.Name)
+		if err == cloudprovider.InstanceNotFound {
+			return false, nil
```
nckturner, line 78: If we group this entire thing into an if else, then we can also add a single `err != nil` and `err == cloudprovider.InstanceNotFound` at the end of the function, and add logs (maybe at level 6?) for not found instances.

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,361 @@
...
+	return i.instanceShutdownByProviderID(ctx, node.Spec.ProviderID)
+}
+
+// InstanceMetadata returns the instance's metadata.
+func (i *instances) InstanceMetadata(ctx context.Context, node *v1.Node) (*cloudprovider.InstanceMetadata, error) {
+	var err error
+	var ec2Instance *ec2.Instance
+	if node.Spec.ProviderID == "" {
+		//  TODO: support node name policy other than private DNS names
+		ec2Instance, err = i.getInstanceByPrivateDNSName(ctx, node.Name)
```
nckturner, line 113: Instead of having the `getInstanceByPrivateDNSName` and then `getInstanceByProviderID` logic in multiple places, can we put all that logic into a `getInstance(ctx context.Context, node *Node)  (*ec2.Instance, error)` function?  That way it will be easier to refactor or add other forms of `getInstance` in the future.

pkg/providers/v2/instances.go
nicolehanjing: gotcha, will add!

pkg/providers/v2/instances.go
nicolehanjing: updated! I wrap the logic in an if-else block so that ensures we only call `getInstanceByProviderID()` when the providerID is not empty

pkg/providers/v2/instances.go
nicolehanjing: the only difference in these two functions is the request, I think we could put the logic into a getInstance(ctx context.Context, node *Node) (*ec2.Instance, error) function but inside that function we need to differentiate the request

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,291 @@
...
+		InstanceType:  aws.StringValue(ec2Instance.InstanceType),
+		NodeAddresses: nodeAddresses,
+	}
+
+	return metadata, nil
+}
+
+// getInstance returns the instance if the instance with the given node info still exists.
+// If false an error will be returned, the instance will be immediately deleted by the cloud controller manager.
+func (i *instances) getInstance(ctx context.Context, node *v1.Node) (*ec2.Instance, error) {
```
nicolehanjing, line 143: @nckturner Updated!
- added the logs
- unified two functions `getInstanceByProviderID` and `getInstanceByPrivateDNSName` into `getInstance` and the only difference is the ec2 request input
- likely unified two functions `InstanceShutdownByProviderID` and `InstanceShutdownByPrivateDNSName` into `InstanceShutdown` 
PTAL! :)

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,291 @@
...
+		// get Instance by provider ID
+		instanceID, err := parseInstanceIDFromProviderID(node.Spec.ProviderID)
+		if err != nil {
+			return nil, err
+		}
+
+		request = &ec2.DescribeInstancesInput{
+			InstanceIds: []*string{aws.String(instanceID)},
+			Filters: []*ec2.Filter{
+				newEc2Filter("instance-state-name", aliveFilter...),
```
wongma7, line 164: It may be better to describe without this aliveFilter and instead filter out terminated instances locally like v1 did. I do not have numbers or documentation but my understanding is that filters can affect describeinstances performance. cc https://github.com/kubernetes/kubernetes/pull/78140

nicolehanjing: gotcha, will update! Thanks for the contexts

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,284 @@
...
+	ec2              EC2
+	region           string
+}
+
+// InstanceExists indicates whether a given node exists according to the cloud provider
+func (i *instances) InstanceExists(ctx context.Context, node *v1.Node) (bool, error) {
+	_, err := i.getInstance(ctx, node)
+
+	if err == cloudprovider.InstanceNotFound {
+		klog.V(6).Infof("instance not found")
```
andrewsykim, line 79: Log the node name here for the instance:
```
klog.V(6).Infof("instance not found for node: %s", node.Name)
```

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,284 @@
...
+	return true, nil
+}
+
+// InstanceShutdown returns true if the instance is shutdown according to the cloud provider.
+func (i *instances) InstanceShutdown(ctx context.Context, node *v1.Node) (bool, error) {
+	ec2Instance, err := i.getInstance(ctx, node)
+	if err != nil {
+		if err == cloudprovider.InstanceNotFound {
+			// We don't have live ec2Instance
+			return true, nil
```
andrewsykim, line 96: I don't think a non-existent instance should be considered "shutdown". This should only return `true, nil` if the instance exists but in a shutdown state.

pkg/providers/v2/instances_test.go
```diff
@@ -0,0 +1,466 @@
...
+
+	mockEC2 := mocks.NewMockEC2(mockCtrl)
+
+	fakeInstances := &instances{
+		ec2: mockEC2,
+	}
+
+	nodeName := "ip-192-168-0-1.ec2.internal"
+
+	tests := []struct {
```
andrewsykim, line 322: These tests are looking great :)

pkg/providers/v2/instances_test.go
```diff
@@ -0,0 +1,466 @@
...
+		name              string
+		node              *v1.Node
+		expectedEc2Output *ec2.DescribeInstancesOutput
```
andrewsykim, line 208: I would call this field `mockedEC2Output` instead, since we are not actually validating against this as an "expected" valuie.

andrewsykim: same applies for other tests

andrewsykim (commented):
Really minor comments

pkg/providers/v2/instances_test.go
nicolehanjing: gotcha, will update!

pkg/providers/v2/instances.go
nicolehanjing: I see, updated!

nicolehanjing:
@andrewsykim updated! PTAL :)

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,293 @@
...
+	ec2Instance, err := i.getInstance(ctx, node)
+	if err != nil {
+		if err == cloudprovider.InstanceNotFound {
+			klog.V(6).Infof("instance not found for node: %s", node.Name)
+			return false, nil
+		}
+
+		if err == ErrInstanceTerminated {
+			klog.V(6).Infof("instance terminated for node: %s", node.Name)
+			return true, nil
```
andrewsykim, line 104: ErrInstanceTerminated is for terminated instances which I don't think apply for the "shutdown" case. Checking `ec2.InstanceStateNameStopped` might be enough for checking shutdown.

andrewsykim: Oh I see you're already checking that below. In that case for terminated state this should return `false, nil` or `false, err`

nicolehanjing: gotcha, updated!

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	return true, nil
+}
+
+// InstanceShutdown returns true if the instance is shutdown according to the cloud provider.
+func (i *instances) InstanceShutdown(ctx context.Context, node *v1.Node) (bool, error) {
+	ec2Instance, err := i.getInstance(ctx, node)
+	if err != nil {
+		return false, err
```
nicolehanjing, line 94: @andrewsykim Updated the shutdown checks, PTAL!
Let me know if you have any other suggestions :)

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,281 @@
...
+	if len(instances) == 0 {
+		return nil, cloudprovider.InstanceNotFound
+	}
+
+	if len(instances) > 1 {
+		return nil, fmt.Errorf("getInstance: multiple instances found")
+	}
+
+	state := instances[0].State.Name
+	if *state == ec2.InstanceStateNameTerminated {
```
andrewsykim, line 191: If `DescribeInstances` returns only 1 instance and that instance has state `InstanceStateNameTerminated`, then I think we should treat it similar to the case of `len(instances) == 0` and here return `nil, cloudprovider.InstanceNotFound`.

pkg/providers/v2/instances.go
```diff
@@ -0,0 +1,281 @@
...
+		}
+		request.NextToken = nextToken
+	}
+
+	if len(instances) == 0 {
+		return nil, cloudprovider.InstanceNotFound
+	}
+
+	if len(instances) > 1 {
+		return nil, fmt.Errorf("getInstance: multiple instances found")
```
andrewsykim, line 187: nit: this should use errors.New

pkg/providers/v2/instances.go
nicolehanjing: gotcha, updated!
I was thinking to have a different log for terminated instances, but as "terminated" is just an intermediate state I agree that we should treat it the same as 'not exist'

andrewsykim (commented):
I think this is a reasonable starting point. Nothing added here is binding so we can revisit some decisions in a follow-up PR. Some functionality we need to revisit is:
* Cluster tagging support
* Support instance ID as the node name
* Configuration API to store the various thing that are configurable. 

Thanks @nicolehanjing 

/approve
/lgtm

PR #140, opened 2020-10-01
latest manifest should point to v1.19.0-alpha.1, not v1.19.1-alpha.1

andrewsykim:
Signed-off-by: Andrew Sy Kim <kim.andrewsy@gmail.com>

**What type of PR is this?**
/kind cleanup

Latest tag is v1.19.0-alpha.1, not v1.19.1-alpha.1, our example manifests should reflect that.

Fixes #

```release-note
NONE
```


nckturner:
/lgtm
Let me also rebuild and push the image with that tag.  I used v1.19.1 for the k/k dependency version and apparently forgot that the patch version was reserved as the cloud provider version.  My bad.  I'll also add some docs around this.  Looks like I got the tag right when I created the release though :confounded:

PR #148, opened 2020-10-27
[WIP] loadbalancer[2]: follow-up implementation of LoadBalancer interface for v2 provider

nicolehanjing:
**What type of PR is this?**
/kind feature

follow-up implementation for https://github.com/kubernetes/cloud-provider-aws/pull/128
implemented `EnsureLoadBalancer` function and addressed several TODOs
TODOs after this PR:
- support NLB

Part of #125

```release-note
Add follow-up implementation of loadbalancer for the v2 provider.
```


nckturner:
Just a note from our discussion, @justinsb mentioned the reason for the auto-generated names was due to the possibility of the load balancer name being publicly discoverable via DNS, so there could be a case where a user prefers that over friendly names, and we considered allowing the user to select the naming policy.

PR #149, opened 2020-10-28
tags: initial implementation of tags

nicolehanjing:
first pass of tags implementation

**What type of PR is this?**
/kind feature

We have two formats of the cluster tags in the legacy implementation:
```
// TagNameKubernetesClusterPrefix is the tag name we use to differentiate multiple
// logically independent clusters running in the same AZ.
// The tag key = TagNameKubernetesClusterPrefix + clusterID
// The tag value is an ownership value
const TagNameKubernetesClusterPrefix = "kubernetes.io/cluster/"

// TagNameKubernetesClusterLegacy is the legacy tag name we use to differentiate multiple
// logically independent clusters running in the same AZ.  The problem with it was that it
// did not allow shared resources.
const TagNameKubernetesClusterLegacy = "KubernetesCluster"
```

As @nckturner [mentioned before](https://github.com/kubernetes/cloud-provider-aws/pull/131#discussion_r501278537), the current format ("`kubernetes.io/cluster/<clusterID>": <ownership>`") causes issues for customers trying to use tags to organize billing. 

After discussion, we'd like to see tags with formats:
```
1) kubernetes.io/cluster/ = <clusterID>
2) kubernetes.io/cluster/<clusterID> = ""
```
the second format is only used for shared resources e.g. subnets

and once we have other workarounds, will make corresponding updates.

Part of #125

```release-note
NONE
```


pkg/providers/v2/cloud.go
```diff
@@ -53,6 +59,22 @@ type cloud struct {
 	region    string
 	ec2       EC2
 	metadata  EC2Metadata
+	tagging   awsTagging
+}
+
+// AWSCloudConfig wraps the settings for the AWS cloud provider.
+type AWSCloudConfig struct {
```
andrewsykim, line 66: Use the AWSCloudConfig type from the new apis package.

pkg/providers/v2/tags.go
```diff
@@ -0,0 +1,211 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+)
+
+const (
+	// TagNameKubernetesClusterPrefix is the tag name we use to differentiate multiple
+	// logically independent clusters running in the same AZ.
+	// we support two tags:
+	// 1) kubernetes.io/cluster/=<clusterName>
+	// 2) for shared resoueces between clusters, kubernetes.io/cluster/<clusterName>=""
+	TagNameKubernetesClusterPrefix = "kubernetes.io/cluster/"
```
andrewsykim, line 36: I think we agreed last time to only use `kubernetes.io/cluster=<cluster-name>` for tagging for now, this means not supporting the "shared" case below.

nicolehanjing: gotcha, removed!

nckturner:
So just to summarize our conversation from this morning, we need to keep the `kubernetes.io/cluster/<clusterID> = shared|owned` for discovery, but we will allow passing in `additionalTags` (or something similar) via the config to aid in billing and other tag use cases, like where users automatically delete all resources from an account that aren't tagged a certain way.  It also makes sense to use "tag-on-create" where possible because we believe it is better supported (although I am verifying that). [There have been some changes to the APIs](https://aws.amazon.com/about-aws/whats-new/2018/07/tag-on-create-for-amazon-ec2-and-amazon-ebs-is-now-available-in-the-aws-govcloud-region/) since V1 was written, so we should take advantage of those.

nckturner:
So actually as of now I think the resource tagging API is fully supported in all regions, so I don't think we need to prefer tag-on-create for that reason.  Tag-on-create might be nice for other reasons, like being eaiser to implement and fewer total number of api calls.

nicolehanjing:
> So just to summarize our conversation from this morning, we need to keep the `kubernetes.io/cluster/<clusterID> = shared|owned` for discovery, but we will allow passing in `additionalTags` (or something similar) via the config to aid in billing and other tag use cases, like where users automatically delete all resources from an account that aren't tagged a certain way. It also makes sense to use "tag-on-create" where possible because we believe it is better supported (although I am verifying that). [There have been some changes to the APIs](https://aws.amazon.com/about-aws/whats-new/2018/07/tag-on-create-for-amazon-ec2-and-amazon-ebs-is-now-available-in-the-aws-govcloud-region/) since V1 was written, so we should take advantage of those.

Thanks @nckturner for the summary
for my current implementation, I only used the first format which is "kubernetes.io/cluster/ = <clusterID>", but as we are keeping `kubernetes.io/cluster/<clusterID> = shared|owned`, I will need to keep both of them in this initial implementation, right? 
Regarding "tag-on-create", if I'm understanding it correctly, so once we have the implementation of tags, I think I need to add tags to resources, like LB, subnets, etc. while creating the resource.

pkg/providers/v2/cloud.go
```diff
@@ -61,6 +77,30 @@ type EC2Metadata interface {
 	GetMetadata(path string) (string, error)
 }
 
+func readAWSCloudConfig(config io.Reader) (*v1alpha1.AWSCloudConfig, error) {
+	if config == nil {
+		return nil, fmt.Errorf("no AWS cloud provider config file given")
```
andrewsykim, line 82: nit: use errors.New instead since there's no args

pkg/providers/v2/cloud.go
```diff
@@ -61,6 +77,30 @@ type EC2Metadata interface {
...
+	}
+
+	var cfg v1alpha1.AWSCloudConfig
+	err = yaml.Unmarshal(data, &cfg)
+	if err != nil {
+		// we got an error where the decode wasn't related to a missing type
+		return nil, err
+	}
+	if cfg.Kind != "AWSCloudConfig" {
+		return nil, fmt.Errorf("invalid cloud configuration object %q", cfg.Kind)
```
andrewsykim, line 98: "invalid Kind for cloud config: %q"

pkg/providers/v2/cloud.go
```diff
@@ -82,7 +122,7 @@ func azToRegion(az string) (string, error) {
 }
 
 // newCloud creates a new instance of AWSCloud.
-func newCloud() (cloudprovider.Interface, error) {
+func newCloud(cfg v1alpha1.AWSCloudConfig) (cloudprovider.Interface, error) {
```
andrewsykim, line 125: nit: pass in the pointer

pkg/providers/v2/tags.go
```diff
@@ -0,0 +1,188 @@
...
+				return "", fmt.Errorf("Found multiple cluster tags with prefix %s (%q and %q)", TagNameKubernetesClusterPrefix, clusterName, name)
+			}
+			clusterName = name
+		}
+	}
+
+	return clusterName, nil
+}
+
+func (t *awsTagging) init(clusterName string) error {
```
andrewsykim, line 73: In general I tend to avoid naming functions `init` because `init` is a reserved function name in Go

andrewsykim: Recommend creating a constructor function for awsTagging like `NewAWSTags` and have it receive clusterName as one of the parameters

pkg/providers/v2/cloud.go
```diff
@@ -53,6 +61,14 @@ type cloud struct {
 	region    string
 	ec2       EC2
 	metadata  EC2Metadata
+	tagging   awsTagging
```
andrewsykim, line 64: nit: just name this `tags`?

pkg/providers/v2/tags.go
```diff
@@ -0,0 +1,188 @@
...
+	// time.
+	createTagInitialDelay = 1 * time.Second
+	createTagFactor       = 2.0
+	createTagSteps        = 9
+)
+
+type awsTagging struct {
+	// ClusterName is our cluster identifier: we tag AWS resources with this value,
+	// and thus we can run two independent clusters in the same VPC or subnets.
+	ClusterName string
```
andrewsykim, line 51: Trying to recall previous discussions on this, did we agree on implementing `additionalTags` in this PR or a follow-up PR?

pkg/providers/v2/tags.go
```diff
@@ -0,0 +1,188 @@
...
+		return fmt.Errorf("AWS cloud failed to find ClusterName")
+	}
+
+	return nil
+}
+
+// Extracts a cluster name from the given tags, if one is present
+// If no clusterName is found, returns "", nil
+// If multiple (different) clusterNames are found, returns an error
+func (t *awsTagging) initFromTags(tags []*ec2.Tag) error {
```
andrewsykim, line 88: I couldn't find where this method was being called.

andrewsykim (commented):
Sorry for the late review @nicolehanjing

pkg/providers/v2/cloud.go
nicolehanjing: gotcha!

pkg/providers/v2/tags.go
nicolehanjing: If I remember it correctly, previously we agreed on making a follow-up PR

pkg/providers/v2/tags.go
nicolehanjing: oh right, I think I should remove this for now, but previously I thought it would be useful in the future

nicolehanjing:
@andrewsykim updated! PTAL :)

pkg/providers/v2/tags.go
nicolehanjing: https://github.com/kubernetes/cloud-provider-aws/pull/156

pkg/providers/v2/tags.go
```diff
@@ -0,0 +1,174 @@
...
+			}
+			clusterName = name
+		}
+	}
+
+	return clusterName, nil
+}
+
+func (t *awsTagging) hasClusterTag(tags []*ec2.Tag) bool {
+	// if the clusterName is not configured -- we consider all instances.
```
nckturner, line 88: Nit: wondering if this logic can be refactored slightly.  IMO returning true from a function called `hasClusterTag` when there isn't a tag to look for is bit confusing (for the sole purpose of causing some side effect behavior in the caller), wondering if its worth separating return values to have distinct values for "is cluster tag configured" and "does cluster tag exist".  We could return a specific error (or bool value) that the caller can check for if cluster tag is not configured.

pkg/providers/v2/tags.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	// no clusterName is a sign of misconfigured cluster, but we can't be tagging the resources with empty
+	// strings
+	if len(t.ClusterName) == 0 {
+		return tags
+	}
+
+	// tag format: kubernetes.io/cluster/=<clusterName>
+	tags[TagNameKubernetesClusterPrefix] = t.ClusterName
```
nckturner, line 116: I thought, for this tag, we were keeping the format: 
```
kubernetes.io/cluster/<clusterID> = shared|owned
```
and adding `additionalTags` for the billing use case?

nckturner: Ah sorry, just saw your above comment: 

> for my current implementation, I only used the first format which is kubernetes.io/cluster/ = <clusterID>, but as we are keeping kubernetes.io/cluster/<clusterID> = shared|owned, I will need to keep both of them in this initial implementation, right?

So I'm currently of the opinion that we should keep that tag as is, which will make a potential migration story easier, and just ask customers to solve their billing and other tagging use cases with `additionalTags`.  Happy to discuss this further if anyone has thoughts on this.

nicolehanjing: gotcha, thanks for the clarification, will update!

pkg/providers/v2/tags.go
nicolehanjing: good point, personally I prefer to return a specific error with some info for the caller to check, will update!

pkg/providers/v2/cloud.go
```diff
@@ -61,6 +78,30 @@ type EC2Metadata interface {
...
+		return nil, fmt.Errorf("unable to read cloud configuration from %q [%v]", config, err)
+	}
+
+	var cfg v1alpha1.AWSCloudConfig
+	err = yaml.Unmarshal(data, &cfg)
+	if err != nil {
+		// we got an error where the decode wasn't related to a missing type
+		return nil, err
+	}
+	if cfg.Kind != "AWSCloudConfig" {
```
ayberk, line 98: Are there any other required fields we have to verify here?

pkg/providers/v2/tags.go
andrewsykim: Sounds good, thanks!

pkg/providers/v2/cloud.go
andrewsykim: Good call, I think for this PR we don't technically have to validate anything since ClusterName is optional (I think) but there will be a need to validate field values at some point. Here's one example of how to do it:

https://github.com/kubernetes/kubernetes/blob/5344afd4fbfc3cb2ed6785d89b6f760886f100b5/pkg/credentialprovider/plugin/config.go#L71-L128

pkg/providers/v2/tags.go
```diff
@@ -0,0 +1,174 @@
...
+	// and thus we can run two independent clusters in the same VPC or subnets.
+	ClusterName string
+}
+
+// newAWSTags is a constructor function for awsTagging
+func newAWSTags(clusterName string) (awsTagging, error) {
+	if clusterName != "" {
+		klog.Infof("AWS cloud filtering on ClusterName: %v", clusterName)
+	} else {
+		return awsTagging{}, errors.New("AWS cloud failed to find ClusterName")
```
ayberk, line 60: If my understanding is correct, wee need to reword this error message. Something along the lines of `No ClusterName found in the config`.

pkg/providers/v2/tags.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (t *awsTagging) buildTags(additionalTags map[string]string, lifecycle string) map[string]string {
+	tags := make(map[string]string)
+	for k, v := range additionalTags {
+		tags[k] = v
+	}
+
+	// no clusterName is a sign of misconfigured cluster, but we can't be tagging the resources with empty
+	// strings
+	if len(t.ClusterName) == 0 {
```
ayberk, line 110: If this is a sign of misconfigured cluster, we should log a warning here.

pkg/providers/v2/tags.go
```diff
@@ -0,0 +1,174 @@
...
+// The error code varies though (depending on what we are tagging), so we simply retry on all errors
+func (t *awsTagging) createTags(ec2Client EC2, resourceID string, lifecycle string, additionalTags map[string]string) error {
+	tags := t.buildTags(additionalTags, lifecycle)
+
+	if tags == nil || len(tags) == 0 {
+		return nil
+	}
+
+	var awsTags []*ec2.Tag
+	for k, v := range tags {
```
ayberk, line 132: nit: can you rename these to `tagKey` and `tagValue`.

pkg/providers/v2/tags.go
```diff
@@ -0,0 +1,174 @@
...
+	// logically independent clusters running in the same AZ.
+	// tag format: kubernetes.io/cluster/<clusterID> = shared|owned
+	// The tag key = TagNameKubernetesClusterPrefix + clusterID
+	// The tag value is an ownership value
+	TagNameKubernetesClusterPrefix = "kubernetes.io/cluster/"
+
+	// createTag* is configuration of exponential backoff for CreateTag call. We
+	// retry mainly because if we create an object, we cannot tag it until it is
+	// "fully created" (eventual consistency). Starting with 1 second, doubling
+	// it every step and taking 9 steps results in 255 second total waiting
```
ayberk, line 42: 255 seconds sound awfully long to me. Any specific reason why we wait so long?

pkg/providers/v2/tags.go
nicolehanjing: sure, will update

pkg/providers/v2/tags.go
nicolehanjing: I was following the legacy implementation: https://github.com/kubernetes/legacy-cloud-providers/blob/master/aws/aws.go#L255-L262, but I didn't get enough contexts of it. 
We could definitely change these variables later

andrewsykim: yeah inheriting backoff logic from legacy seems fine, can we add a TODO comment to revisit these values?

pkg/providers/v2/cloud.go
```diff
@@ -61,6 +78,30 @@ type EC2Metadata interface {
 	GetMetadata(path string) (string, error)
 }
 
+func readAWSCloudConfig(config io.Reader) (*v1alpha1.AWSCloudConfig, error) {
```
andrewsykim, line 81: Can we add a unit test for this? Here's a similar example: https://github.com/kubernetes/kubernetes/blob/5344afd4fbfc3cb2ed6785d89b6f760886f100b5/pkg/credentialprovider/plugin/config_test.go#L30-L255 

I think you would have to initiate a buffer (bytes.NewBuffer) with the file contents, as opposed to passing the file name but the rest of the test logic should be similar.

pkg/providers/v2/tags.go
andrewsykim: Does this mean `clusterName` is actually required? If so we should validate it when we read the config and return an error if it's empty. I thought clusterName was optional and it can be discovered through instances but I might be mis-remembering the details.

andrewsykim: I think requiring clusterName is reasonable, it's an explicit step from the user requiring them to understand how/why the cluster name is needed.

pkg/providers/v2/tags.go
nicolehanjing: sure, will add

pkg/providers/v2/cloud.go
nicolehanjing: gotcha, will update!

pkg/providers/v2/tags.go
ayberk: If there's a way to infer the cluster name, I think that'd be ideal. "We will add the cluster name automatically, but you can override it via config" sounds reasonable to me. Although I'm not sure if there's a case where the user would want to override it.

nicolehanjing: yeah I agree, will figure out if we can discover the cluster name through instances
for now I leave a TODO note here and a warning log in https://github.com/kubernetes/cloud-provider-aws/pull/149/files#diff-78811b2c178b96d8cf3e546ccbc24defbe15e8ef228da31b36d17ecfe29174ceR183

pkg/providers/v2/cloud_test.go
```diff
@@ -44,3 +49,96 @@ func TestAzToRegion(t *testing.T) {
...
+	`,
+			config:    nil,
+			expectErr: true,
+		},
+	}
+
+	for _, testcase := range testcases {
+		t.Run(testcase.name, func(t *testing.T) {
+			var buffer bytes.Buffer
+			_, err := buffer.WriteString(testcase.configData)
```
andrewsykim, line 123: This can be simplieifed to:
```go
bytes.NewBufferString(testcase.configData)
```

andrewsykim (commented):
We also need to update the Instances implementation to use the clusterName tags for API calls.

pkg/providers/v2/cloud.go
```diff
@@ -30,12 +32,26 @@ import (
...
 
 func init() {
 	cloudprovider.RegisterCloudProvider(ProviderName, func(config io.Reader) (cloudprovider.Interface, error) {
-		return newCloud()
+		cfg, err := readAWSCloudConfig(config)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read AWS cloud provider config file: %v", err)
+		}
+
+		errs := validateAWSCloudConfig(cfg)
```
nicolehanjing, line 49: @andrewsykim updated! PTAL :) 
- added validation function & unit tests
- update the Instances implementation to use the clusterName tags for API calls

pkg/providers/v2/cloud.go
```diff
@@ -30,12 +32,26 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	cloudprovider "k8s.io/cloud-provider"
+	"k8s.io/cloud-provider-aws/pkg/apis/config/v1alpha1"
```
andrewsykim, line 37: import alias this to `awsconfigv1alpha1`

pkg/providers/v2/cloud.go
```diff
@@ -61,6 +85,52 @@ type EC2Metadata interface {
...
+// validateAWSCloudConfig validates AWSCloudConfig
+// clusterName is required
+func validateAWSCloudConfig(config *v1alpha1.AWSCloudConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if config.Kind != "AWSCloudConfig" {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("Kind"), "invalid Kind for cloud config: %q", config.Kind))
+	}
+
+	if config.APIVersion != "config.aws.io/v1alpha1" {
```
andrewsykim, line 118: use `awsconfigv1alpha1.SchemeGroupVersion.String()` instead of `config.aws.io/v1alpha1`

pkg/providers/v2/cloud.go
```diff
@@ -61,6 +85,52 @@ type EC2Metadata interface {
...
+	return &cfg, nil
+}
+
+// validateAWSCloudConfig validates AWSCloudConfig
+// clusterName is required
+func validateAWSCloudConfig(config *v1alpha1.AWSCloudConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if config.Kind != "AWSCloudConfig" {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("Kind"), "invalid Kind for cloud config: %q", config.Kind))
```
andrewsykim, line 115: field value should be the json serialized version `field.NewPath("kind")`

pkg/providers/v2/cloud.go
```diff
@@ -61,6 +85,52 @@ type EC2Metadata interface {
...
+	}
+
+	if config.APIVersion != "config.aws.io/v1alpha1" {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("APIVersion"), "invalid APIVersion for cloud config: %q", config.APIVersion))
```
andrewsykim, line 119: APIVersion -> apiVersion

pkg/providers/v2/cloud.go
```diff
@@ -61,6 +85,52 @@ type EC2Metadata interface {
...
+	}
+
+	fieldPath := field.NewPath("Config")
```
andrewsykim, line 122: Config -> config

pkg/providers/v2/cloud.go
```diff
@@ -61,6 +85,52 @@ type EC2Metadata interface {
...
+	if config.Config == (v1alpha1.AWSConfig{}) {
```
andrewsykim, line 123: I don't think this check is required, it would fail in field-specific validation anyways

pkg/providers/v2/cloud.go
```diff
@@ -61,6 +85,52 @@ type EC2Metadata interface {
...
+		allErrs = append(allErrs, field.Required(fieldPath, "Config is required, should not be empty"))
+	}
+
+	if len(config.Config.ClusterName) == 0 {
+		allErrs = append(allErrs, field.Required(fieldPath.Child("ClusterName"), "cluster name cannot be empty"))
```
andrewsykim, line 128: ClusterName -> clusterName

nicolehanjing:
@andrewsykim @nckturner @ayberk PTAL :)

andrewsykim (commented):
/approve
/lgtm

There's a few TODOs in the comments and the code but I think we can address those with https://github.com/kubernetes/cloud-provider-aws/pull/156

PR #151, opened 2020-11-02
Bump k8s.io/kubernetes@v1.20.0

nicolehanjing:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Pull in the latest v1.20.0 tagged release to get any bug fixes/improvements.
Reference: https://github.com/kubernetes/cloud-provider/blob/master/sample/basic_main.go

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:
There is a lot of changes there for v1.20 so pull in the latest branch of k8s.io/kubernetes and compile to make sure everything works
This PR will not merge until there is an official v1.20 branch

**Does this PR introduce a user-facing change?**:

```release-note
Bump k8s.io/kubernetes@v1.20.0
```

PR #154, opened 2020-11-12
Clean-up Kubernetes dependencies

zuzzas:
**What type of PR is this?**

/kind cleanup

1. Move away from unreadable "replace" directives in the go.mod
2. Provide a script that simplifies "replace" directives generation in the future

Starting with Kubernetes v1.17, all staging repos are tagged with respective tags. But the major version is zero, e.g. v1.17.4 -> v0.17.4. This allows is to move away from unreadable hashes in "replace" directives.

```release-note
NONE
```


nckturner:
I'm going to close this as I think its covered by https://github.com/kubernetes/cloud-provider-aws/pull/151.  That being said, I wouldn't be opposed to merging something like this onto the release-1.18 and release-1.19 branches (not sure if we need to script though).

PR #162, opened 2020-12-08
Fix cloudbuild and simplify Makefile & cloudbuild

nckturner:
- Remove cloudbuild logic from Makefile
- Fix ldflags version
- Use gcr.io/cloud-builders/docker for cloudbuild
  (tangentially, this should also fix the cloudbuild
  build, which was failing because of what appeared
  to be an env var collision between the Makefile
  and the build container with $IMAGE).

**What type of PR is this?**
>

/kind cleanup

Simplifies cloudbuild and Makefile (should fix it in the process, but we will need to merge it to see).  

Fixes #

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,20 +32,18 @@ ARG DISTROLESS_IMAGE=gcr.io/distroless/static@sha256:c6d5981545ce1406d33e61434c6
 # libc, muscl, etc.
 FROM ${GOLANG_IMAGE} as builder
 
-# This build arg is the version to embed in the CPI binary
-ARG VERSION=unknown
-
-# This build arg controls the GOPROXY setting
-ARG GOPROXY
+ARG VERSION
+ARG GOPROXY=https://goproxy.io,direct
+ARG GOOS=linux
 
 WORKDIR /build
 COPY go.mod go.sum ./
 COPY cmd/ cmd/
 COPY pkg/ pkg/
-COPY Makefile ./Makefile
-ENV CGO_ENABLED=0
-ENV GOPROXY ${GOPROXY:-https://proxy.golang.org}
-RUN make aws-cloud-controller-manager
+RUN GO111MODULE=on CGO_ENABLED=0 GOOS=${GOOS} GOPROXY=${GOPROXY} go build \
+		-ldflags="-w -s -X 'main.version=${VERSION}'" \
+		-o=aws-cloud-controller-manager \
+		cmd/aws-cloud-controller-manager/main.go
 
 ################################################################################
 ##                               MAIN STAGE                                   ##
--- a/Makefile
+++ b/Makefile
@@ -13,58 +13,26 @@
 # limitations under the License.
 #
 
-# Set by cloudbuild (see cloudbuild.yaml)
-# TAG will contain a tag of the form vYYYYMMDD-hash, vYYYYMMDD-tag, or vYYYYMMDD-tag-n-ghash,
-# depending on the git tags on your repo. The test-infra docs recommend using $_GIT_TAG
-# to tag our images.
-# PULL_BASE_REF will contain the base ref that was pushed to - for instance, master or
-# release-0.2 for a PR merge, or v0.2 for a tag.
-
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -13,58 +13,26 @@
...
-	DEV_TAG = $(TAG)
-else
-	DEV_TAG = $(GIT_VERSION)
-endif
-IMAGE ?= $(REGISTRY)/$(IMAGE_NAME):$(DEV_TAG)
-
-RELEASE_TAG := $(shell git describe --abbrev=0 2>/dev/null)
-RELEASE_IMAGE ?= $(REGISTRY)/$(IMAGE_NAME):$(RELEASE_TAG)
+VERSION ?= $(GIT_VERSION)
+IMAGE := amazon/cloud-controller-manager:$(VERSION)
```
andrewsykim, line 21: is this a real image or just a dummy replacement?

Dockerfile
```diff
@@ -32,20 +32,19 @@ ARG DISTROLESS_IMAGE=gcr.io/distroless/static@sha256:c6d5981545ce1406d33e61434c6
...
 
 WORKDIR /build
 COPY go.mod go.sum ./
 COPY cmd/ cmd/
 COPY pkg/ pkg/
 COPY Makefile ./Makefile
-ENV CGO_ENABLED=0
-ENV GOPROXY ${GOPROXY:-https://proxy.golang.org}
-RUN make aws-cloud-controller-manager
+RUN GO111MODULE=on CGO_ENABLED=0 GOOS=${GOOS} GOPROXY=${GOPROXY} go build \
```
ayberk, line 44: Any specific reason why we're not simply running make anymore?

Makefile
nckturner: Dummy replacement for now, but would be replaced by something real.

Dockerfile
nckturner: I didn't like the circular dependency of make -> dockerfile -> make.  It makes the env var/docker args logic hard to follow IMO.

Makefile
nckturner: Well to be clear, I was really intending this to just be the name we gave it so that it makes sense locally when listed with `docker images`, and the official image logic for gcr, ecr (when we add it) would be explicit, e.g. the the gcr logic is entirely contained in the cloudbuild.yaml file.

Dockerfile
ayberk: I see. I was more concerned about changing the build command on Makefile and then forgetting to update it here. I'm not super familiar with the best practices here, so I'll leave it to your judgement.

nckturner: Yeah I considered the downside but I don't think the command will change much at all

Dockerfile
```diff
@@ -32,20 +32,19 @@ ARG DISTROLESS_IMAGE=gcr.io/distroless/static@sha256:c6d5981545ce1406d33e61434c6
...
-ARG GOPROXY
+ARG VERSION
+ARG GOPROXY=https://goproxy.io,direct
+ARG GOOS=linux
 
 WORKDIR /build
 COPY go.mod go.sum ./
 COPY cmd/ cmd/
 COPY pkg/ pkg/
 COPY Makefile ./Makefile
```
wongma7, line 43: dont need to copy anymore

cloudbuild.yaml
```diff
@@ -1,19 +1,16 @@
...
-    - release-staging
+    - build
+    - --tag=gcr.io/$PROJECT_ID/k8s-staging-provider-aws:$_GIT_TAG
+    - --tag=gcr.io/$PROJECT_ID/k8s-staging-provider-aws:latest
+    - --build-arg=VERSION=$_GIT_TAG \
+    - .
 substitutions:
-  # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
-  # can be used as a substitution
   _GIT_TAG: '12345'
```
wongma7, line 15: is this intentional

wongma7 (approved):

PR #163, opened 2020-12-09
Fix cloudbuild image name

nckturner:
**What type of PR is this?**

/kind bug

Fixes cloudbuild

Fixes #

```release-note
NONE
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -7,8 +7,8 @@ steps:
   - name: gcr.io/cloud-builders/docker
     args:
     - build
-    - --tag=gcr.io/$PROJECT_ID/k8s-staging-provider-aws:$_GIT_TAG
-    - --tag=gcr.io/$PROJECT_ID/k8s-staging-provider-aws:latest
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -1,6 +1,6 @@
 images:
-  - 'gcr.io/$PROJECT_ID/cloud-controller-manager:$_GIT_TAG'
-  - 'gcr.io/$PROJECT_ID/cloud-controller-manager:latest'
+  - 'gcr.io/$PROJECT_ID/k8s-staging-provider-aws:$_GIT_TAG'
+  - 'gcr.io/$PROJECT_ID/k8s-staging-provider-aws:latest'
```
wongma7, line 3: both need to change i think to 

'gcr.io/k8s-staging-provider-aws/cloud-controller-manager:latest'

I am not sure what PROJECT_ID is, see https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/cloudbuild.yaml for reference

PR #164, opened 2020-12-11
Pass in cloud config file to initialize cloud provider

nicolehanjing:
pass the value from `--cloud-config` flag for reading the **clusterName** value
initialize cloud provider with the cloud provider name and config file provided

**What type of PR is this?**
/kind feature

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/aws-cloud-controller-manager/main.go
+++ b/cmd/aws-cloud-controller-manager/main.go
@@ -109,6 +109,11 @@ func main() {
 				}
 			}
 
+			cloudConfigFlag := cmd.Flags().Lookup("cloud-config")
+			if cloudConfigFlag.Value.String() == "" {
+				klog.Warning("empty cloud config file path")
+			}
+
 			cliflag.PrintFlags(cmd.Flags())
 
 			c, err := s.Config(KnownControllers(), app.ControllersDisabledByDefault.List())
@@ -118,7 +123,8 @@ func main() {
 			}
 
 			// initialize cloud provider with the cloud provider name and config file provided
-			cloud, err := cloudprovider.InitCloudProvider(cloudProvider, "")
+			cloudConfigFile := cloudConfigFlag.Value.String()
+			cloud, err := cloudprovider.InitCloudProvider(cloudProvider, cloudConfigFile)
 			if err != nil {
 				klog.Fatalf("Cloud provider could not be initialized: %v", err)
 			}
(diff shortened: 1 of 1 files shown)


cmd/aws-cloud-controller-manager/main.go
```diff
@@ -109,6 +109,11 @@ func main() {
 				}
 			}
 
+			cloudConfigFlag := cmd.Flags().Lookup("cloud-config")
+			if cloudConfigFlag.Value.String() == "" {
```
andrewsykim, line 113: I _think_ that the cloud config file would actually be available as `c.ComponentConfig.KubeCloudShared.CloudProvider.CloudConfigFile` after we run `s.Config()` on line 119.

andrewsykim: so no need to use `cmd` to look up the flag value.

nicolehanjing: I see!

cmd/aws-cloud-controller-manager/main.go
```diff
@@ -117,8 +117,13 @@ func main() {
 				os.Exit(1)
 			}
 
+			cloudConfigFile := c.ComponentConfig.KubeCloudShared.CloudProvider.CloudConfigFile
```
nicolehanjing, line 120: @andrewsykim updated! I think it would be good to warn when the file path is empty, wdyt?

andrewsykim: I think the warning isn't needed if the process would exit and fail if it was empty. But we would need to double check if that's the case. If `--cloud-config` is optionally the warning is probably not needed either.

nicolehanjing: gotcha, I think the configuration file path is optional by checking the code, so I remove the warning here

andrewsykim (commented):
/approve
/lgtm

Thanks @nicolehanjing

cmd/aws-cloud-controller-manager/main.go
andrewsykim: The examples should probably be updated as well instead of using a hardcoded path https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/sample/basic_main.go#L66

PR #173, opened 2021-02-05
feat: Helm chart for aws cloud controller manager

### LINKED ISSUE #119: A Helm chart for deployment

A Helm chart for deployment. Based on a few other AWS projects, it probably makes sense to generate the example manifest from the Helm chart too, to support a single-source-of-truth.

**Why is this needed**:

Managing things by raw manifest doesn't give you version tracking, rollback, configurability, or any of the other nice features that Helm provides. Particularly for an in-development system, being able to rollback and inspect differences between working and non-working k8s resources is very helpful.

/kind feature

JESWINKNINAN:
/kind feature

Helm Chart For AWS Cloud Controller Manager.

Fixes #119

NONE

```release-note: Helm Chart For AWS Cloud Controller Manager.
```

### DIFF
--- a/charts/aws-cloud-controller-manager/Chart.yaml
+++ b/charts/aws-cloud-controller-manager/Chart.yaml
@@ -1,0 +1,10 @@
+---
+apiVersion: v1
+name: aws-cloud-controller-manager
+description: This will install Cloud Controller Manager for AWS Cloud Provider
+version: 0.0.1
+appVersion: v1.19.0-alpha.1
+maintainers:
+- name: Jeswin K Ninan
+  email: jeswinjkn@gmail.com
+---
--- a/charts/aws-cloud-controller-manager/templates/cluserrolebinding.yaml
+++ b/charts/aws-cloud-controller-manager/templates/cluserrolebinding.yaml
@@ -1,0 +1,17 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ .Values.clusterRoleName }}
+  labels:
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+roleRef:
+  kind: ClusterRole
+  name: {{ .Values.clusterRoleName }}
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- apiGroup: ""
+  kind: ServiceAccount
+  name: {{.Values.serviceAccountName}}
+  namespace: {{.Values.namespace}}
+---
--- a/charts/aws-cloud-controller-manager/values.yaml
+++ b/charts/aws-cloud-controller-manager/values.yaml
@@ -1,0 +1,116 @@
+namespace: "kube-system"
+
+extraArgs: {}
+
+
+image:
+    repository: gcr.io/k8s-staging-provider-aws/cloud-controller-manager
+    tag:  v1.19.0-alpha.1
+
+# nameOverride -- String to partially override `aws-cloud-controller-manager.fullname`
+nameOverride: "aws-cloud-controller-manager"
+
+# nodeSelector -- Node labels for pod assignment. Ref: https://kubernetes.io/docs/user-guide/node-selection/.
+
+nodeSelector:
+  node-role.kubernetes.io/master: ""
+
+clusterRoleRules:
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - create
+  - patch
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - '*'
+- apiGroups:
+  - ""
+  resources:
+  - nodes/status
+  verbs:
+  - patch
+- apiGroups:
+  - ""
+  resources:
+  - services
+  verbs:
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - services/status
+  verbs:
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - create
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumes
+  verbs:
+  - get
+  - list
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - endpoints
+  verbs:
+  - create
+  - get
+  - list
+  - watch
+  - update
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - create
+  - get
+  - list
+  - watch
+  - update
+
+# resources -- Pod resource requests and limits.
+resources:
+  # limits:
+  #   cpu: 100m
+  #   memory: 300Mi
+  requests:
+    cpu: 200m
+  #   memory: 300Mi
+
+# tolerations -- List of node taints to tolerate (requires Kubernetes >= 1.6).
+tolerations:
+- key: node.cloudprovider.kubernetes.io/uninitialized
+  value: "true"
+  effect: NoSchedule
+- key: node-role.kubernetes.io/master
+  effect: NoSchedule
+
+clusterRoleName : system:cloud-controller-manager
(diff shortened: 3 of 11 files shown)


charts/aws-cloud-controller-manager/templates/cluserrolebinding.yaml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ .Values.clusterRoleName }}
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
```
wongma7, line 7: how about "helm.sh/chart" instead of "chart", the helm docs recommend it https://helm.sh/docs/chart_best_practices/labels/

> helm.sh/chart | REC | This should be the chart name and version: {{ .Chart.Name }}-{{ .Chart.Version \\| replace "+" "_" }}.

charts/aws-cloud-controller-manager/values.yaml
```diff
@@ -0,0 +1,116 @@
+namespace: "kube-system"
+
+extraArgs: {}
```
wongma7, line 3: this doesn't appear to be used

charts/aws-cloud-controller-manager/values.yaml
```diff
@@ -0,0 +1,116 @@
...
+- key: node-role.kubernetes.io/master
+  effect: NoSchedule
+
+clusterRoleName : system:cloud-controller-manager
+
+roleBindingName: cloud-controller-manager:apiserver-authentication-reader
+
+serviceAccountName: cloud-controller-manager
+
+refRoleName: extension-apiserver-authentication-reader
```
wongma7, line 116: how about roleRefName since the field is `roleRef.name`? or just roleName?

If someday we need to give ccm extra permissions then I guess we would create a role that has a superset of extension-apiserver-authentication-reader permissions and then change this default value.

charts/aws-cloud-controller-manager/values.yaml
JESWINKNINAN: @wongma7 This is correct, but expecting optional flags to be added in the future. So just kept with null values. I have few other values that are also added which are commented on. If you want to remove or wanted to add as a comment. Please let me know. I am okay with either of these changes.

charts/aws-cloud-controller-manager/values.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+# nameOverride -- String to partially override `aws-cloud-controller-manager.fullname`
+nameOverride: "aws-cloud-controller-manager"
+
+# nodeSelector -- Node labels for pod assignment. Ref: https://kubernetes.io/docs/user-guide/node-selection/.
+
+nodeSelector:
+  node-role.kubernetes.io/master: ""
+
+clusterRoleRules:
```
wongma7, line 18: what is benefit of putting in values versus embedding in yaml?

(I don't have much opinion on putting it in one place or the other, just wondering.)

charts/aws-cloud-controller-manager/values.yaml
JESWINKNINAN: Sure @wongma7 I will add as roleName.

charts/aws-cloud-controller-manager/values.yaml
wongma7: okay, i'm fine with keeping it as placeholder

wongma7 (approved):
overall lgtm just left some nitpciks and questions

charts/aws-cloud-controller-manager/values.yaml
JESWINKNINAN: Okay Thanks, @wongma7. I am resolving this review comment

charts/aws-cloud-controller-manager/values.yaml
nckturner, line 18: Can we have something like:

```
rbac:
    create: true

serviceAccount:
    create: true
```

In case users want to set it up a little differently or configure the RBAC to come from a different source?

JESWINKNINAN: I thought to add values that are KV as embedded for easy readability, Do you want to change. I think this looks good for a few values such as for tolerations, resources?

nckturner (changes requested):

charts/aws-cloud-controller-manager/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: v1
```
ayberk, line 2: Can we include appVersion as well? `"v1.19.0-alpha.1"`

charts/aws-cloud-controller-manager/values.yaml
JESWINKNINAN: @wongma7 This is changed as mentioned

charts/aws-cloud-controller-manager/Chart.yaml
```diff
@@ -0,0 +1,9 @@
...
+name: aws-cloud-controller-manager
+description: This will install Cloud Controller Manager for AWS Cloud Provider
+version: 0.0.1
```
ayberk, line 5: I'd say we should at least match the major and minor app version here. Thoughts?

charts/aws-cloud-controller-manager/values.yaml
JESWINKNINAN: @nckturner Sure, You want to change this in the current version, or can be added later? Please let me know

ayberk: +1. This was requested multiple times for the csi drivers. We need to support existing service accounts.

nckturner: If you prefer to add later that's fine

JESWINKNINAN: Sure @nckturner I will add a follow-up issue on this, Thanks :)

charts/aws-cloud-controller-manager/Chart.yaml
nckturner: I think the chart version is usually independent, isn't it?

ayberk: It is, but I personally find it less confusing if it matches the major and minor version. No strong preference here :)

nckturner: I think that would encourage it to be conflated with the app version, but also not a strong preference.

charts/aws-cloud-controller-manager/values.yaml
wongma7: I'm okay with it, no need to change.

charts/aws-cloud-controller-manager/Chart.yaml
JESWINKNINAN: @nckturner @ayberk Usually it is followed not with respect to the version which we are releasing. Please let me know your thoughts on this :)

charts/aws-cloud-controller-manager/values.yaml
JESWINKNINAN: Have a new squashed commit that resolves all comments you have added, kindly please review and let me know

charts/aws-cloud-controller-manager/Chart.yaml
JESWINKNINAN: Sure @ayberk I will add appVersion

charts/aws-cloud-controller-manager/Chart.yaml
nckturner: That's fine w/ me

nckturner (approved):

charts/aws-cloud-controller-manager/Chart.yaml
JESWINKNINAN: Resolve the conversation for now :), Have added appVersion in the chart as suggested by @ayberk. This will help the end-users to verify the version of the App I guess. Thanks

PR #183, opened 2021-02-25
Release 1.20

ayberk:

This is the release commit for v1.20. We should tag this commit for the release.

Note: There's an additional commit(the release documentation) because we created the branch before it was merged. It'll make merging back to master much easier.

```release-note
NONE
```


nckturner:
Thanks! 

/lgtm
/approve

PR #192, opened 2021-03-22
Use load balancer name annotation for classic ELBs

### LINKED ISSUE #124: Friendly Load Balancer Names and Adopting Existing Load Balancers

Continuation of https://github.com/kubernetes/kubernetes/issues/29789 which is closed because of deprecation of in-tree providers:

> `a44e18e4c552b11e683bb02fff13e176` is not a very friendly ELB name. The AWS console does not allow for searching/sorting by tags for every resource. We use something like this:
> 
> ```
> [environment]-[app]
> staging-my-app
> ```
> 
> It would be nice if we could pick a template to use for naming. A combination of cluster name and pod name might work well.

**Why is this needed**:

Deletion for a Service Type of Load Balancer is a common occurrence, being able to adopt existing load balancers and give them easy to read names will greatly enhance the user experience of kubernetes on AWS.

Also adding in comment from @KIVagant as it's a very popular one:

> Why just not add something like `service.beta.kubernetes.io/aws-load-balancer-name` [to annotations](https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/aws.go#L99) and allow users to manage this?

/kind feature

nckturner:
**What type of PR is this?**
>

/kind feature

- The aws-load-balancer-name annotation (`service.beta.kubernetes.io/aws-load-balancer-name`) can be used to override the default load balancer name.
- The load balancer name uses the string provided in the annotation, truncated and sanitized.

Fixes #124

None.

```release-note
Allows a user to override the default ELB name with `service.beta.kubernetes.io/aws-load-balancer-name`
```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -239,6 +239,9 @@ const volumeAttachmentStuck = "VolumeAttachmentStuck"
 // Indicates that a node has volumes stuck in attaching state and hence it is not fit for scheduling more pods
 const nodeWithImpairedVolumes = "NodeWithImpairedVolumes"
 
+// Allows the user to control the load balancer name from an annotation
+const ServiceAnnotationLoadBalancerName = "service.beta.kubernetes.io/aws-load-balancer-name"
+
 const (
 	// volumeAttachmentConsecutiveErrorLimit is the number of consecutive errors we will ignore when waiting for a volume to attach/detach
 	volumeAttachmentStatusConsecutiveErrorLimit = 10
@@ -4209,9 +4212,20 @@ func (c *Cloud) GetLoadBalancer(ctx context.Context, clusterName string, service
 	return status, true, nil
 }
 
+var inValidLBName = regexp.MustCompile(`[^a-zA-Z0-9-]`)
+
 // GetLoadBalancerName is an implementation of LoadBalancer.GetLoadBalancerName
 func (c *Cloud) GetLoadBalancerName(ctx context.Context, clusterName string, service *v1.Service) string {
-	// TODO: replace DefaultLoadBalancerName to generate more meaningful loadbalancer names.
+	if name, ok := service.Annotations[ServiceAnnotationLoadBalancerName]; ok {
+		if len(name) > 32 {
+			name = name[:32]
+			klog.Warningf("Load balancer name is over 32 characters, truncating: %s\n", name)
+		}
+		sanitised := inValidLBName.ReplaceAllString(name, "")
+		klog.Infof("Using sanitised load balancer name from annotations: %s\n", sanitised)
+		return sanitised
+
+	}
 	return cloudprovider.DefaultLoadBalancerName(service)
 }
 


pkg/providers/v1/aws.go
```diff
@@ -239,6 +239,9 @@ const volumeAttachmentStuck = "VolumeAttachmentStuck"
 // Indicates that a node has volumes stuck in attaching state and hence it is not fit for scheduling more pods
 const nodeWithImpairedVolumes = "NodeWithImpairedVolumes"
 
+// Allows the user to control the load balancer name from an annotation
+const ServiceAnnotationLoadBalancerName = "service.beta.kubernetes.io/aws-load-balancer-name"
```
wongma7, line 243: can you include more context in PR description about this change. Like are we trying to respect the same annotation as https://github.com/kubernetes-sigs/aws-load-balancer-controller? Or just coincidence it's the same?

pkg/providers/v1/aws.go
```diff
@@ -4209,9 +4212,20 @@ func (c *Cloud) GetLoadBalancer(ctx context.Context, clusterName string, service
...
+
 // GetLoadBalancerName is an implementation of LoadBalancer.GetLoadBalancerName
 func (c *Cloud) GetLoadBalancerName(ctx context.Context, clusterName string, service *v1.Service) string {
-	// TODO: replace DefaultLoadBalancerName to generate more meaningful loadbalancer names.
+	if name, ok := service.Annotations[ServiceAnnotationLoadBalancerName]; ok {
+		if len(name) > 32 {
+			name = name[:32]
+			klog.Warningf("Load balancer name is over 32 characters, truncating: %s\n", name)
+		}
+		sanitised := inValidLBName.ReplaceAllString(name, "")
```
wongma7, line 4224: test? Or comment what the regexp is doing. My reading of it is that it is removing all non-alphanumeric chars, is that right

Also, not sure if we should sanitize or error. Is sanitize what lb controller does?

pkg/providers/v1/aws.go
nckturner: Yeah its intentional that we are using the same annotation.

pkg/providers/v1/aws.go
nckturner: The LB controller sanitizes.  I considered erroring as well.  I am also not sure what would be preferrable... if there is ever a case where we want to look for a service given a load balancer, then erroring would be preferrable so that the elb name matches the annotation and the matching service can be easily found.  I'm not sure if that's important to support though.

wongma7: I dont like it but if lb controller is already doing it then best to be consistent. Still keep in mind accepting arbitrary inputs now means it will be very hard to change later.

rifelpet: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/1880 hasn't been included in an LB controller release yet, so its behavior could still be changed to error in this case

nckturner: Yeah good point, I'm leaning toward starting with error since it seems to be easier to switch to using a sanitized name later vs the other way around.

wongma7: As long as the error shows up somewhere discoverable (like as an event on the Service/Ingress object?) then I prefer error as well. Otherwise you could argue it's better to just sanitize and let the user figure out what happened. I don't fee lthat strongly either way, as long as cloudprovider and controller are consistent .

nckturner:
As the PR is now, adoption works, meaning if the `aws-load-balancer-name` service annotation is changed, and the new name points to an existing ELB, then Kubernetes will take over and modify that ELB.  It will not currently clean up the current (old) ELB, which I think we want to do, otherwise we're leaking ELBs.

nckturner:
The difficulty here is that the service controller util functions that we use, which call `EnsureLoadBalancer`, do not pass the old service so we don't know what the previous name was.  We could try to upstream a change to the service controller code that we're importing that passes the old service through to the cloud provider (unlikely unless other providers want it), or we could stop importing the service controller util functions and replace it with our own stuff (probably a lot of work), or we could try to come up with a different method of tracking the old ELB.

sciyoshi:
Is there still work going on for this @nckturner? Would be really helpful for us as we're trying to get an EKS cluster to re-use an existing classic loadbalancer.

PR #197, opened 2021-04-04
Add a docs page for the service controller

nckturner:
**What type of PR is this?**

/kind documentation

Add documentation for load balancer annotations.

Fixes #

```release-note
NONE
```


docs/service_controller.md
```diff
@@ -0,0 +1,29 @@
+# The Service Controller
+
+The service controller is responsible for watch for service and node object changes, so that it can create, update, or delete cloud load balancers corresponding to load balanced services.  Like the other controllers, we import the cloud-provider provided utility functions for managing the controller itself, which calls into cloud provider defined methods `GetLoadBalancer`, `GetLoadBalancerName`, `EnsureLoadBalancer`, `UpdateLoadBalancer`, and `EnsureLoadBalancerDeleted`.
+
```
ayberk, line 4: Annotation values have no constraints, right? I looked at the docs and looks like there are rules only for the keys. Just making sure so we don't give the users the wrong impression with `Valid Values : - `

mkdocs.yml
```diff
@@ -5,6 +5,8 @@ repo_name: kubernetes/cloud-provider-aws/
 repo_url: https://github.com/kubernetes/cloud-provider-aws/
 nav:
   - 'Home': README.md
+  - 'Components':
```
ayberk, line 8: This is minor, but I think these tabs are listed by the order they have here. We might want to move it to between user guide and development.

ayberk (approved):
Nothing to block this PR. Minor suggestions that can be done later.

docs/service_controller.md
nckturner: Yeah it might not be quite the correct way to phrase it..

mkdocs.yml
nckturner: That makes sense.  We can do in a follow up.

PR #199, opened 2021-04-12
Initial commit of an e2e test framework

### LINKED ISSUE #198: Test framework for e2e tests
Let's create a framework for e2e testing.  I suggest we borrow the structure of the ebs-csi-driver e2e tests (see [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/tests/e2e)).

nckturner:
**What type of PR is this?**

> /kind feature

- Uses same structure as the ebs-csi-driver
- Imports the upstream framework, which unfortunately means
  we have to reintroduce a dependency on k/k

Fixes #198

```release-note
NONE
```


nckturner:
@ayberk @wongma7 anything you would change about how e2e is run with ebs-csi-driver?

wongma7:
This won't work on its own will it, what is creating the cluster? I was thinking we could create a kops cluster with cloudprovider installed and then inject ours? (build a cloudprovider image, push, edit the deployment to refer to ours, and then run tests).

IMO we should add a presubmit job to test-infra first that runs `make test-e2e`, say we call it pull-kops. And then we can come back here and type `/test pull-kops` until it looks like we want it to and then merge this. as opposed to merging things bit by bit, we will have a fully functional test suite in one PR.

hack/e2e/run.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
```
wongma7, line 1: i want to avoid us maintaining 4 copies (ebs, fsx, efs, provider) of similar but not identical testing bash scripts. AT the moment this script just wraps ginkgo, do we need it?

wongma7: efs is a copy/paste of ebs. Are we willing to copy/paste it here too? https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/hack/e2e

wongma7: to avoid duplication of efforts wrt common testing stuff (create kops cluster, build image, test it, delete it) we can either:

1. Continue this copy/paste / git read-tree pattern, with ebs's hack/e2e tree as the canonical source.
2. Move ebs's hack/e2e tree into its own repo and use the kubernetes-csi/csi-release-tools git subtree pattern: https://github.com/kubernetes-csi/csi-release-tools
3. Design something else entirely from the ground up. (REALLY want to avoid this. As much as I dislike bash, it really is the best tool for the job. For example, if at some point we add kops to kubetest2 then it will be easy to shell out to kubetest2 instead of kops. csi-release-tools is mostly bash, gce pd's testing scripts are bash https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/tree/master/test, etc.)

nckturner: Sounds reasonable, but I like the CSI drivers depending on cloud provider rather than the other way around.  I can see that happening anyway in the future, and there's no reason for cloud provider to depend on any of the CSI repositories, so I'm fine with unifying as long as the source of truth is here, or a separate repo (but that's probably unnecessary).

nckturner: I take that back, we probably want a separate repo to avoid the dependencies having to import a bunch of go code along with the bash scripts that they want.

olemarkus: Note that kOps already is using kubetest2. The runs against the EBS CSI driver can be seen here: https://testgrid.k8s.io/kops-misc#kops-scenario-aws-ebs-csi-driver

The script for launching cluster + running tests is here: https://github.com/kubernetes/kops/blob/master/tests/e2e/scenarios/aws-ebs-csi/run-test.sh

I imagine you could set up something more or less identical with minimal amount of boilerplate/duplication.

wongma7: yeah the scenarios script in kops repo don't look bad, I think the cloud provider repo should copy that approach @nckturner

for CSI drivers, I need specifically to test using eksctl, so unless I implement a kubetest2 wrapper around eksctl the CSI drivers are still going to use the hack/e2e scripts for foreseeable future.  https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/hack/e2e/eksctl.sh

for testing on EKS in general, https://github.com/aws/aws-k8s-tester implements kubetest2, but I haven't tried yet.

Anyway, I might be worrying too much about the potential for code duplication, I think basically we should converge on relying on kubetest2 as much as possible and any other boilerplate around can be easily scripted/minimized as you point out in those run-test.sh scripts

PR #205, opened 2021-04-20
Bump dependencies to v1.21.0

nckturner:
Bump dependencies to v1.21.0

* Move to NewCloudControllerManagerCommand to remove copy/paste template
  code, we don't do anything too special here that requires
  duplication of the command creation.

```release-note
NONE
```


nckturner:
Tested successfully in kops, of note, `--configure-cloud-routes` must now be set to false and an additional permission needs to be set on the cluster role (create serviceaccounts/token in kube-system), but I don't understand why this is required for cloud-controller-manager, because ideally only kube-controller-manager should need to create service account tokens?

nckturner:
This is probably mergable to allow for testing 1.21 alpha images, we do need to address the permission in kops when we add the 1.21 image there (there is a hardcoded list of the alpha image versions right now). 

/unhold

PR #221, opened 2021-05-18
Use kustomize for example manifest

nckturner:
**What type of PR is this?**

/kind cleanup

- Default to `--use-service-account-credentials=true` in base manifest.
- reduce permissions on default manifest for use with `--use-service-account-credentials=true`.
- still give a "superset" configuration overlay for when this flag is disabled.
- disable route controller by default with `--configure-cloud-routes=false`.

Fixes #

```release-note
Use kustomize for manifest installation and reduce role to minimum required permissions when using service account credentials per controller.
```


manifests/overlays/superset-role/cluster-role-patch.yaml
```diff
@@ -86,8 +58,8 @@ rules:
   - create
   - get
   - list
-  - watch
   - update
+  - watch
```
nckturner, line 62: No change, just reordered.

manifests/overlays/superset-role/cluster-role-patch.yaml
```diff
@@ -96,25 +68,15 @@ rules:
...
```
nckturner, line 72: No change, just reordered.

manifests/overlays/superset-role/cluster-role-patch.yaml
```diff
@@ -38,7 +17,11 @@ rules:
   resources:
   - nodes
   verbs:
-  - '*'
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
nckturner, line 24: The cloud node controllers do not need `create node` or `delete node`.

manifests/overlays/superset-role/cluster-role-patch.yaml
```diff
@@ -96,25 +68,15 @@ rules:
   - create
   - get
   - list
-  - watch
   - update
+  - watch
 - apiGroups:
   - ""
+  resourceNames:
```
nckturner, line 75: Only allow service account creation for the cloud controller resource names.

manifests/overlays/superset-role/cluster-role-patch.yaml
```diff
@@ -59,25 +42,14 @@ rules:
   resources:
   - services/status
   verbs:
-  - list
   - patch
   - update
-  - watch
```
nckturner, line 65: I don't think list is possible with subresources, and I'm not sure about watch but we definitely don't do it.

docs/getting_started.md
```diff
@@ -13,10 +13,8 @@ Follow these steps when upgrading an existing cluster by launching the aws-cloud
 1. Add the `--cloud-provider=external` to the kube-apiserver config.
 1. Add the `--cloud-provider=external` to each the kubelet's config.
 1. Add the tag kubernetes.io/cluster/your_cluster_id=owned (if resources are owned and managed by the cluster) or kubernetes.io/cluster/your_cluster_id=shared (if resources are shared between clusters, and should not be destroyed if the cluster is destroyed) to your instances.
-1. Deploy the required RBAC to your cluster:
-    `kubectl apply -f https://raw.githubusercontent.com/kubernetes/cloud-provider-aws/master/manifests/rbac.yaml`
-1. Deploy the cloud-controller-manager to your cluster:
-    `kubectl apply -f https://raw.githubusercontent.com/kubernetes/cloud-provider-aws/master/manifests/aws-cloud-controller-manager-daemonset.yaml`
+1. Apply the kustomize configuration:
+   `kubectl apply -k ./manifests/base`
```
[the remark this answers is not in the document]
nckturner: Correct.  I thought it was a reasonable to require customers to clone the repo if we end up doing kustomize, but there are ways around it--kustomize supports some kinds of remote builds, i.e: `kustomize build "github.com/kubernetes-sigs/kustomize/examples/multibases/dev/?ref=v1.0.6"` is given as an example.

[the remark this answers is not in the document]
nckturner: Its actually pretty cool, try this: `kustomize build 'github.com/nckturner/cloud-provider-aws/manifests/base/?ref=use-kustomize-to-reduce-permissions'`.  I'll update the build command.

PR #230, opened 2021-06-05
Add support for returning IPv6 node addresses

hakman:
**What type of PR is this?**
/kind feature
/priority important-longterm

IPv6 support has been around for some time now both in AWS and Kubernetes. [Dual-Stack](https://kubernetes.io/blog/2021/04/08/kubernetes-1-21-release-announcement/#ipv4-ipv6-dual-stack-support) is also one of the major themes in Kubernetes 1.21.
I am trying to add IPv6 support for kOps and bumping into various issues because the IPv6 addresses are not returned by the cloud provider implementation.
xRef: https://github.com/kubernetes/kops/issues/8432

I am not sure if there was any reason for not implementing this until now, but this PR should be a good conversation starter.

n/a

I added the IPv6 addresses to the internal address list as it seems to be the convention in the related feature docs.
I also considered that adding the IPv6 addresses after all internal IPv4 would best preserve current behavior.

```release-note
Added support for returning IPv6 node addresses
```

/cc @nckturner @wongma7 @justinsb @aojea

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/http"
 	"path"
 	"regexp"
 	"sort"
@@ -1449,6 +1450,17 @@ func (c *Cloud) HasClusterID() bool {
 	return len(c.tagging.clusterID()) > 0
 }
 
+// isAWSNotFound returns true if the error was caused by an AWS API 404 response.
+func isAWSNotFound(err error) bool {
+	if err != nil {
+		var aerr awserr.RequestFailure
+		if errors.As(err, &aerr) {
+			return aerr.StatusCode() == http.StatusNotFound
+		}
+	}
+	return false
+}
+
 // NodeAddresses is an implementation of Instances.NodeAddresses.
 func (c *Cloud) NodeAddresses(ctx context.Context, name types.NodeName) ([]v1.NodeAddress, error) {
 	if c.selfAWSInstance.nodeName == name || len(name) == 0 {
@@ -1500,6 +1512,25 @@ func (c *Cloud) NodeAddresses(ctx context.Context, name types.NodeName) ([]v1.No
 					continue
 				}
 				addresses = append(addresses, v1.NodeAddress{Type: v1.NodeInternalIP, Address: internalIP})
+			}
+		}
+
+		for _, macID := range macIDs {
+			ipv6Path := path.Join("network/interfaces/macs/", macID, "ipv6s")
+			internalIPv6s, err := c.metadata.GetMetadata(ipv6Path)
+			if isAWSNotFound(err) {
+				// 404 -> no IPv6 addresses
+				continue
+			} else if err != nil {
+				return nil, fmt.Errorf("error querying AWS metadata for %q: %q", ipv6Path, err)
+			}
+			// return only the "first" address for each ENI
+			for _, internalIPv6 := range strings.Split(internalIPv6s, "\n") {
+				if internalIPv6 == "" {
+					continue
+				}
+				addresses = append(addresses, v1.NodeAddress{Type: v1.NodeInternalIP, Address: internalIPv6})
+				break
 			}
 		}
 
@@ -1590,6 +1621,22 @@ func extractNodeAddresses(instance *ec2.Instance) ([]v1.NodeAddress, error) {
 				addresses = append(addresses, v1.NodeAddress{Type: v1.NodeInternalIP, Address: ip.String()})
 			}
 		}
+	}
+
+	// handle internal network interfaces with IPv6 addresses
+	for _, networkInterface := range instance.NetworkInterfaces {
+		// skip network interfaces that are not currently in use
+		if aws.StringValue(networkInterface.Status) != ec2.NetworkInterfaceStatusInUse || len(networkInterface.Ipv6Addresses) == 0 {
+			continue
+		}
+
+		// return only the "first" address for each ENI
+		internalIPv6 := aws.StringValue(networkInterface.Ipv6Addresses[0].Ipv6Address)
+		ip := net.ParseIP(internalIPv6)
+		if ip == nil {
+			return nil, fmt.Errorf("EC2 instance had invalid IPv6 address: %s (%q)", aws.StringValue(instance.InstanceId), internalIPv6)
+		}
+		addresses = append(addresses, v1.NodeAddress{Type: v1.NodeInternalIP, Address: ip.String()})
 	}
 
 	// TODO: Other IP addresses (multiple ips)?
--- a/pkg/providers/v1/aws_test.go
+++ b/pkg/providers/v1/aws_test.go
@@ -589,7 +589,7 @@ func testHasNodeAddress(t *testing.T, addrs []v1.NodeAddress, addressType v1.Nod
 	t.Errorf("Did not find expected address: %s:%s in %v", addressType, address, addrs)
 }
 
-func makeInstance(num int, privateIP, publicIP, privateDNSName, publicDNSName string, setNetInterface bool) ec2.Instance {
+func makeInstance(num int, privateIP, publicIP, privateDNSName, publicDNSName string, ipv6s []string, setNetInterface bool) ec2.Instance {
 	var tag ec2.Tag
 	tag.Key = aws.String(TagNameKubernetesClusterLegacy)
 	tag.Value = aws.String(TestClusterID)
@@ -619,6 +619,13 @@ func makeInstance(num int, privateIP, publicIP, privateDNSName, publicDNSName st
 				},
 			},
 		}
+		if len(ipv6s) > 0 {
+			instance.NetworkInterfaces[0].Ipv6Addresses = []*ec2.InstanceIpv6Address{
+				{
+					Ipv6Address: aws.String(ipv6s[0]),
+				},
+			}
+		}
 	}
 	return instance
 }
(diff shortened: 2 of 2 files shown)


pkg/providers/v1/aws.go
```diff
@@ -1503,6 +1503,23 @@ func (c *Cloud) NodeAddresses(ctx context.Context, name types.NodeName) ([]v1.No
...
+				// Interfaces without an IPv6 address don't get the "ipv6s" data
+				klog.V(4).Infof("error querying AWS metadata for %q: %q", ipv6Path, err)
+				continue
+			}
+
+			for _, internalIPv6 := range strings.Split(internalIPv6s, "\n") {
+				if internalIPv6 == "" {
+					continue
+				}
+				addresses = append(addresses, v1.NodeAddress{Type: v1.NodeInternalIP, Address: internalIPv6})
```
aojea, line 1519: Should we should use it only if ip.IsGlobalUnicast()

johngmyers: Why should it exclude site-local unicast?

aojea: they are not routable, so other nodes will not be able to reach the node on that address

aojea: ah, wait , you mean site-local , those were deprecated on RFC 3879, there no site-local since then.

AFAIK https://golang.org/pkg/net/#IP.IsGlobalUnicast covers all the routable addresses.

hakman: Didn't add any check because wanted to keep it as close as possible to the IPv4 handling.
Maybe this would be something to do in a follow-up PR and try to improve both.

aojea: yeah, keeping it consistent looks better to me, you can discard this comment

hakman: 👍

pkg/providers/v1/aws.go
```diff
@@ -1592,6 +1609,24 @@ func extractNodeAddresses(instance *ec2.Instance) ([]v1.NodeAddress, error) {
...
+		// skip network interfaces that are not currently in use
+		if aws.StringValue(networkInterface.Status) != ec2.NetworkInterfaceStatusInUse {
+			continue
+		}
+
+		for _, internalIPv6 := range networkInterface.Ipv6Addresses {
+			if ipv6Address := aws.StringValue(internalIPv6.Ipv6Address); ipv6Address != "" {
+				ip := net.ParseIP(ipv6Address)
+				if ip == nil {
+					return nil, fmt.Errorf("EC2 instance had invalid private IPv6 address: %s (%q)", aws.StringValue(instance.InstanceId), ipv6Address)
```
anguslees, line 1623:
```suggestion
					return nil, fmt.Errorf("EC2 instance had invalid IPv6 address: %s (%q)", aws.StringValue(instance.InstanceId), ipv6Address)
```

pkg/providers/v1/aws.go
```diff
@@ -1592,6 +1609,24 @@ func extractNodeAddresses(instance *ec2.Instance) ([]v1.NodeAddress, error) {
...
+				}
+				addresses = append(addresses, v1.NodeAddress{Type: v1.NodeInternalIP, Address: ip.String()})
```
anguslees, line 1625: Ditto above comment about only reporting the first address on each ENI.

pkg/providers/v1/aws.go
```diff
@@ -1503,6 +1503,23 @@ func (c *Cloud) NodeAddresses(ctx context.Context, name types.NodeName) ([]v1.No
 			}
 		}
 
+		for _, macID := range macIDs {
+			ipv6Path := path.Join("network/interfaces/macs/", macID, "ipv6s")
+			internalIPv6s, err := c.metadata.GetMetadata(ipv6Path)
+			if err != nil {
```
anguslees, line 1509: If we ignore all errors (like this) then an IMDS timeout/throttle will result in our IPv6 node addresses disappearing sometimes.  That seems undesirable.

I think this needs to be more explicit:
```golang
// isAWSNotFound returns true if the error was caused by an AWS API 404 response.
func isAWSNotFound(err error) bool {
	if err != nil {
		var aerr awserr.RequestFailure
		if errors.As(err, &aerr) {
			return aerr.StatusCode() == http.StatusNotFound
		}
	}
	return false
}

// ...
if isAWSNotFound(err) {
    internalIPv6s = nil  // 404 -> no IPv6 addresses
} else if err != nil {
    return nil, fmt.Errorf("error querying AWS metadata for %q: %q", ipv6Path, err)
}
```

pkg/providers/v1/aws.go
```diff
@@ -1503,6 +1503,23 @@ func (c *Cloud) NodeAddresses(ctx context.Context, name types.NodeName) ([]v1.No
...
+				// Interfaces without an IPv6 address don't get the "ipv6s" data
+				klog.V(4).Infof("error querying AWS metadata for %q: %q", ipv6Path, err)
+				continue
+			}
+
+			for _, internalIPv6 := range strings.Split(internalIPv6s, "\n") {
```
anguslees, line 1515: I think we should add only the 'first' address (from each ENI).

In IPv4, AWS has a strong distinction between 'primary' IPv4 and all the 'secondary' addresses (per ENI).  In IPv6, the AWS API does not make a distinction, and it's just a single list of all IPv6 addresses (per ENI).

So if/when we want to add multiple IPv6 addresses to an ENI, eg so we have VPC addresses for pods, then they will _all_ turn up in this `ipv6s` list.  I think we need to be less greedy here :) and only use the first address, so the 'rest' can be left for other purposes.  From my testing ordering is preserved, so first means naive `ipv6s[0]`.

pkg/providers/v1/aws.go
anguslees, line 1519: > Should we should use it only if ip.IsGlobalUnicast()

Fwiw, I think we should use whatever address(es?) are reported by VPC (in metadata).  If VPC is giving us ULA addresses (for example), then there's a very good reason, and the code here should trust the VPC address is relevant and use it.

I have no reason to believe this will ever happen to be clear! but if VPC _is_ reporting something 'weird', then I think there's a very good chance that VPC is correct for that situation and our `if IsGlobalUnicast()` logic is going to be incorrect.

anguslees (changes requested):

pkg/providers/v1/aws.go
aojea: ULA are `IsGlobalUnicast()` , this just removes link-local, multicast and loopback
But agreed, this validation shouldn't be here and VPC is the source of truth

pkg/providers/v1/aws.go
hakman: Sounds good to me. Done, also added a test for this case.

hakman:
Thanks for the review @anguslees. I think I addressed all the comments.

anguslees (approved):
/lgtm

hakman:
Good news! I built a custom image to test this change and managed to get Conformance tests passing using an IPv6-only cluster created with kOps:
https://testgrid.k8s.io/kops-ipv6#kops-grid-scenario-ipv6-conformance

nckturner:
Sorry for the delay, I was on vacation.  This looks good to me.  
/approve

hakman:
@nckturner One final Q from me, if you don't mind. While playing with IPv6 clusters I noticed that k8s is picky about the order of the addresses. Would it make sense to move the IPv6 addresses first as they are optional and there is no reason to enable them unless someone wants to use them?

nckturner:
I do remember that address order being important, would you mind adding some more detail though? Also I see the gce provider adding ipv6 node addresses [here](https://github.com/kubernetes/cloud-provider-gcp/pull/240) and it looks like they are not first.

hakman:
As the fix is now, conformance tests pass on IPv6 only clusters, so this is not a huge issue.
Kubernetes reports node addresses as an ordered list and kubectl shows node IP as the first address (mostly aesthetic), though pods that run on host network report that address as their main address too.
Not sure if most tools support nodes and pods with multiple IPs so, ideally, there should be some preference.
With kube-controller-manager, kubelet allowed `--node-ip=::` to specify family preference, but with external CCM this is no longer possible:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/nodestatus/setters.go#L148

nckturner:
Spoke to @justinsb and he thinks it will be less likely to break users if we keep the ipv6 addresses at the end for now, with the potential of adding a configuration to return ipv6 first/only return ipv6.  I don't have a good idea of what will happen when we have dual stack configured for nodes, but I'm wondering about [this PR](https://github.com/kubernetes/kubernetes/pull/95239):

> On nodes with dual-stack node.status.addresses (regardless of how they got them), hostNetwork pods will now have dual-stack pod.status.podIPs, allowing the pod to be the target of either a single-stack IPv4 or single-stack IPv6 Service (or, once dual stack services #91824 merges, a dual-stack Service).

It seems that PR adds pod.status.podIPs, which can be set by the cloud provider node addresses in the case of hostNetwork pods.  Am I understanding that correctly?

aojea:
> It seems that PR adds pod.status.podIPs, which can be set by the cloud provider node addresses in the case of hostNetwork pods. Am I understanding that correctly?

hostNetwork pods use the IPs of the Host, until recently there was only one Node.Status.Address, hence hostNetwork pods always have only one PodIP.
Since that PR, hostNetwork pods in dual-stack clusters, will use one one IP of each family from node.Status.Addresses for its PodIPs.
Important, node.status.Addresses order matters and defines the "primary" ip family, i.e., a dual-stack cluster with IPv4,IPv6 has IPv4 as primary ip family.

hakman:
Sorry for my late reply, last week was pretty hectic for me. I think this is a good start and will get more eyes on IPv6 with AWS. We can adjust how it works in future PRs. Thanks @nckturner & @aojea.
/hold cancel

anguslees:
> I do remember that address order being important, would you mind adding some more detail though?

Just for the archives, order is extremely important here.  Basically this whole Addresses array is (and never has been) used for anything at all :P  Kubernetes only uses the first InternalIP address in this array (modifyable via `apiserver --kubelet-preferred-address-types`, but it's always only the first of each address type) - so adding the v6 addresses after the v4 address means they're unused in all but a v6-only scenario.

aojea:
with dual stack , addresses has a meaning, they define the order of the cluster ipv4 - ipv6 or ipv4 - ipv6

PR #238, opened 2021-07-11
Add support for consuming web identity credentials

olemarkus:
CCM builds its own credential providers list and configures them directly instead of using those from the session.
This PR checks the presence of the web identity env vars and adds a web identity role provider.

**What type of PR is this?**
/kind feature

I have been working on IRSA support for the AWS CCM for kOps, and I had to add something like this in order for CCM to pick up the AWS credentials. Without this, it would just use the instance role, which no longer has the necessary AWS permissions to run CCM.

```release-note
Cloud Controller Manager can now consume credentials through web identity role credentials.
```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -36,7 +36,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
-	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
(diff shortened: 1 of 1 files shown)


pkg/providers/v1/aws.go
```diff
@@ -1194,7 +1195,16 @@ func init() {
 		}
 
 		var provider credentials.Provider
-		if cfg.Global.RoleARN == "" {
+		if os.Getenv("AWS_WEB_IDENTITY_TOKEN_FILE") != "" {
```
nckturner, line 1198: I know we already have this unfortunate if statement, but it seems like the purpose of `credentials.NewChainCredentials` is to allow an order to be specified -- can we simply add the `NewWebIdentityRoleProvider` to the chain?  

I think what we should have done instead of `if cfg.Global.RoleARN == "" ...`, is to just create a custom provider similar to AssumeRoleProvider which takes an arn from the config file, and just add it to the NewChainCredentials call (if the role is not provided it should return an error and the [next provider will be called](https://github.com/aws/aws-sdk-go/blob/e2d6cb448883e4f4fcc5246650f89bde349041ec/aws/credentials/chain_provider.go#L74-L81)).  

Let me know if you already tried something like that--perhaps I'm misunderstanding why it was done this way.

olemarkus: Actually, the identity session should have been picked up from the original session that we pass on with:
```
	&credentials.StaticProvider{
		Value: sess.Config.Credentials.Get(),
	}
```

Maybe it is as simple as enabling the shared config:
```
		sess, err = session.NewSessionWithOptions(session.Options{
			Config:            *cfg,
			SharedConfigState: session.SharedConfigEnable,
		})
```

Let me try that.

olemarkus:
@nckturner I changed the PR to obtain credentials from the default credentials chain instead. The effect should be similar, but doesn't duplicate the logic for building the web identity credentials.

olemarkus:
Just realised the problem with this approach is that credentials probably will not renew itself. I tweaked this approach a bit to only pass on the list of credentials providers if we are actually using the role from cloudconfig. Passing on nil otherwise will make sessions just use the default credential list otherwise, which will now also support web identity credentials.

nckturner:
Agreed, using the default chain is a better approach.  Thanks!

/lgtm

PR #253, opened 2021-08-25
Add script to facilitate cherry-picking from k/k

nckturner:
Until legacy-cloud-provider is deleted, there is a possibility that
  bug fixes are merged into k/k before this respository. This script
  makes it easier to cherry-pick commits from there to here, which
  requires a path rewrite and is a pain to do by hand.  This will
  probably stop being useful if the v1 provider in this repository
  diverges from the legacy provider in k/k, and we should encourage
  contributions to come here first.

**What type of PR is this?**

/kind cleanup

Quick script to facilitate cherry picking and path rewriting from k/k
 
```release-note
NONE
```

### DIFF
--- a/hack/cherry-pick-kk.sh
+++ b/hack/cherry-pick-kk.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+UPSTREAM_SHA="${UPSTREAM_SHA:-}"
+REWRITE_PATH="${REWRITE_PATH:-pkg/providers/v1}"
+LEGACY_PROVIDER_PATH="${LEGACY_PROVIDER_PATH:-staging/src/k8s.io/legacy-cloud-providers/aws}"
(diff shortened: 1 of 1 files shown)


hack/cherry-pick-kk.sh
```diff
@@ -0,0 +1,58 @@
...
+commit_author=$(git log -1 --pretty=format:'%an <%ae>' ${UPSTREAM_SHA})
+commit_body=$(git log -1 --pretty=format:'%B' ${UPSTREAM_SHA})
+
+patch_file=$(mktemp)
+git diff "${UPSTREAM_SHA}^" "${UPSTREAM_SHA}" -- ${LEGACY_PROVIDER_PATH} > ${patch_file}
+echo "Saving diff to ${patch_file}"
+popd > /dev/null
+
+echo "Modifying paths in patch..."
+sed -i "s|staging/src/k8s.io/legacy-cloud-providers/aws|pkg/providers/v1|g" "${patch_file}"
```
wongma7, line 46: git can do this with -p and --directory btw https://stackoverflow.com/questions/931882/how-to-apply-a-git-patch-from-one-repository-to-another

Issue #271, opened 2021-09-10
Deprecate and remove support of the KubernetesCluster tag 


rifelpet:
The `KubernetesCluster` tag is considered "legacy" and is replaced by the more-flexible `kubernetes.io/cluster/<clustername>` tag:

https://github.com/kubernetes/cloud-provider-aws/blob/58bc609b9f79e1283b8896011f1590691604b2aa/pkg/providers/v1/tags.go#L30-L39

We should adopt a timeline for a formal deprecation and removal of any use of `KubernetesCluster`.

/kind deprecation

nckturner:
Agreed.  It has already been deprecated, so I think we can drop it in the next minor version release.

nckturner:
The safer alternative, if we think this tag is still being used, is to amplify the existing deprecation messages for a release or two, and then drop it.

nckturner:
Alas, I thought there was a deprecation log message, but I can't find one... only the comment you linked.  Looks like we will need to add one and determine a schedule, as you mentioned.

PR #272, opened 2021-09-14
Filter CLB instances by cluster tag

### LINKED ISSUE #269: Classic ELB reconciliation requires unfiltered DescribeInstances call
**What happened**:

The Classic ELB reconciliation code uses ec2.DescribeInstances to find all relevant instance IDs that should be added to the Classic ELB:

`findInstancesForELB` calls `describeAllInstancesCached` which calls `describeAllInstancesUncached` which calls `describeInstances`

https://github.com/kubernetes/cloud-provider-aws/blob/954bbd1f9b5ef824d1783b3e0ee8672061a3b052/pkg/providers/v1/aws_loadbalancer.go#L1600-L1617

https://github.com/kubernetes/cloud-provider-aws/blob/954bbd1f9b5ef824d1783b3e0ee8672061a3b052/pkg/providers/v1/instances.go#L196-L208

https://github.com/kubernetes/cloud-provider-aws/blob/954bbd1f9b5ef824d1783b3e0ee8672061a3b052/pkg/providers/v1/instances.go#L152-L162

https://github.com/kubernetes/cloud-provider-aws/blob/954bbd1f9b5ef824d1783b3e0ee8672061a3b052/pkg/providers/v1/aws.go#L5125-L5141

Notice that describeAllInstancesUncached does not provide any filters to the underlying `ec2.DescribeInstances`.

In accounts with many (10s of thousands) of instances in one region, this call is very slow and even fails with HTTP 500 responses. This causes reconciliation of the Classic ELB's instances to fail. Over time this can lead to fewer and fewer instances behind the ELB due to node attrition which could lead to service disruption.

The `ec2.DescribeInstances` call should have sufficient filters to reduce its scope as much as possible.

Adding filters to DescribeInstances could reduce the scope sufficiently that the AWS API will consistently succeed and in reasonable time.

Notice in the last code block above that the list of instances in the response is then filtered to only those that include the cluster tag:

https://github.com/kubernetes/cloud-provider-aws/blob/954bbd1f9b5ef824d1783b3e0ee8672061a3b052/pkg/providers/v1/tags.go#L133-L151

This checks if the instance has either `kubernetes.io/cluster/clustername=` or `KubernetesCluster=clustername` tags.

I'm wondering if we could include these in the request's filters instead. The EC2 API responds with the intersection of multiple filters, so it may be difficult to filter for the union of those tags. We may also be able to explore the [ResourceGroupTagging API](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/resourcegroupstaggingapi#Client.GetResources) to get the list of instance IDs. If the controller doesn't have the cluster name (ClusterID) set, it could fallback to the existing behavior.

Additionally, since Classic ELBs [can only target instances in the same VPC](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-deregister-register-instances.html#elb-register-instances-prereq), if the VPC ID is known around this area of code we could include it in the filter too.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.22

/kind bug

rifelpet:
**What type of PR is this?**

/kind bug

This reduces the response time and size of this ec2.DescribeInstances call.

This has a side-effect of no longer recognizing instances that are only tagged with the legacy KubernetesCluster tag.

Fixes #269

The `KubernetesCluster` tag is considered legacy (see https://github.com/kubernetes/cloud-provider-aws/issues/269#issuecomment-917155670 and https://github.com/kubernetes/cloud-provider-aws/issues/271) so this may be the first case of AWS CCM no longer recognizing `KubernetesCluster`.

The filtering that happens with the response was already checking for either `kubernetes.io/cluster/<clustername>` or `KubernetesCluster`, so the only behavior change is no longer recognizing `KubernetesCluster`:

https://github.com/kubernetes/cloud-provider-aws/blob/ccbcc7a926acb5fb8e0e0115df34b949d39a7590/pkg/providers/v1/aws.go#L5137-L5141

Looking at the [hasClusterTag implementations in v1 and v2](https://github.com/kubernetes/cloud-provider-aws/search?q=hasClusterTag) it appears that v2 already does not recognize `KubernetesCluster`.

```release-note
EC2 Instances not tagged with `kubernetes.io/cluster/<clustername>` will no longer be considered for Classic Load Balancer registration
```


wongma7:
My understanding is that a call with a filter is as expensive behind the scenes as a call with no filter. I don't have documentation to prove this, it's from heresy and internal discussions. The recommended way to make a call less expensive in terms of response time / size and so less likely to receive 5xx is to:

1. pass maxresults and next token, i.e. paginate. We should be doing this in EVERY DescribeInstances call no matter what. The only reason calls are not paginated by default by ec2 API is because the ec2 API must be backwards compatible and it's old but as a client we ought to paginate. @nckturner is working on this here https://github.com/kubernetes/cloud-provider-aws/pull/274/

2. pass instance ids. If we were to pass instance ids in this call, it would be a slight behavior change that could result in higher call volume. Going up the call stack, the nodes provided to this function are Ready nodes in the cluster that do NOT have the "exclude load balancer" annotation: https://github.com/kubernetes/kubernetes/blob/3723713c550f649b6ba84964edef9da6cc334f9d/staging/src/k8s.io/cloud-provider/controllers/service/controller.go#L668. So if we were to call DescribeInstances with just the instance ids corresponding to those provided nodes then we would miss instances corresponding to NotReady nodes or with the exclude label. And if later a node became Ready or the label were removed then our cache would be stale and we would need to call DescribeInstances again to add those nodes to the cache, that's how call volume could increase.

wongma7:
also you can't do 1 and 2 at the same time. So for a given call we have to either paginate it or provide instance ids, not both, or the call will fail

nckturner:
I've asked internally for confirmation that this is, in fact better, and will update with a response.

PR #274, opened 2021-09-17
Set MaxResults if it is not set

nckturner:
* EC2 pagination doesn't work if MaxResults is not set
* Only set if InstanceIds isn't set.

**What type of PR is this?**
>

/kind bug

The describeInstances API doesn't paginate unless maxResults is set.  
Fixes #

/hold 

```release-note
Ensure pagination of EC2 describeInstances calls
```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -970,6 +970,13 @@ func (s *awsSdkEC2) DescribeInstances(request *ec2.DescribeInstancesInput) ([]*e
 	results := []*ec2.Instance{}
 	var nextToken *string
 	requestTime := time.Now()
+
+	if request.MaxResults == nil && request.InstanceIds == nil {
+		// MaxResults must be set in order for pagination to work
+		// MaxResults cannot be set with InstanceIds
+		request.MaxResults = aws.Int64(1000)
+	}
+
 	for {
 		response, err := s.ec2.DescribeInstances(request)
 		if err != nil {


wongma7:
https://github.com/kubernetes/kubernetes/pull/102927 this updated the cache to have a ttl of just 10 minutes. Seems like a very  bad idea if we are trying to optimize our calls.

wongma7:
do you mind double checking that describeAllInstancesUncached https://github.com/kubernetes/cloud-provider-aws/blob/58bc609b9f79e1283b8896011f1590691604b2aa/pkg/providers/v1/instances.go#L153 ultimately calls this function? I am confident in my ctrl+clicking skills but just want to make 100% sure that we're covering all calls given that there are quite a few layers (describeAllInstancesUncached, describeInstances lowercase, DescribeInstances uppercase, sdk go's DescribeInstances uppercase...

wongma7:
describeAllInstancesUncached calls c.cloud.describeInstances (c is for cache)
https://github.com/kubernetes/cloud-provider-aws/blob/58bc609b9f79e1283b8896011f1590691604b2aa/pkg/providers/v1/instances.go#L159
https://github.com/kubernetes/cloud-provider-aws/blob/58bc609b9f79e1283b8896011f1590691604b2aa/pkg/providers/v1/aws.go#L5002
c.cloud.describeInstances calls c.ec2.DescribeInstances (now c is for cloud)
https://github.com/kubernetes/cloud-provider-aws/blob/58bc609b9f79e1283b8896011f1590691604b2aa/pkg/providers/v1/aws.go#L5007
cloud.ec2 is a v1.EC2 interface 
https://github.com/kubernetes/cloud-provider-aws/blob/58bc609b9f79e1283b8896011f1590691604b2aa/pkg/providers/v1/aws.go#L334
and the "real"/non-mock implementation of it by awsSdkEC2 is
https://github.com/kubernetes/cloud-provider-aws/blob/58bc609b9f79e1283b8896011f1590691604b2aa/pkg/providers/v1/aws.go#L964
which calls s.ec2.DescribeInstances (now ec2 is an aws-sdk-go ec2.EC2 not a v1.EC2/awsSdkEC2)
https://github.com/kubernetes/cloud-provider-aws/blob/58bc609b9f79e1283b8896011f1590691604b2aa/pkg/providers/v1/aws.go#L731

PR #277, opened 2021-09-22
Add test for DescribeInstances

nckturner:
* Add test for DescribeInstances (the lower level wrapper around the API call).
* Use interface in awsSdkEC2 to allow for mock.
* Check `len(request.InstanceIds) == 0` to account for both empty structs and nil pointers.

See https://github.com/kubernetes/cloud-provider-aws/pull/274 for the change and https://github.com/kubernetes/cloud-provider-aws/issues/269 for some context.  

**What type of PR is this?**

/kind cleanup

Add a test and make a comparison more resilient. 

```release-note
NONE
```

### DIFF
--- a/pkg/providers/v1/aws_test.go
+++ b/pkg/providers/v1/aws_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/aws/aws-sdk-go/service/elb"
 	"github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


wongma7:
`
/home/prow/go/src/k8s.io/cloud-provider-aws/pkg/providers/v1/aws_test.go:3756:6: don't use underscores in Go names; func expectDescribeInstances_EmptyInput should be expectDescribeInstancesEmptyInput `

pkg/providers/v1/aws_test.go
```diff
@@ -3741,3 +3742,86 @@ func TestGetZoneByProviderIDForFargate(t *testing.T) {
...
+					&ec2.DescribeInstancesInput{
+						MaxResults: aws.Int64(3),
+					},
+				).Return(
+					&ec2.DescribeInstancesOutput{},
+					nil,
+				)
+			},
+			false,
+		},
```
wongma7, line 3811: case where instance ids are provided ?

pkg/providers/v1/aws_test.go
```diff
@@ -3741,3 +3742,86 @@ func TestGetZoneByProviderIDForFargate(t *testing.T) {
...
+
+func newMockedEC2API() *MockedEC2API {
+	return &MockedEC2API{}
+}
+
+func TestDescribeInstances(t *testing.T) {
+	tests := []struct {
+		name    string
+		input   *ec2.DescribeInstancesInput
+		expect  func(interface{})
```
wongma7, line 3764: why is it interface{}? what is expect supposed to do? what would a failure look like? if in every test case I tell DescribeInstances what to take as input and what to return, what is being tested?

nckturner: I think I'm missing an `AssertExpectations` call in the test (I was a little confused as to whether I needed it while writing the test, because the test does fail if DescribeInstances is not called as expected).   Let me see if I can determine if its needed and clear up the logic.

wongma7: Documentation for On: https://pkg.go.dev/github.com/stretchr/testify/mock#Call.On
so I am telling it what to return when given so and so input. And I think ultimately I have to assert that this call happened at all by doing https://pkg.go.dev/github.com/stretchr/testify/mock#Mock.AssertExpectations 

(new to this package, slightly more familiar with gomock)

wongma7: we dont need Assert since Called basically does it for us . https://pkg.go.dev/github.com/stretchr/testify/mock?utm_source=godoc#Mock.Called

nckturner: Right, that's the reason that I didn't have an explicit call to AssertExpectations -- because I was seeing tests fail if I called the expected functions with unexpected arguments.  That being said, I think we do want an explicit call to `AssertExpectations` because of [the missing case](https://github.com/stretchr/testify/blob/e209ca88af1919e710960205e887e57db33c9be0/mock/mock.go#L529-L531). If there is an assert, and a missing call, it will only be caught by the explicit `AssertExpectations`, whereas `Called` will catch calls with improper arguments.

PR #283, opened 2021-11-04
use sts regional endpoint for aws session

jyotimahapatra:
**What type of PR is this?**
> /kind bug

STS [suggests](https://docs.aws.amazon.com/general/latest/gr/sts.html) using regional endpoints. 
```
AWS recommends using Regional STS endpoints to reduce latency, build in redundancy, and increase session token validity
```
The PR creates a way for CCM users to set an env var `USE_REGIONAL_STS` to use regional endpoints. The default when the flag is unset is false. Only when the flag is explicitly `true`/`TRUE` the regional endpoint is chosen.

Fixes #

```release-note
CCM users can set an env var `USE_REGIONAL_STS` to use regional endpoints for sts calls. The default when the flag is unset is false. Only when the flag is explicitly `true`/`TRUE` the regional endpoint is chosen.

```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"os"
 	"path"
 	"regexp"
 	"sort"
@@ -287,6 +288,10 @@ const (
 
 	// privateDNSNamePrefix is the prefix added to ENI Private DNS Name.
 	privateDNSNamePrefix = "ip-"
+
+	// Environment Variable to create sts session with regional endpoint
+	// If unset or false, uses global sts endpoint
+	useRegionalSts = "USE_REGIONAL_STS"
 )
 
 const (
@@ -1208,8 +1213,12 @@ func init() {
 			return nil, fmt.Errorf("unable to validate custom endpoint overrides: %v", err)
 		}
 
+		awsSessionConfig, err := getSessionConfig(&awsSDKProvider{cfg: cfg})
+		if err != nil {
+			return nil, fmt.Errorf("unable to create aws session: %v", err)
+		}
 		sess, err := session.NewSessionWithOptions(session.Options{
-			Config:            aws.Config{},
+			Config:            *awsSessionConfig,
 			SharedConfigState: session.SharedConfigEnable,
 		})
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/providers/v1/aws.go
```diff
@@ -287,6 +288,10 @@ const (
 
 	// privateDNSNamePrefix is the prefix added to ENI Private DNS Name.
 	privateDNSNamePrefix = "ip-"
+
+	// Environment Variable to create sts session with regional endpoint
+	// If unset or false, uses global sts endpoint
+	useRegionalSts = "USE_REGIONAL_STS"
```
nckturner, line 294: I think the switch should be the opposite because 1. we want everyone to use regional STS endpoints and 2. we are still publishing alpha containers, and we can document in the upgrade notes that if you have STS regional endpoints disabled in your AWS account, you need to change this setting.  Also, I think it should be a flag instead of an environment variable (and/or a cloud config option).  We can do something more complicated, like a feature gate of sorts that initially is opt-in but when it moves to beta becomes opt-out (I.e. the feature gate is called "UseRegionalSTS" and is false in alpha, but true in beta, and eventually is removed and cannot be opted out of).  Whatever we end up with though, we should have users opt-out of regional endpoints, rather than opt-in to them, because STS wants to move traffic off the global endpoint.

jyotimahapatra: i agree with most of the things. Few things i wanted to discuss about how we make the changes that could affect users. 
Since sts has an ability to disable regional endpoints, shifting the default to regional could be a breaking change. I propose that we keep the original behavior intact and deprecate. In the next release of ccm, we could shift the default to regional and put a release note(included in PR description)

From my investigation so far, using a flag is difficult because of the how the cloud provider is instatiated. The kube controller passes no configurations while initializing the cloud provider. [Looking](https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws.go#L1201) at the signature `cloudprovider.RegisterCloudProvider(ProviderName, func(config io.Reader) (cloudprovider.Interface, error) {` there is no other config apart from CloudConfig file reader. It would take a refactoring in k/k to pass additional flagset. Do you find this explanation correct? If so we'll have to use an env var until flagset can be passed.

wongma7: flag can be added to https://github.com/kubernetes/cloud-provider-aws/blame/master/cmd/aws-cloud-controller-manager/main.go no problem

nckturner:
Replaced by https://github.com/kubernetes/cloud-provider-aws/pull/313

/close

PR #293, opened 2021-12-08
Tag on create

olemarkus:
```release-note
Target groups, security groups and listeners are now tagged on creation rather than as separate API calls.
```


rifelpet:
LGTM though I'm curious if all AWS partitions support tag-on-create for these resource types

olemarkus:
At least AWS Load Balancer Controller use tag on create for these resources.
CCM also use them on the other resources it manages.

nckturner:
I'm unable to find information about ELB tag on create status across all regions, but if AWS Load Balancer Controller uses it then I'm comfortable using it.

olemarkus:
As I understand it from [this code](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/deploy/elbv2/load_balancer_manager.go#L57-L62) tags are applied on create.

PR #295, opened 2021-12-28
Use go-runner as base image

wongma7:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: To collect/persist logs (versus just emitting to stdout/stderr and letting container runtime deal with them) go-runner should be used. kubernetes already uses this image as base for apiserver, scheduler, and KCM https://github.com/kubernetes/kubernetes/blob/release-1.23/build/common.sh#L94. Since cloud provider is a "kubernetes component" basically split off from KCM it makes sense to use this image as base for cloud provider as well. More details are here: https://github.com/kubernetes/release/tree/master/images/build/go-runner. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,16 +14,15 @@
 ##                               BUILD ARGS                                   ##
 ################################################################################
 # This build arg allows the specification of a custom Golang image.
-ARG GOLANG_IMAGE=golang:1.17.4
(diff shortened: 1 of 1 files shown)


olemarkus:
Been thinking if kops should also log CCM to disk. This will certainly help with that.

/lgtm

PR #298, opened 2022-01-21
Update charts for v1.23.0-alpha.0

nckturner:
**What type of PR is this?**
/kind cleanup

/hold
* Wait for https://github.com/kubernetes/k8s.io/pull/3287
* Check image is in us.gcr.io/k8s-artifacts-prod/provider-aws/cloud-controller-manager

```release-note
NONE
```


nckturner:
If we are going to update charts on release branches (i.e. release-1.22) then it might make sense to major version bump the chart version with Kubernetes major version upgrades, so that we can avoid conflicting chart versions.

PR #299, opened 2022-01-21
Update charts and examples on 1.22 release branch to v1.22.0-alpha.1

nckturner:
**What type of PR is this?**

/kind documentation

* Merging updated chart versions into 1.22 release branch.

```release-note
NONE
```


nckturner:
@wongma7 and I decided not to do this, rather just release the helm chart from the master branch and ensure it is compatible with all ccm versions that we support.

PR #303, opened 2022-01-28
canonicalize ipv6 metadata ip address

jyotimahapatra:
Signed-off-by: Jyoti Mahapatra <jyotima@amazon.com>

**What type of PR is this?**
/kind bug

The ipv6 address from imds metadata endpoint is not a canonicalized format. This caused an issue in kubelet https://github.com/kubernetes/kubernetes/issues/107735 . Handling ip as string is error prone, hence this PR parses the ip before passing it to the kubelet.
https://github.com/kubernetes/kubernetes/issues/107735

Fixes #

```release-note

```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -1691,7 +1691,7 @@ func (c *Cloud) ipv6AddressesFromMetadata() ([]v1.NodeAddress, error) {
 			if internalIPv6 == "" {
 				continue
 			}
-			addresses = append(addresses, v1.NodeAddress{Type: v1.NodeInternalIP, Address: internalIPv6})
+			addresses = append(addresses, v1.NodeAddress{Type: v1.NodeInternalIP, Address: net.ParseIP(internalIPv6).String()})
 			break
 		}
 	}
--- a/pkg/providers/v1/aws_fakes.go
+++ b/pkg/providers/v1/aws_fakes.go
@@ -373,6 +373,14 @@ func (m *FakeMetadata) GetMetadata(key string) (string, error) {
 			}
 		}
 
+		if len(keySplit) == 5 && keySplit[4] == "ipv6s" {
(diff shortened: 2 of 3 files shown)


anguslees (approved):

anguslees:
lgtm.

My only question is whether we should do the same for IPv4 IPs too.  Eg: what happens if we get "10.00.0.1" in local-ipv4s?

olemarkus:
Is this only affecting IPs fetched from metadata? These codepaths are never used in external cloud providers. The IPs are always fetched from the EC2 API since the cloud provider doesn't run on all nodes.

Think the plan is to remove the metadata functions.

jyotimahapatra:
> 10.00.0.1

`net.ParseIP("10.00.0.1")` fails and passes back `nil` which has the potential to break existing usecases where such erroneous ips are passed as strings. Im certain ipv6 change is not going to break any usecases and hence the focussed change.

Do you think we could break any scenarios on canonicalizing ipv4 ips? May be we should do it as part of a minor k8s release?

jyotimahapatra:
> These codepaths are never used in external cloud providers. The IPs are always fetched from the EC2 API since the cloud provider doesn't run on all nodes.

Got it..in that case is this PR relevant for anymore? If we're getting rid of metadata functions?

nckturner:
Correct, I haven't looked at the PR yet, but if it only touches code called by kubelet, then we can safely not merge this.

jyotimahapatra:
The change is in the code path here https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws.go#L1509 in `NodeAddresses` and i feel this is called from the kubelet only(since kubelet deals with nodes). 

I'll check https://github.com/kubernetes/cloud-provider/blob/master/controllers/node/node_controller.go#L320-L366 to make sure the node controller does not go through this path

PR #309, opened 2022-02-28
[WIP] Issue 124 - Implement AWS GetLoadBalancerName, get name from annotations

### LINKED ISSUE #124: Friendly Load Balancer Names and Adopting Existing Load Balancers

Superbasil3:
**What type of PR is this?**
/kind feature

This implement the GetLoadBalancerName function.
What it does, is implement to naming of AWS load-balancer as described here : https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/guide/service/annotations.md#load-balancer-name

Fixes #124

I test that the annotations name is valid against AWS rules ( less than 32 characters, only alphanumeric and hyphen (except first and last for hyphen).
If the annotation is present and is not valid, I throw an error. I don't know if I should default the the current generation.
Else, if no annotation, then I default as before.

```release-note
This change allows user to specify with annotation human readable name that will be displayed on AWS load-balancer page, when creating the associated Ingress.
```


docs/service_controller.md
```diff
@@ -26,4 +26,4 @@ The service controller is responsible for watch for service and node object chan
 | service.beta.kubernetes.io/aws-load-balancer-ssl-negotiation-policy            | -                                   | ELBSecurityPolicy-2016-08 | Specifies SSL negotiation settings for the HTTPS/SSL listeners of your load balancer. Defaults to the default ELB policy. |
 | service.beta.kubernetes.io/aws-load-balancer-ssl-ports                         | Comma-separated list                                   | *   | Specifies a comma-separated list of ports that will use SSL/HTTPS listeners. Defaults to all. |
 | service.beta.kubernetes.io/aws-load-balancer-type                              | [nlb]                                   | -   | Indicates the type of Load Balancer. The only valid value is nlb.  Leaving this field blank is equivalent to selecting ELB. |
-
+| service.beta.kubernetes.io/aws-load-balancer-name                              | regexp(^[a-zA-Z0-9-]{0.32}$)            | -   | Indicates the type name of AWS Load Balancer Ressource, which should be a maximum of 32 alphanumeric characters including hyphens string. Leaving this field blank is using the default implementation ("a" + uuid(31), without hyphens). |
```
jaypipes, line 29: Recommend rewording the description to:

```
The name of the AWS Load Balancer resource. If this annotation is missing, new load balancers
will have auto-generated names consisting of 'a' + uuid(31), without hyphens
```

pkg/providers/v1/aws.go
```diff
@@ -4371,8 +4375,24 @@ func (c *Cloud) GetLoadBalancer(ctx context.Context, clusterName string, service
 
 // GetLoadBalancerName is an implementation of LoadBalancer.GetLoadBalancerName
 func (c *Cloud) GetLoadBalancerName(ctx context.Context, clusterName string, service *v1.Service) string {
-	// TODO: replace DefaultLoadBalancerName to generate more meaningful loadbalancer names.
-	return cloudprovider.DefaultLoadBalancerName(service)
+	if val := service.Annotations[ServiceAnnotationLoadBalancerName]; val != nil {
+		r := regexp.MustCompile(`^[a-zA-Z0-9-]{0,32}`)
```
jaypipes, line 4379: Instead of compiling the regex every time that `*Cloud.GetLoadBalancerName()` is called, consider placing this as a package-level var at the top of the file:

```go
var lbNameRegex = regexp.MustCompile(`^[a-zA-Z0-9-]{0,32}`)
```

and simply refer to it here...

```go
if lbNameRegex.MatchString(val) ...
```

pkg/providers/v1/aws.go
```diff
@@ -4371,8 +4375,24 @@ func (c *Cloud) GetLoadBalancer(ctx context.Context, clusterName string, service
...
+	}
+
+	//AWS requires that the name of a load balancer is shorter than 32 bytes.
+	//Default implementation, currently deprecated
+	ret := "a" + string(service.UID)
+	ret = strings.Replace(ret, "-", "", -1)
+	if len(ret) > 32 {
+		ret = ret[:32]
+	}
+	return ret
```
jaypipes, line 4395: Instead of the above code, simply fallback to:

```go
return cloudprovider.DefaultLoadBalancerName(service)
```

pkg/providers/v1/aws.go
```diff
@@ -4371,8 +4375,24 @@ func (c *Cloud) GetLoadBalancer(ctx context.Context, clusterName string, service
...
 }
```
jaypipes, line 4396: Please add a unit test similar to https://github.com/kubernetes/cloud-provider-aws/blob/f4b0643c854739ef2ba954881c0ac569a3adb091/pkg/providers/v1/aws_test.go#L1145 to validate this new functionality.

jaypipes (changes requested):
@Superbasil3 thank you for the contribution! Seems relatively straightforward to me, though I've added some suggestions and requests inline.

docs/service_controller.md
Superbasil3: Description updated.

pkg/providers/v1/aws.go
Superbasil3: Update the regex as a package-level var.

pkg/providers/v1/aws.go
Superbasil3: Hello @jaypipes, for this one, I was wondering what to do exactly, on the parent project, the method is marked as deprecated, so should we really use the method, or implement it instead here ? 
https://github.com/kubernetes/cloud-provider/blob/c819ca6bd71c22c47e38c26461a0a0db96d8b9b0/cloud.go#L84

pkg/providers/v1/aws.go
Superbasil3: Added some unit tests to validate all the error case

Superbasil3:
Hello @jaypipes, if you have some time, can you reviews the test I added ? 
Thanks a lot,
Kind regards,

Thomas

jaypipes (approved):
:+1: this PR is good to go, IMHO. Short and contained.

Superbasil3:
Hello @jaypipes ,
I'm sorry, there was an issue on the lint, that made the test fail. I added a commit to resolve this issue.
Thanks again for your time for the review.
Kind regards,

jaypipes (approved):
Yep, still looks great, thank you @Superbasil3 :)

pkg/providers/v1/aws.go
```diff
@@ -4410,8 +4415,22 @@ func (c *Cloud) GetLoadBalancer(ctx context.Context, clusterName string, service
 
 // GetLoadBalancerName is an implementation of LoadBalancer.GetLoadBalancerName
 func (c *Cloud) GetLoadBalancerName(ctx context.Context, clusterName string, service *v1.Service) string {
-	// TODO: replace DefaultLoadBalancerName to generate more meaningful loadbalancer names.
-	return cloudprovider.DefaultLoadBalancerName(service)
+	if val := service.Annotations[ServiceAnnotationLoadBalancerName]; val != "" {
```
kishorj, line 4418: This doesn't handle existing load balancers provisioned for the service and introduces a potential for leaked LB resources. For example, if you create a service without a custom lb name, controller first provisions a LB with the "a..." name, then later you specify the custom name, controller doesn't clean up the previously created load balancer resource. Previous LB resource get leaked everytime you edit the custom name as well.

kishorj (commented):
Overriding the GetLoadBalancerName is not sufficient to achieve custom lb name on this controller.

pkg/providers/v1/aws.go
jaypipes: @kishorj OK, so because the existing behaviour is static -- you cannot change the LB name from the service UID -- that essentially means that we're stuck with non-meaningful LB names, yes? Unless we check for a previous name and delete the LB with the previous name after creating a new LB with the new name. In other words, we would want cloud-provider-aws to be built like any "normal" reconciler ... grab the latest observed state, compare against the desired state, and reconcile the observed to match the desired. The fact that cloud-provider-aws apparently isn't designed in this way makes changes like this one basically dead in the water, no? Sounds like a complete redesign of cloud-provider-aws is needed... :(

PR #312, opened 2022-03-02
Get e2e tests working in prow

nckturner:
**What type of PR is this?**

/kind failing-test

Fixes failing e2e test.

* Create test run directory earlier to avoid failure
* Install e2e prereqs

Fixes #

```release-note
NONE
```

### DIFF
--- a/hack/e2e/run.sh
+++ b/hack/e2e/run.sh
@@ -29,10 +29,10 @@ test_output_root="${output}/test"
 test_run="${test_output_root}/${test_run_id}"
 
 # Set in Makefile
-MAKE_VERSION="${MAKE_VERSION:-}"
-MAKE_IMAGE="${MAKE_IMAGE:-}"
+BUILD_VERSION="${BUILD_VERSION:-}"
+BUILD_IMAGE="${BUILD_IMAGE:-}"
 
-if [[ -z "${MAKE_VERSION}" || -z "${MAKE_IMAGE}" ]]; then
+if [[ -z "${BUILD_VERSION}" || -z "${BUILD_IMAGE}" ]]; then
     echo "$0: Execute with 'make test-e2e'"
     exit 1
 fi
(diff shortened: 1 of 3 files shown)


hack/e2e/run.sh
```diff
@@ -155,9 +149,9 @@ if [[ "${UP}" = "yes" ]]; then
 fi
 
 pushd ./tests/e2e
-${GINKGO_BIN} . -p -nodes="${GINKGO_NODES}" -v --focus="${GINKGO_FOCUS}" --skip="${GINKGO_SKIP}" "" -- -kubeconfig="${KUBECONFIG}" -report-dir="${test_run}" -gce-zone="${ZONES%,*}" "${EXPANDED_TEST_EXTRA_FLAGS}"
+ginkgo . -p -nodes="${GINKGO_NODES}" -v --focus="${GINKGO_FOCUS}" --skip="${GINKGO_SKIP}" "" -- -kubeconfig="${KUBECONFIG}" -report-dir="${test_run}" -gce-zone="${ZONES%,*}" "${EXPANDED_TEST_EXTRA_FLAGS}"
 popd
 
 if [[ "${DOWN}" = "yes" ]]; then
-    kops delete cluster --name "${CLUSTER_NAME}" --yes
+    ${test_run}/kops delete cluster --name "${CLUSTER_NAME}" --yes
```
wongma7, line 156: how did the binary get here? is kops also an e2e-tool?

nckturner: Good question, its downloaded by the kubetest2 kops deployer: https://github.com/kubernetes/kops/blob/master/tests/e2e/pkg/kops/download.go#L32-L55

wongma7: should we call kubetest2 --down (or whatever the right arguments are) then instead, so we don't have to care about that implementation detail?

PR #313, opened 2022-03-03
add sts regional endpoint logic

prasita123:
**What type of PR is this?**
> kind bug

STS suggests using regional endpoints. ([source](https://docs.aws.amazon.com/general/latest/gr/sts.html))

> AWS recommends using Regional STS endpoints to reduce latency, build in redundancy, and increase session token validity.

The PR uses sts regional endpoints, fetched using the metadata, instead of global endpoints.

Fixes #

```release-note
Updates to the cloud-controller-manager to use regional endpoint instead of global endpoint, as AWS recommends using Regional STS endpoints to reduce latency, build in redundancy, and increase session token validity.
```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -1209,8 +1209,12 @@ func init() {
 			return nil, fmt.Errorf("unable to validate custom endpoint overrides: %v", err)
 		}
 
+		regionName, err := getRegionFromMetadata(cfg)
+		if err != nil {
+			return nil, err
+		}
 		sess, err := session.NewSessionWithOptions(session.Options{
-			Config:            aws.Config{},
+			Config:            *aws.NewConfig().WithRegion(regionName).WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint),
 			SharedConfigState: session.SharedConfigEnable,
 		})
 		if err != nil {
@@ -5129,4 +5133,29 @@ func (c *Cloud) describeNetworkInterfaces(nodeName string) (*ec2.NetworkInterfac
 		return nil, fmt.Errorf("multiple interfaces found with same id %q", eni.NetworkInterfaces)
 	}
 	return eni.NetworkInterfaces[0], nil
+}
+
+func getRegionFromMetadata(cfg *CloudConfig) (string, error) {
+	klog.Infof("Get AWS region from metadata client")
+	metadata, err := newAWSSDKProvider(nil, cfg).Metadata()
+	if err != nil {
+		return "", fmt.Errorf("error creating AWS metadata client: %q", err)
+	}
+
+	err = updateConfigZone(cfg, metadata)
+	if err != nil {
+		return "", fmt.Errorf("unable to determine AWS zone from cloud provider config or EC2 instance metadata: %v", err)
+	}
+
+	zone := cfg.Global.Zone
+	if len(zone) <= 1 {
+		return "", fmt.Errorf("invalid AWS zone in config file: %s", zone)
+	}
+
+	regionName, err := azToRegion(zone)
+	if err != nil {
+		return "", err
+	}
+
+	return regionName, nil
 }
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -1214,13 +1214,13 @@ func init() {
 			return nil, fmt.Errorf("error creating AWS metadata client: %q", err)
 		}
 
-		regiondetails, err := getRegionFromMetadata(*cfg, metadata)
+		regionName, _, err := getRegionFromMetadata(*cfg, metadata)
 		if err != nil {
 			return nil, err
 		}
 
 		sess, err := session.NewSessionWithOptions(session.Options{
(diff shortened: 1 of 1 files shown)


pkg/providers/v1/aws.go
```diff
@@ -5130,3 +5134,28 @@ func (c *Cloud) describeNetworkInterfaces(nodeName string) (*ec2.NetworkInterfac
 	}
 	return eni.NetworkInterfaces[0], nil
 }
+
+func getRegionFromMetadata(cfg *CloudConfig) (string, error) {
```
wongma7, line 5138: since this is almost identical to the passage 

https://github.com/kubernetes/cloud-provider-aws/blob/8cdf51280cc494a9fd150d8b3c26706f1235c516/pkg/providers/v1/aws.go#L1300-L1317

could you refactor that to also use this function?

pkg/providers/v1/aws.go
```diff
@@ -5130,3 +5134,28 @@ func (c *Cloud) describeNetworkInterfaces(nodeName string) (*ec2.NetworkInterfac
...
+		return "", fmt.Errorf("invalid AWS zone in config file: %s", zone)
+	}
+
+	regionName, err := azToRegion(zone)
+	if err != nil {
+		return "", err
+	}
+
+	return regionName, nil
+}
```
wongma7, line 5161: newline is needed her.

Run ./hack/verify-gofmt.sh locally to verify.

ref: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/cloud-provider-aws/313/pull-cloud-provider-aws-check/1499213287563202560

pkg/providers/v1/aws.go
```diff
@@ -5130,3 +5130,34 @@ func (c *Cloud) describeNetworkInterfaces(nodeName string) (*ec2.NetworkInterfac
 	}
 	return eni.NetworkInterfaces[0], nil
 }
+
+type regionDetails struct {
+	regionName string
+	zone       string
+}
+
+func getRegionFromMetadata(cfg CloudConfig, metadata EC2Metadata) (*regionDetails, error) {
```
wongma7, line 5139: consider returning multiple return values like
```
(string, string, err)
region, zone, err 
```
instead. I like it better than a special struct

pkg/providers/v1/aws.go
```diff
@@ -1302,19 +1312,9 @@ func newAWSCloud(cfg CloudConfig, awsServices Services) (*Cloud, error) {
...
-
-	zone := cfg.Global.Zone
-	if len(zone) <= 1 {
-		return nil, fmt.Errorf("invalid AWS zone in config file: %s", zone)
-	}
-	regionName, err := azToRegion(zone)
-	if err != nil {
-		return nil, err
-	}
+	regiondetails, err := getRegionFromMetadata(cfg, metadata)
```
wongma7, line 1315: must check for err here

wongma7: I think the lack of err check causes TestNewAWSCloud to fail now after the refactor, try the test locally as well to verify .. https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/cloud-provider-aws/313/pull-cloud-provider-aws-test/1499235237555605504

pkg/providers/v1/aws.go
```diff
@@ -1302,19 +1312,7 @@ func newAWSCloud(cfg CloudConfig, awsServices Services) (*Cloud, error) {
...
-
-	zone := cfg.Global.Zone
-	if len(zone) <= 1 {
-		return nil, fmt.Errorf("invalid AWS zone in config file: %s", zone)
-	}
-	regionName, err := azToRegion(zone)
-	if err != nil {
-		return nil, err
-	}
+	regionName, zone, err := getRegionFromMetadata(cfg, metadata)
```
wongma7, line 1315: err check missing here
	if err != nil {
		return nil, err
	}

prasita123:
tested the change on a 1.22 cluster (manifest `bd810fe3d5ba620f3d7cad5e4b40c11bc6efa1471d2e5075f459d29f48dca745`)
- validated the sts endpoint on a newly launched instance by terminating an existing node instance of the cluster
```
"tlsDetails": {
        "tlsVersion": "TLSv1.2",
        "cipherSuite": "ECDHE-RSA-AES128-GCM-SHA256",
        "clientProvidedHostHeader": "sts.us-west-2.amazonaws.com"
    }
```
https://tiny.amazon.com/a7l9nrne/IsenLink

pkg/providers/v1/aws.go
prasita123: added

wongma7:
/lgtm
but does it need a release note? maybe yes for the miniscule non-zero chance somebody has not enabled regional endpoint in opt-in region? if not put NONE in the field at the bottom of PR description

nckturner:
Yeah let's add one.  Which reminds me... we need some mechanism to get release notes out of this don't we?  I'm assuming that is buried in the kubernetes release automation but doesn't apply to other repositories.

wongma7:
not sure it is usable outside k/k but https://github.com/kubernetes/sig-release/blob/d30b2484d7b18f869dd7f560ef5fb52a3b536a32/release-team/role-handbooks/release-notes/editing-flow.md#workflow-operation . 

@prasita123 plz draft a release note, we can review that as well

PR #319, opened 2022-03-17
Fix route controller create/delete spam: use instanceIDToNodeName in case node name != private DNS

### LINKED ISSUE #318: Instance IDs as node names causes route controller to spam create/delete route
**What happened**:

in kops 1.23 if external CCM is enabled nodes are created with instance ID.
- this breaks listroutes which assumes nodenames are privateDNS https://github.com/kubernetes/cloud-provider-aws/blob/58bc609b9f79e1283b8896011f1590691604b2aa/pkg/providers/v1/aws_routes.go#L70
- this may break checkIfAttachedToNode which also calls mapInstanceToNodeName https://github.com/kubernetes/cloud-provider-aws/blob/58bc609b9f79e1283b8896011f1590691604b2aa/pkg/providers/v1/volumes.go#L110 (need to test)

**What you expected to happen**: instace ids as node names is supproted

**How to reproduce it (as minimally and precisely as possible)**: create 1.23 kops cluster https://github.com/kubernetes/kops/issues/13376

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

/kind bug

wongma7:
**What type of PR is this?**
/kind bug

The function `mapInstanceToNodeName` assumes that node names are always equal to private dns names. This assumption does not always hold, specifically in 1.23 kops clusters with CCM enabled where  node names are overridden to be instance IDs.

I added `instanceIDToNodeName` as a replacement. It relies on an indexer in the node informer to find nodeName by instance ID. This solution might be heavy-handed, I don't know how expensive exactly it is to have this indexer. If we assume that node names are going to be limited to either private dns or instance ids (or fargate?) then the solution could be simpler: it could check if a node with name equal to instance id exists, as suggested here https://github.com/kubernetes/kops/issues/13376#issuecomment-1069591289, and if not then fall back to assuming that node names are private dns names. Open to suggestions of course...

I tested this in my own cluster, I'll need to work on unit tests. And am considering how we can test this end to end or if such a test is necessary. We only caught this issue in the first place because AWS alerted us that we were making an unusual # of calls.

Test:
1. delete routes with destination 100.96.0.0/24 and 100.96.1.0/24 in VPC console
2. check logging of cloud controller: ~ $ k logs -n kube-system aws-cloud-controller-manager-xnmj5
```
I0317 21:29:45.399248       1 route_controller.go:194] Creating route for node i-001c99c6268df2574 100.96.1.0/24 with hint bd6fb3d1-708e-45ef-8019-01090827d461, throttled 10.433µs
I0317 21:29:45.399295       1 route_controller.go:194] Creating route for node i-001ab2674ef942145 100.96.0.0/24 with hint 24e777a5-cba7-4079-82c6-57d215eb4865, throttled 5.411µs
I0317 21:29:46.129790       1 route_controller.go:214] Created route for node i-001c99c6268df2574 100.96.1.0/24 with hint bd6fb3d1-708e-45ef-8019-01090827d461 after 730.545279ms
I0317 21:29:46.315747       1 route_controller.go:214] Created route for node i-001ab2674ef942145 100.96.0.0/24 with hint 24e777a5-cba7-4079-82c6-57d215eb4865 after 916.451209ms
```
3. check logging of cloud controller 5 mins later, there are no unnecessary Deletes or Creates:  ~ $ k logs -n kube-system aws-cloud-controller-manager-xnmj5

Fixes #318

```release-note
Fix route controller making unnecessary CreateRoute/DeleteRoute AWS API requests in case node names are not private DNS names
```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -818,12 +818,32 @@ func (p *awsSDKProvider) getCrossRequestRetryDelay(regionName string) *CrossRequ
 	return delayer
 }
 
+// InstanceIDIndexFunc indexes based on a Node's instance ID found in its spec.providerID
+func InstanceIDIndexFunc(obj interface{}) ([]string, error) {
+	node, ok := obj.(*v1.Node)
+	if !ok {
+		return []string{""}, fmt.Errorf("%+v is not a Node", obj)
+	}
+	if node.Spec.ProviderID == "" {
+		// provider ID hasn't been populated yet
+		return []string{""}, nil
+	}
+	instanceID, err := KubernetesInstanceID(node.Spec.ProviderID).MapToAWSInstanceID()
+	if err != nil {
+		return []string{""}, fmt.Errorf("error mapping node %q's provider ID %q to instance ID: %v", node.Name, node.Spec.ProviderID, err)
+	}
+	return []string{string(instanceID)}, nil
+}
+
 // SetInformers implements InformerUser interface by setting up informer-fed caches for aws lib to
 // leverage Kubernetes API for caching
 func (c *Cloud) SetInformers(informerFactory informers.SharedInformerFactory) {
 	klog.Infof("Setting up informers for Cloud")
 	c.nodeInformer = informerFactory.Core().V1().Nodes()
 	c.nodeInformerHasSynced = c.nodeInformer.Informer().HasSynced
+	c.nodeInformer.Informer().AddIndexers(cache.Indexers{
+		"instanceID": InstanceIDIndexFunc,
+	})
 }
 
 func (p *awsSDKProvider) Compute(regionName string) (EC2, error) {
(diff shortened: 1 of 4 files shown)


pkg/providers/v1/aws.go
```diff
@@ -5028,7 +5044,7 @@ func isFargateNode(nodeName string) bool {
 	return strings.HasPrefix(nodeName, fargateNodeNamePrefix)
 }
 
-func (c *Cloud) nodeNameToProviderID(nodeName types.NodeName) (InstanceID, error) {
+func (c *Cloud) nodeNameToInstanceID(nodeName types.NodeName) (InstanceID, error) {
```
wongma7, line 5047: the provider id / instance id change is not super relevant, will put in separate commit . but I made it cuz strictly speaking all instance ids are provider ids but provider ids are not instance ids (since provider ids should have prefix aws://region or something) so this helped with m y understanding personally ...

wongma7: will fix unit tests due to rename as well

olemarkus:
This looks good to me. Like the indexer approach.

pkg/providers/v1/aws_routes.go
```diff
@@ -112,9 +112,13 @@ func (c *Cloud) ListRoutes(ctx context.Context, clusterName string) ([]*cloudpro
 		// Capture instance routes
 		instanceID := aws.StringValue(r.InstanceId)
 		if instanceID != "" {
-			instance, found := instances[instanceID]
+			_, found := instances[instanceID]
```
nckturner, line 115: Do we need this check or can we remove the `instances, err := c.getInstancesByIDs(instanceIDs)` entirely?

pkg/providers/v1/aws.go
```diff
@@ -5051,6 +5071,32 @@ func (c *Cloud) nodeNameToProviderID(nodeName types.NodeName) (InstanceID, error
...
+	if c.nodeInformerHasSynced == nil || !c.nodeInformerHasSynced() {
+		return "", fmt.Errorf("node informer has not synced yet")
+	}
+
+	_, err := c.nodeInformer.Lister().Get(string(instanceID))
+	if err == nil {
+		return types.NodeName(instanceID), nil
+	}
+
+	nodes, err := c.nodeInformer.Informer().GetIndexer().IndexKeys("instanceID", string(instanceID))
```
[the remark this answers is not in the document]
wongma7: indexer should work for all cases, the Lister Get is redundant and I was thinking of it as a fast path but in practice I don't think it's any faster than Informer().GetIndexer().IndexKeys, so we an remove.

The Lister  Get would error if the node name is NOT equal to instance ID e.g. if the node name is still set to private dns.

pkg/providers/v1/aws_routes.go
wongma7: I don't know, I didn't think too much of it bc I think it's out of scope of this PR.

But I think the check is wrong  because ListRoutes should return all routes in the cluster's table and let the route controller decide whether it should be deleted. Because right now if there is a zombie route (say a node gets terminated while route controller is down) and then you hit the warning, the zombie will never get cleaned up?

but again I think it's out of scope so I'd rather fix this later

nckturner: > the zombie will never get cleaned up?

This is what I was wondering about, and I agree I'm pretty sure that check was only there due to necessity and we can remove it now.  The route controller can remove routes if the nodes don't exist: https://github.com/kubernetes/cloud-provider/blob/master/controllers/route/route_controller.go#L236-L257

PR #324, opened 2022-03-29
Use Makefile in cloudbuild

nckturner:
**What type of PR is this?**

/kind cleanup

* We need to use the makefile in cloudbuild so that we can have more
  control over our versions.  Currently we use GIT_TAG but we want
  it to always be semver compatible (expected by component-base)
  and we want to be able to filter out the helm chart tags
  (currently done by the Makefile but not by cloudbuild).

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,11 @@ IMAGE ?= amazon/cloud-controller-manager:$(VERSION)
 OUTPUT ?= $(shell pwd)/_output
 INSTALL_PATH ?= $(OUTPUT)/bin
 LDFLAGS ?= -w -s -X k8s.io/component-base/version.gitVersion=$(VERSION)
+PLATFORMS ?= linux/amd64,linux/arm64
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -50,26 +55,29 @@ ecr-credential-provider.exe: $(wildcard ./cmd/ecr-credential-provider/*.go)
...
 
-.PHONY: docker-build
-docker-build:
-	docker buildx build --output=type=registry \
-		--build-arg LDFLAGS=$(LDFLAGS) \
+.PHONY: docker-build-push
+docker-build-push:
+	docker buildx build \
+		--output=type=registry \
+		--build-arg VERSION=$(VERSION) \
```
olemarkus, line 78: I am wondering a bit about using `VERSION` here. That means that the staging versions will keep overriding the _last_ tag that was made in git. I would rather use what would be the _next_ tag. Either that or create a patch suffix.

PR #328, opened 2022-04-02
Add environment and securityContexts

### LINKED ISSUE #327: helm chart: Using IRSA for cloud-provider-aws

Hi, I would like to discuss if its useful to use [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) instead Instance profiles cloud-provider-aws. I would like reduce the permissions of an instance profile to an absolute minimum and use dedicated roles for each component.

In case this make sense, I can do a pull request to extend the helm chart.

Since control plane and non worker needs a [different set of permissions](https://cloud-provider-aws.sigs.k8s.io/prerequisites/), I suggest to split the existing DaemonSet into two. One for the control plane, one for worker node. Then I'm able to pass different service account tokens to different DaemonSets.

Before coding it, I would like ask, if someone is interest into it.

**Why is this needed**:
Increased security and traceability of IAM permissions. If the instance profile has close to zero permissions, no can abuse the permissions.

/kind feature

jkroepke:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Support IRSA for CCM and securityContext for additional hardening.

Fixes #327 

```release-note
chart: Expose daemonset env vars to support IRSA
```

### DIFF
--- a/charts/aws-cloud-controller-manager/values.yaml
+++ b/charts/aws-cloud-controller-manager/values.yaml
@@ -105,6 +105,13 @@ resources:
     cpu: 200m
   #   memory: 300Mi
 
+# env -- Pod environment variables
+env: {}
+# securityContext -- Container Security Context.
+securityContext: {}
+# podSecurityContext -- Pods Security Context.
+podSecurityContext: {}
+
 # tolerations -- List of node taints to tolerate (requires Kubernetes >= 1.6).
 tolerations:
 - key: node.cloudprovider.kubernetes.io/uninitialized
--- a/charts/aws-cloud-controller-manager/Readme.md
+++ b/charts/aws-cloud-controller-manager/Readme.md
@@ -1,6 +1,6 @@
-# AWS Cloud-Controler-Manager Helm Chart
+# AWS cloud-controller-manager Helm Chart
 
-Installs the [aws cloud-controller-manager ](https://github.com/kubernetes/cloud-provider-aws).
+Installs the [aws cloud-controller-manager](https://github.com/kubernetes/cloud-provider-aws).
 
 ## Get Repo Info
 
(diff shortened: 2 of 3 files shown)


charts/aws-cloud-controller-manager/values.yaml
```diff
@@ -105,6 +105,16 @@ resources:
     cpu: 200m
   #   memory: 300Mi
 
+# env -- Pod environment variables
+env: {}
+
+hostNetworking: false
```
olemarkus, line 111: This won't work. CCM in many cases need to work before CNI and therefore must use the host's network.

jkroepke: Hey @olemarkus 

please correct me if I'm wrong, but I'm only declare the current default here. I will remove it from PR, since it's out of scope.

Anyway, why CCM is needed for an CNI? https://github.com/aws/amazon-vpc-cni-k8s does not depend on CCM and  ignores any taints by default.

Please explain a case, I guess I miss one.

olemarkus: See https://kubernetes.io/docs/concepts/architecture/cloud-controller/#route-controller

Its fine to have this configurable, although I don't think that serves much purpose. The point of using IRSA with CCM is typically to remove privileges from the control plane role, not protect the control plane role (which is usually the case otherwise). Thus that it runs hostNetworking is not a problem.

While the VPC CNI helm chart tolerates all taints, it is not given that other CNIs' install instructions do so. Which means that the CNI DS and CCM DS block each other.

jkroepke: Thanks!

> Its fine to have this configurable, although I don't think that serves much purpose. 

It's already configurable on master, introduced in https://github.com/kubernetes/cloud-provider-aws/pull/173. The if condition leads to an undefined value by default.

olemarkus (approved):

jkroepke:
Hi, is everything fine or some changes are still required?

Issue #325, opened 2022-03-30
Add support for e2e test and non-retryable error in the tagging controller


nguyenkndinh:
**What would you like to be added**:
- Add e2e testing for tagging controller.
- Currently, all work items with errors from EC2 regarding tagging and untagging would be requeue to the workqueue. We would like to only requeue on transient errors.
Ref: https://github.com/kubernetes/cloud-provider-aws/pull/308#issuecomment-1082533927

**Why is this needed**:
- To have a comprehensive testing suite 
- To address a potential issue where failed jobs would stay in the queue indefinitely causing unnecessary calls to EC2.

/kind feature

saurav-agarwalla:
> Currently, all work items with errors from EC2 regarding tagging and untagging would be requeue to the workqueue. We would like to only requeue on transient errors.

This AI is now complete.

PR #334, opened 2022-04-05
[Issue #325] Added retry count to abort workitem after a few failed retries

nguyenkndinh:
**What type of PR is this?**
/kind feature

- Added a logic to abort work item after a certain number of retries in case EC2 responds with non-retryable errors.
- Added metrics collection to the controller.

NONE

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/tagging/metrics.go
+++ b/pkg/controllers/tagging/metrics.go
@@ -1,0 +1,56 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tagging
+
+import (
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+	"sync"
+)
+
+var register sync.Once
+
+var (
+	workItemDuration = metrics.NewHistogramVec(
+		&metrics.HistogramOpts{
+			Name:           "cloudprovider_aws_tagging_controller_work_item_duration_seconds",
+			Help:           "workitem latency of workitem being in the queue and time it takes to process",
+			StabilityLevel: metrics.ALPHA,
+		},
+		[]string{"latency_type"})
+
+	workItemError = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Name:           "cloudprovider_aws_tagging_controller_work_item_errors_total",
+			Help:           "any error in dequeueing the work queue and processing workItem",
+			StabilityLevel: metrics.ALPHA,
+		},
+		[]string{"error_type", "instance_id"})
+)
+
+// registerMetrics registers tagging-controller metrics.
+func registerMetrics() {
+	register.Do(func() {
+		legacyregistry.MustRegister(workItemDuration)
+		legacyregistry.MustRegister(workItemError)
+	})
+}
+
+func recordWorkItemLatencyMetrics(latencyType string, timeTaken float64) {
+	workItemDuration.With(metrics.Labels{"latency_type": latencyType}).Observe(timeTaken)
+}
+
+func recordWorkItemErrorMetrics(errorType string, instanceID string) {
+	workItemError.With(metrics.Labels{"error_type": errorType, "instance_id": instanceID}).Inc()
+}
--- a/pkg/controllers/tagging/tagging_controller.go
+++ b/pkg/controllers/tagging/tagging_controller.go
@@ -31,12 +31,30 @@ import (
 
 // workItem contains the node and an action for that node
 type workItem struct {
-	node   *v1.Node
-	action func(node *v1.Node) error
+	node           *v1.Node
+	action         func(node *v1.Node) error
+	requeuingCount int
+	enqueueTime    time.Time
 }
 
+const (
+	maxRequeuingCount = 9
+
+	// The label for depicting total number of errors a work item encounter and succeed
+	totalErrorsWorkItemErrorMetric = "total_errors"
+
+	// The label for depicting total time when work item gets queued to processed
+	workItemProcessingTimeWorkItemMetric = "work_item_processing_time"
+
+	// The label for depicting total time when work item gets queued to dequeued
+	workItemDequeuingTimeWorkItemMetric = "work_item_dequeuing_time"
+
+	// The label for depicting total number of errors a work item encounter and fail
+	errorsAfterRetriesExhaustedWorkItemErrorMetric = "errors_after_retries_exhausted"
+)
+
 // Controller is the controller implementation for tagging cluster resources.
-// It periodically check for Node events (creating/deleting) to apply appropriate
+// It periodically checks for Node events (creating/deleting) to apply/delete appropriate
 // tags to resources.
 type Controller struct {
 	nodeInformer coreinformers.NodeInformer
@@ -44,6 +62,7 @@ type Controller struct {
 	cloud        *awsv1.Cloud
 	workqueue    workqueue.RateLimitingInterface
 	nodesSynced  cache.InformerSynced
+
 	// Value controlling Controller monitoring period, i.e. how often does Controller
 	// check node list. This value should be lower than nodeMonitorGracePeriod
 	// set in controller-manager
@@ -71,15 +90,16 @@ func NewTaggingController(
 		return nil, err
 	}
 
+	registerMetrics()
 	tc := &Controller{
 		nodeInformer:      nodeInformer,
 		kubeClient:        kubeClient,
 		cloud:             awsCloud,
-		nodeMonitorPeriod: nodeMonitorPeriod,
 		tags:              tags,
 		resources:         resources,
 		workqueue:         workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "Tagging"),
 		nodesSynced:       nodeInformer.Informer().HasSynced,
+		nodeMonitorPeriod: nodeMonitorPeriod,
 	}
 
 	// Use shared informer to listen to add/update/delete of nodes. Note that any nodes
@@ -135,19 +155,42 @@ func (tc *Controller) process() bool {
 		workItem, ok := obj.(*workItem)
 		if !ok {
 			tc.workqueue.Forget(obj)
-			utilruntime.HandleError(fmt.Errorf("expected workItem in workqueue but got %#v", obj))
+			err := fmt.Errorf("expected workItem in workqueue but got %#v", obj)
+			utilruntime.HandleError(err)
 			return nil
 		}
 
-		err := workItem.action(workItem.node)
+		timeTaken := time.Since(workItem.enqueueTime).Seconds()
+		recordWorkItemLatencyMetrics(workItemDequeuingTimeWorkItemMetric, timeTaken)
+
+		instanceID, err := awsv1.KubernetesInstanceID(workItem.node.Spec.ProviderID).MapToAWSInstanceID()
 		if err != nil {
-			// Put the item back on the workqueue to handle any transient errors.
-			tc.workqueue.AddRateLimited(workItem)
-			return fmt.Errorf("error finishing work item '%v': %s, requeuing", workItem, err.Error())
+			err = fmt.Errorf("Error in getting instanceID for node %s, error: %v", workItem.node.GetName(), err)
+			utilruntime.HandleError(err)
+			return nil
+		}
+
+		err = workItem.action(workItem.node)
+
+		if err != nil {
+			if workItem.requeuingCount < maxRequeuingCount {
+				// Put the item back on the workqueue to handle any transient errors.
+				workItem.requeuingCount++
+				tc.workqueue.AddRateLimited(workItem)
+
+				recordWorkItemErrorMetrics(totalErrorsWorkItemErrorMetric, string(instanceID))
+				return fmt.Errorf("error processing work item '%v': %s, requeuing count %d", workItem, err.Error(), workItem.requeuingCount)
+			}
+
+			klog.Errorf("error processing work item '%v': %s, requeuing count exceeded", workItem, err.Error())
+			recordWorkItemErrorMetrics(errorsAfterRetriesExhaustedWorkItemErrorMetric, string(instanceID))
+		} else {
+			klog.Infof("Finished processing %v", workItem)
+			timeTaken = time.Since(workItem.enqueueTime).Seconds()
+			recordWorkItemLatencyMetrics(workItemProcessingTimeWorkItemMetric, timeTaken)
 		}
 
 		tc.workqueue.Forget(obj)
-		klog.Infof("Finished processing %v", workItem)
 		return nil
 	}(obj)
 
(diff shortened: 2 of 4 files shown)


pkg/controllers/tagging/tagging_controller.go
```diff
@@ -31,19 +31,25 @@ import (
...
 type workItem struct {
-	node   *v1.Node
-	action func(node *v1.Node) error
+	node           *v1.Node
+	action         func(node *v1.Node) error
+	requeuingCount int
 }
 
+const (
+	maxRequeuingCount = 10
```
saurav-agarwalla, line 40: Do we know what this translates into in terms of retry duration?

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -141,9 +147,15 @@ func (tc *Controller) process() bool {
 
 		err := workItem.action(workItem.node)
 		if err != nil {
-			// Put the item back on the workqueue to handle any transient errors.
-			tc.workqueue.AddRateLimited(workItem)
-			return fmt.Errorf("error finishing work item '%v': %s, requeuing", workItem, err.Error())
+			if workItem.requeuingCount < maxRequeuingCount {
```
saurav-agarwalla, line 150: Can we add unit tests which verify this behavior?

pkg/controllers/tagging/tagging_controller.go
nguyenkndinh: We are using the [DefaultControllerRateLimiter](https://github.com/kubernetes/client-go/blob/4fc3881b7ffe4a5c1c7c3cffc7dd9d9667989730/util/workqueue/default_rate_limiters.go#L39), which use the worst case of the 2 limiters:
- [ItemExponentialFailureRateLimiter](https://github.com/kubernetes/client-go/blob/4fc3881b7ffe4a5c1c7c3cffc7dd9d9667989730/util/workqueue/default_rate_limiters.go#L67), with a base delay of 5 milliseconds and max delay of 1000 seconds
- [BucketRateLimiter](https://github.com/kubernetes/client-go/blob/4fc3881b7ffe4a5c1c7c3cffc7dd9d9667989730/util/workqueue/default_rate_limiters.go#L48), with a limit of 10, and 100 burst.
If not counting BucketRateLimiter cause that's a different beast, the work items would have to wait anywhere from 50 milliseconds to 10000 seconds in retry time (excluding work time and queue wait time as they can differ). However, the real avg number is a bit more complicated depending on the number of items in the queue, if the item has been enqueue many times, which rate limiter was used. So, `When` to add an item back in the queue depends on either [BucketRateLimiter](https://github.com/kubernetes/client-go/blob/4fc3881b7ffe4a5c1c7c3cffc7dd9d9667989730/util/workqueue/default_rate_limiters.go#L54) or [ItemExponentialFailureRateLimiter](https://github.com/kubernetes/client-go/blob/4fc3881b7ffe4a5c1c7c3cffc7dd9d9667989730/util/workqueue/default_rate_limiters.go#L89) is selected as worst case. However, I wouldn't expect the worst case of 10000 to ever happens, though we need metrics data to fine tune this as it goes I assume.

pkg/controllers/tagging/tagging_controller.go
nguyenkndinh: done

nckturner:
Can you squash your commits

nckturner:
/hold
for squash

pkg/controllers/tagging/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+	})
+}
+
+func recordWorkitemMetrics(actionName string, timeTaken float64, err error) {
+	if err != nil {
+		workItemError.With(metrics.Labels{"workqueue": actionName}).Inc()
+	}
+
+	workItemLatency.With(metrics.Labels{"workqueue": actionName}).Observe(timeTaken)
+}
```
saurav-agarwalla, line 56: Should we break this down into separate functions for count and latency metrics? Seems like at least for error counts, we're going to have zero latency all the time which doesn't seem desirable?

pkg/controllers/tagging/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+func registerMetrics() {
+	register.Do(func() {
+		legacyregistry.MustRegister(workItemLatency)
+		legacyregistry.MustRegister(workItemError)
+	})
+}
+
+func recordWorkitemMetrics(actionName string, timeTaken float64, err error) {
+	if err != nil {
+		workItemError.With(metrics.Labels{"workqueue": actionName}).Inc()
```
saurav-agarwalla, line 52: Have a question around what the dimensions for this metric is going to look like - are we able to query this metric using actionName as well? Because we seem to be passing strings with node names wherever we're calling this function so need to understand if that makes sense or if we need actionName to be somewhat static.

pkg/controllers/tagging/tagging_controller_test.go
```diff
@@ -135,15 +135,26 @@ func Test_NodesJoiningAndLeaving(t *testing.T) {
...
 			}
-			tc.process()
+
+			for tc.workqueue.Len() > 0 {
+				tc.process()
+
+				// sleep briefly because of exponential backoff when requeueing failed workitem
+				// resulting in workqueue to be empty if checked immediately
+				time.Sleep(1500 * time.Millisecond)
+			}
```
saurav-agarwalla, line 145: Can you please explain why this is needed?

nguyenkndinh: It is because of [this](https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/controllers/tagging/tagging_controller.go#L145). [AddRateLimited](https://github.com/kubernetes/kubernetes/blob/a88462e901710ab5df1fa6e99a046aa921dcdd55/staging/src/k8s.io/client-go/util/workqueue/rate_limiting_queue.go#L24) only adds back the item to the queue after some period of wait time, and it gets longer each try. As such, I put this sleep function in here so that we are sure to capture all retries for testing. I admit this depends on the implementation of the workqueue, which is not ideal because the test could fail not for its own reason.

pkg/controllers/tagging/metrics.go
nguyenkndinh: That's a great call out, as I'm based this change off of [aws_metrics.go](https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws_metrics.go), I assume the dimensions for this are similar to [this](https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws_metrics.go#L54). I'll check in CW and see how they are being produced. However, I think you're right as the usage in [aws.go](https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws.go#L1030) is also somewhat static.

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -135,19 +143,34 @@ func (tc *Controller) process() bool {
...
+				// Put the item back on the workqueue to handle any transient errors.
+				workItem.requeuingCount++
+				tc.workqueue.AddRateLimited(workItem)
+
+				return fmt.Errorf("error processing work item '%v': %s, requeuing count %d", workItem, err.Error(), workItem.requeuingCount)
+			}
+
+			klog.Errorf("error processing work item '%v': %s, requeuing count exceeded", workItem, err.Error())
+			recordWorkItemErrorMetrics("process_work_item")
```
saurav-agarwalla, line 166: I think we may want to track both total errors and errors after all retries are exhausted. Can you please emit two labels for this: `total_errors` and `errors_after_retries_exhausted`? You can emit `total_errors` inside the `if` above and `errors_after_retries_exhausted` outside it.

pkg/controllers/tagging/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+		legacyregistry.MustRegister(workItemError)
+	})
+}
+
+func recordWorkItemLatencyMetrics(actionName string, timeTaken float64) {
+	workItemLatency.With(metrics.Labels{"workqueue": actionName}).Observe(timeTaken)
+}
+
+func recordWorkItemErrorMetrics(actionName string) {
+	workItemError.With(metrics.Labels{"workqueue": actionName}).Inc()
```
saurav-agarwalla, line 55: Can we change the label name to `error_type` instead and change the param name accordingly?

pkg/controllers/tagging/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+// registerMetrics registers tagging-controller metrics.
+func registerMetrics() {
+	register.Do(func() {
+		legacyregistry.MustRegister(workItemLatency)
+		legacyregistry.MustRegister(workItemError)
+	})
+}
+
+func recordWorkItemLatencyMetrics(actionName string, timeTaken float64) {
+	workItemLatency.With(metrics.Labels{"workqueue": actionName}).Observe(timeTaken)
```
saurav-agarwalla, line 51: Same for this. Can we change this to `latency_type` instead and change the param name accordingly?

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -135,19 +143,34 @@ func (tc *Controller) process() bool {
...
+		} else {
+			klog.Infof("Finished processing %v", workItem)
+			timeTaken = time.Since(workItem.enqueueTime).Seconds()
+			recordWorkItemLatencyMetrics("process_work_item", timeTaken)
```
saurav-agarwalla, line 170: Can we change this to `work_item_processing_time`?

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -135,19 +143,34 @@ func (tc *Controller) process() bool {
...
 			tc.workqueue.Forget(obj)
-			utilruntime.HandleError(fmt.Errorf("expected workItem in workqueue but got %#v", obj))
+			err := fmt.Errorf("expected workItem in workqueue but got %#v", obj)
+			utilruntime.HandleError(err)
+			recordWorkItemErrorMetrics("dequeue_work_item")
 			return nil
 		}
 
+		timeTaken := time.Since(workItem.enqueueTime).Seconds()
+		recordWorkItemLatencyMetrics("dequeue_work_item", timeTaken)
```
saurav-agarwalla, line 153: Can we change this to `work_item_dequeuing_time`?

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -135,19 +143,34 @@ func (tc *Controller) process() bool {
 		workItem, ok := obj.(*workItem)
 		if !ok {
 			tc.workqueue.Forget(obj)
-			utilruntime.HandleError(fmt.Errorf("expected workItem in workqueue but got %#v", obj))
+			err := fmt.Errorf("expected workItem in workqueue but got %#v", obj)
+			utilruntime.HandleError(err)
+			recordWorkItemErrorMetrics("dequeue_work_item")
```
saurav-agarwalla, line 148: I don't think we need to emit a latency metric here since this means that the workitem is invalid, right?

pkg/controllers/tagging/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func recordWorkItemErrorMetrics(actionName string) {
```
saurav-agarwalla, line 54: Can we add a label for `instance_id` in this function and take in the appropriate params? Might help to see if there are certain instances running into errors.

pkg/controllers/tagging/tagging_controller.go
nguyenkndinh: do you mean to count the number of total errors inside the if with a counter and the number of errors outside the if?

nguyenkndinh: I think I understand it now but pls correct me if I was wrong, it is simply to add `recordWorkItemErrorMetrics(TotalErrors)` in the if and `recordWorkItemErrorMetrics(ErrorsAfterRetriesExhausted)` outside the if so that we can aggregate it later.

saurav-agarwalla (approved):
Looks good to me.

pkg/controllers/tagging/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+	"sync"
+)
+
+var register sync.Once
+
+var (
+	workItemLatency = metrics.NewHistogramVec(
+		&metrics.HistogramOpts{
+			Name:           "cloudprovider_tagging_controller_work_item_latency",
+			Help:           "workitem latency of workitem being in the queue and time it takes to process",
```
nckturner, line 28: Based on the [best practices](https://prometheus.io/docs/practices/naming/), the label name should include the unit:
`cloudprovider_tagging_controller_work_item_latency_seconds` or `cloudprovider_tagging_controller_work_item_duration_seconds` might be better since its the total time of the process?

nguyenkndinh: will update to `cloudprovider_tagging_controller_work_item_duration_seconds`

saurav-agarwalla (approved):

pkg/controllers/tagging/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+		&metrics.HistogramOpts{
+			Name:           "cloudprovider_tagging_controller_work_item_duration_seconds",
+			Help:           "workitem latency of workitem being in the queue and time it takes to process",
+			StabilityLevel: metrics.ALPHA,
+		},
+		[]string{"latency_type"})
+
+	workItemError = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Name:           "cloudprovider_tagging_controller_work_item_error",
```
nckturner, line 35: should be at least `errors`?  or `errors_total`?  [Same reference](https://prometheus.io/docs/practices/naming/). 

> Note that an accumulating count has total as a suffix, in addition to the unit if applicable.

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -178,14 +213,9 @@ func (tc *Controller) tagNodesResources(node *v1.Node) error {
 // tagEc2Instances applies the provided tags to each EC2 instance in
 // the cluster.
 func (tc *Controller) tagEc2Instance(node *v1.Node) error {
-	instanceID, err := awsv1.KubernetesInstanceID(node.Spec.ProviderID).MapToAWSInstanceID()
+	instanceID, _ := awsv1.KubernetesInstanceID(node.Spec.ProviderID).MapToAWSInstanceID()
```
nckturner, line 216: Error isn't likely/possible?

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -31,19 +31,30 @@ import (
...
+	requeuingCount int
+	enqueueTime    time.Time
 }
 
+const (
+	maxRequeuingCount           = 9
+	totalErrors                 = "total_errors"
+	workItemProcessingTime      = "work_item_processing_time"
+	workItemDequeuingTime       = "work_item_dequeuing_time"
+	errorsAfterRetriesExhausted = "errors_after_retries_exhausted"
```
nckturner, line 45: Why does this need to be a separate metric from `total_errors`?

pkg/controllers/tagging/metrics.go
nguyenkndinh: done

pkg/controllers/tagging/tagging_controller.go
nguyenkndinh: It's because we want to know the number of error a particular item encountered before failing out in this case, while the `totalErrors` is for the number of errors before succeeding. I think it's gonna be easier to query the metrics later on if we separate them like so. Pls let me know your recommendations.

pkg/controllers/tagging/tagging_controller.go
nguyenkndinh: error is not possible as we've already checked and handled it earlier in the process: https://github.com/kubernetes/cloud-provider-aws/pull/334/files#diff-ac633e66d7cfdfa6934f05558a33311d493f704ab69e5443c4ad046a47ae1266R158

pkg/controllers/tagging/tagging_controller.go
nckturner: I think what you are saying makes sense--you wouldn't be able to tell the how many errors were seen before a success otherwise (I don't think?)

nckturner (approved):

nguyenkndinh:
squash and /unhold

PR #386, opened 2022-05-24
[FIXES] [Tagging Controller] Bug fixes for tagging controller

saurav-agarwalla:
* Don't tag Fargate nodes
* Label nodes to prevent retagging them again
* Increase bucket size for latency metrics

**What type of PR is this?**
/kind bug

Fixes issues with Tagging Controller

```release-note
Update Tagging Controller to label nodes to prevent re-tagging of the same node. The tagging controller now restricts the tag keys to 63 characters or less since Kubernetes label keys can't be more than 63 characters long.
```

### DIFF
--- a/pkg/controllers/options/tagging_controller.go
+++ b/pkg/controllers/options/tagging_controller.go
@@ -37,6 +37,13 @@ func (o *TaggingControllerOptions) Validate() error {
 		return fmt.Errorf("--tags must not be empty and must be a form of key=value")
 	}
 
+	for key := range o.Tags {
+		// We label the nodes with the tag keys for a cache-hit check so restrict tag keys to 63 characters.
+		if len(key) > 63 {
+			return fmt.Errorf("Tag keys must not be more than 63 characters long")
+		}
+	}
+
 	if len(o.Resources) == 0 {
 		return fmt.Errorf("--resources must not be empty")
 	}
--- a/pkg/controllers/tagging/tagging_controller.go
+++ b/pkg/controllers/tagging/tagging_controller.go
@@ -25,7 +25,9 @@ import (
 	cloudprovider "k8s.io/cloud-provider"
 	opt "k8s.io/cloud-provider-aws/pkg/controllers/options"
 	awsv1 "k8s.io/cloud-provider-aws/pkg/providers/v1"
+	nodehelpers "k8s.io/cloud-provider/node/helpers"
 	"k8s.io/klog/v2"
+	"strings"
 	"time"
 )
 
@@ -35,6 +37,10 @@ type workItem struct {
 	action         func(node *v1.Node) error
 	requeuingCount int
 	enqueueTime    time.Time
+}
+
+func (w workItem) String() string {
+	return fmt.Sprintf("[Node: %s, RequeuingCount: %d, EnqueueTime: %s]", w.node.GetName(), w.requeuingCount, w.enqueueTime)
 }
 
 const (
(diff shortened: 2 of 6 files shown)


pkg/controllers/tagging/tagging_controller.go
```diff
@@ -281,3 +327,31 @@ func (tc *Controller) enqueueNode(obj interface{}, action func(node *v1.Node) er
...
+	return false
+}
+
+func convertTagsToLabels(tags map[string]string) map[string]string {
+	// Kubernetes label names can't contain ':' whereas it is very common in tags so convert that to '.' instead.
+	// Also, Kubernetes label keys and values shouldn't be more than 63 characters long so we ignore the values
+	// while labeling so that we can allow tag values to be longer than that.
+	labels := map[string]string{}
+	for key := range tags {
+		labels[strings.ReplaceAll(key, ":", ".")] = ""
```
nckturner, line 354: So this will mean there is a possible collision where `tagkey:xyz` and `tagkey.xyz` will use the same label, so we should probably at least print a warning in the logs.  Is that the only difference in allowed characters?

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -105,9 +111,29 @@ func NewTaggingController(
...
-		AddFunc:    func(obj interface{}) { tc.enqueueNode(obj, tc.tagNodesResources) },
-		UpdateFunc: func(oldObj, newObj interface{}) { tc.enqueueNode(newObj, tc.tagNodesResources) },
-		DeleteFunc: func(obj interface{}) { tc.enqueueNode(obj, tc.untagNodeResources) },
+		AddFunc: func(obj interface{}) {
+			node := obj.(*v1.Node)
+			tc.enqueueNode(node, tc.tagNodesResources)
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			node := newObj.(*v1.Node)
+			// We have a check for this when tagging instances too. The reason this check exists here is to make sure that
```
nckturner, line 120: > We have a check for this

Use more descriptive wording than `this`.  It seems like its explained partially in the next sentence so maybe just reword so the explanation doesn't come after the first sentence.  E.g.:

> Check if tagging is required by converting desired tags to labels and checking if node has already been labelled.  The reason this check exists...

pkg/controllers/options/tagging_controller.go
```diff
@@ -37,6 +37,13 @@ func (o *TaggingControllerOptions) Validate() error {
 		return fmt.Errorf("--tags must not be empty and must be a form of key=value")
 	}
 
+	for key := range o.Tags {
+		// We label the nodes with the tag keys for a cache-hit check so restrict tag keys to 63 characters.
+		if len(key) > 63 {
+			return fmt.Errorf("Tag keys must not be more than 63 characters long")
```
nckturner, line 43: Remind me -- does this cause the controller to exit?  And does it cause the entire CCM to crash if one controller exits?

nckturner:
Also open this PR against master, and cherry pick back to 1.22.  

/hold

pkg/controllers/tagging/tagging_controller.go
saurav-agarwalla: It is hard to find out exactly what the difference is because not all AWS services have the same tag restrictions and this can change. I am now wondering if instead of us converting the tags to labels and using that for check, maybe we should have a tagging controller specific label which we can use to identify if a node was already processed?

pkg/controllers/options/tagging_controller.go
saurav-agarwalla: Yes, it would and on thinking again, it is probably not desirable since some customers might have a dynamic tag which gets set during runtime and we don't want it to cause an unrecoverable failure. Like I mentioned earlier, changing to a controller-specific label should fix this problem.

saurav-agarwalla:
> 

Addressed comments and opened a PR against master - https://github.com/kubernetes/cloud-provider-aws/pull/387.

nckturner:
Going to close this so that you can use the cherry pick script to go from master to release branches.
/close

PR #387, opened 2022-05-24
[FIXES] [Tagging Controller] Fix issues in tagging controller

saurav-agarwalla:
* Don't tag Fargate nodes
* Label nodes to prevent retagging them again
* Increase bucket size for latency metrics

**What type of PR is this?**
/kind bug

Fixes issues with Tagging Controller

```release-note
Update Tagging Controller to label nodes to prevent re-tagging of the same node. This adds a new label to add nodes processed by the tagging controller.
```

### DIFF
--- a/pkg/controllers/tagging/tagging_controller.go
+++ b/pkg/controllers/tagging/tagging_controller.go
@@ -14,6 +14,7 @@ limitations under the License.
 package tagging
 
 import (
+	"crypto/md5"
 	"fmt"
 	v1 "k8s.io/api/core/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -25,7 +26,10 @@ import (
 	cloudprovider "k8s.io/cloud-provider"
 	opt "k8s.io/cloud-provider-aws/pkg/controllers/options"
 	awsv1 "k8s.io/cloud-provider-aws/pkg/providers/v1"
+	nodehelpers "k8s.io/cloud-provider/node/helpers"
 	"k8s.io/klog/v2"
+	"sort"
+	"strings"
 	"time"
 )
 
@@ -37,7 +41,13 @@ type workItem struct {
 	enqueueTime    time.Time
 }
 
+func (w workItem) String() string {
+	return fmt.Sprintf("[Node: %s, RequeuingCount: %d, EnqueueTime: %s]", w.node.GetName(), w.requeuingCount, w.enqueueTime)
+}
+
 const (
+	taggingControllerLabelKey = "k8s.io/cloud-provider-aws"
+
 	maxRequeuingCount = 9
 
 	// The label for depicting total number of errors a work item encounter and succeed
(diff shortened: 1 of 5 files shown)


pkg/controllers/tagging/tagging_controller.go
```diff
@@ -281,3 +327,15 @@ func (tc *Controller) enqueueNode(obj interface{}, action func(node *v1.Node) er
 	tc.workqueue.Add(item)
 	klog.Infof("Added %s to the workqueue", item)
 }
+
+func (tc *Controller) isTaggingRequired(node *v1.Node) bool {
+	if node.Labels != nil {
+		if labelValue, ok := node.Labels[taggingControllerLabelKey]; !ok || labelValue != taggingControllerLabelValue {
```
rifelpet, line 333: This would mean if the `--tags` flag value ever changed, CCM wouldn't update the tags on existing nodes to match, correct? I'm wondering if the label value should be a hash of the tags that it ensured exists on the instance

saurav-agarwalla: That's correct, yes. And thanks for the call out. I was hoping that these tags would stay the same and we would possibly add logic in the future to take these from a Config Map instead of flags to make them dynamic although I do agree that some customers might want to update whatever tags they're adding. Let me see if I can support that.

nckturner: @rifelpet that's a great idea :100:

saurav-agarwalla: Updated the PR with this check. 😄

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -281,3 +329,24 @@ func (tc *Controller) enqueueNode(obj interface{}, action func(node *v1.Node) er
...
 }
+
+func (tc *Controller) isTaggingRequired(node *v1.Node) bool {
+	if node.Labels != nil {
+		if labelValue, ok := node.Labels[taggingControllerLabelKey]; !ok || labelValue != tc.getChecksumOfTags() {
+			return true
+		}
+	} else {
+		return true
+	}
```
nckturner, line 340: minor de-nest nit: 

```suggestion
	if node.Labels == nil {
		return true
	}

	if labelValue, ok := node.Labels[taggingControllerLabelKey]; !ok || labelValue != tc.getChecksumOfTags() {
		return true
	}

```

pkg/controllers/tagging/tagging_controller_test.go
```diff
@@ -70,6 +70,54 @@ func Test_NodesJoiningAndLeaving(t *testing.T) {
 			toBeTagged:       true,
 			expectedMessages: []string{"Successfully tagged i-0001"},
 		},
+		{
+			name: "node0 joins the cluster and was tagged earlier with different tags.",
+			currNode: &v1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:              "node0",
+					CreationTimestamp: metav1.Date(2012, 1, 1, 0, 0, 0, 0, time.UTC),
+					Labels: map[string]string{
```
nckturner, line 79: Can you have a case with >1 labels as well?

saurav-agarwalla: Did you mean where multiple labels exist already on the node? Because I already modified the tests to have two labels and added a test where one of the labels was already present on the node.

nckturner: Yeah that covers it, thanks. :+1:

hakman:
@saurav-agarwalla I only see this cherry-plicked to 1.22, 1.21 and 1.20.
Probably it should be in 1.23 too, right?

saurav-agarwalla:
@hakman yes, my bad. Forgot that there's a release branch for 1.23 as well now. I'll raise a PR for that.

PR #408, opened 2022-06-13
Fix the kops-example script

olemarkus:
**What type of PR is this?**
/kind cleanup

```release-note
Update the `kops-test.sh` script to work with newer kOps versions.
```

### DIFF
--- a/examples/kops-new-cluster/base/cluster.yaml
+++ b/examples/kops-new-cluster/base/cluster.yaml
@@ -10,7 +10,7 @@ spec:
     rbac: {}
   channel: stable
   cloudProvider: aws
-  containerRuntime: docker
+  containerRuntime: containerd
   etcdClusters:
   - cpuRequest: 200m
     etcdMembers:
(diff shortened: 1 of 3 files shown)


examples/kops-new-cluster/base/cluster.yaml
```diff
@@ -47,16 +47,14 @@ spec:
     anonymousAuth: false
   kubernetesApiAccess:
   - 0.0.0.0/0
-  kubernetesVersion: 1.21.0
   masterInternalName: api.internal.aws-external-cloud-provider-example.k8s.local
   masterPublicName: api.aws-external-cloud-provider-example.k8s.local
   networkCIDR: 172.20.0.0/16
   networking:
-    amazonvpc: {}
+    cilium: {}
```
nckturner, line 54: Reasoning?

olemarkus: The configuration does not work with amazonvpc because the cluster was expecting an overlay network. This could also have been fixed, but would require more changes.
I just picked cilium (essentially over calico) by random.

examples/kops-new-cluster/base/cluster.yaml
```diff
@@ -10,7 +10,7 @@ spec:
     rbac: {}
```
nckturner, line 10: Per your comment on the other PR, we should drop these cluster specs?

olemarkus: When 1.24 is out, I can restructure this bit.

PR #409, opened 2022-06-13
Add note about node naming conventions

olemarkus:
/kind documentation

A number of people try to use CCM with naming conventions that does not work (e.g `ip-xxx-xxx-xxx-xxx.my.domain`). This should make the requirements more clear.

```release-note
Document the supported node naming conventions.
```

### DIFF
--- a/docs/prerequisites.md
+++ b/docs/prerequisites.md
@@ -1,5 +1,13 @@
 # Prerequisites
 
+## Node name conventions
+
+AWS supports two naming conventions: [IP-based or resource-based naming](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-naming.html).
+
+When _IP-based naming_ is used, the nodes must be named after the instance followed by the regional domain name (`ip-xxx-xxx-xxx-xxx.ec2.<region>.internal`). If you have custom domain name set in the DHCP options, you must set `--hostname-override` on kube-proxy and kubelet to match the above-mentioned naming convention.
+
+When _resource based naming_ is used, the node must be named after the instance without any domain name (`i-1234567890abcdefg`). Custom domain name may be used as long as the output of `hostname` does not include the domain name. `--hostname-override` should not be set on any components when using resource-based naming.
+
 ## IAM Policies
 For the `aws-cloud-controller-manager` to be able to communicate to AWS APIs, you will need to create a few IAM policies for your EC2 instances. The control plane (formerly master) policy is a bit open and can be scaled back depending on the use case. Adjust these based on your needs.
 


docs/prerequisites.md
```diff
@@ -1,5 +1,13 @@
 # Prerequisites
 
+## Node name conventions
+
+AWS supports two naming conventions: [IP-based or resource-based naming](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-naming.html).
+
+When _IP-based naming_ is used, the nodes must be named after the instance followed by the regional domain name (`ip-xxx-xxx-xxx-xxx.ec2.<region>.internal`). Custom domain names must not be used. If you have custom domain name set in the DHCP options, you must set `--hostname-override` on kube-proxy and kubelet to match the above-mentioned naming convention.
+
+When _resource based naming_ is used, the node must be named after the instance without any domain name (`i-1234567890abcdefg`). Custom domain name may be used as long as the output of `hostname` does not include the domain name. `--hostname-override` should not be set on any components when using resource-based naming.
```
nckturner, line 9: > as long as the output of `hostname` does not include the domain name.

What breaks here?

docs/prerequisites.md
```diff
@@ -1,5 +1,13 @@
 # Prerequisites
 
+## Node name conventions
+
+AWS supports two naming conventions: [IP-based or resource-based naming](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-naming.html).
+
+When _IP-based naming_ is used, the nodes must be named after the instance followed by the regional domain name (`ip-xxx-xxx-xxx-xxx.ec2.<region>.internal`). Custom domain names must not be used. If you have custom domain name set in the DHCP options, you must set `--hostname-override` on kube-proxy and kubelet to match the above-mentioned naming convention.
```
nckturner, line 7: So should it be `Custom domain names must not be used, unless...`?

olemarkus: I just dropped that sentence.

docs/prerequisites.md
```diff
@@ -1,5 +1,13 @@
 # Prerequisites
 
+## Node name conventions
+
+AWS supports two naming conventions: [IP-based or resource-based naming](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-naming.html).
```
nckturner, line 5: Maybe an additional sentence for further clarity, something like: `In order to use either of the naming conventions, simply name the nodes in your cluster following either of the two naming conventions.  Both naming conventions can be used simultaneously in the same cluster.` (they can right?)

docs/prerequisites.md
```diff
@@ -1,5 +1,13 @@
...
+
+When _IP-based naming_ is used, the nodes must be named after the instance followed by the regional domain name (`ip-xxx-xxx-xxx-xxx.ec2.<region>.internal`). If you have custom domain name set in the DHCP options, you must set `--hostname-override` on kube-proxy and kubelet to match the above-mentioned naming convention.
+
+When _resource based naming_ is used, the node must be named after the instance without any domain name (`i-1234567890abcdefg`). Custom domain name may be used as long as the output of `hostname` does not include the domain name. `--hostname-override` should not be set on any components when using resource-based naming.
```
rohitagarwal003, line 9: > Custom domain name may be used as long as the output of `hostname` does not include the domain name.

@olemarkus Can you clarify what this line means? Custom names can be used for what if they can't be used for node names?

docs/prerequisites.md
```diff
@@ -1,5 +1,13 @@
 # Prerequisites
 
+## Node name conventions
+
+AWS supports two naming conventions: [IP-based or resource-based naming](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-naming.html).
+
+When _IP-based naming_ is used, the nodes must be named after the instance followed by the regional domain name (`ip-xxx-xxx-xxx-xxx.ec2.<region>.internal`). If you have custom domain name set in the DHCP options, you must set `--hostname-override` on kube-proxy and kubelet to match the above-mentioned naming convention.
```
rohitagarwal003, line 7: > `ip-xxx-xxx-xxx-xxx.ec2.<region>.internal`

Is this correct? Should this be `ip-xxx-xxx-xxx-xxx.<region>.compute.internal` (for non us-east-1 and `ip-xxx-xxx-xxx-xxxx.ec2.internal` for us-east-1?

docs/prerequisites.md
olemarkus: instance-level domain name. Typically DHCP option domain name for the VPC.

docs/prerequisites.md
olemarkus: Yeah you are correct. This one mixed up the us-east-1 convention with the other region.

Issue #430, opened 2022-07-12
Helm chart does not select control-plane nodes on 1.24


kwohlfahrt:
**What happened**:

When deploying the cloud provider from the Helm Chart using the configuration below, the daemonset selects no nodes to deploy on.

```
USER-SUPPLIED VALUES:
args:
- --cloud-provider=aws
- --cluster-cidr=10.0.0.0/12
- --cluster-name=kai-test-cluster
- --configure-cloud-routes=false
image:
  tag: v1.24.1
```

This additional configuration is required (note that `node-role.kubernetes.io/master: null` is required to _remove_ this selector):

```
nodeSelector:
  node-role.kubernetes.io/control-plane: ""
  node-role.kubernetes.io/master: null
tolerations:
- effect: NoSchedule
  key: node-role.kubernetes.io/control-plane
- effect: NoSchedule
  key: node-role.kubernetes.io/master
```

**What you expected to happen**:

I expect the helm chart to successfully deploy without having to specify custom labels and taints.

**How to reproduce it (as minimally and precisely as possible)**:

Provision a kubernetes version 1.24 cluster, for example with `kubeadm`. Add the above configuration snippets to `values.yaml`, and install the provider with `helm install --repo 'https://kubernetes.github.io/cloud-provider-aws' --values aws-values.yaml --version 0.0.6 aws-cloud-controller-manager aws-cloud-controller-manager`.

**Anything else we need to know?**:

The `control-plane` node labels and taints have been present [since v1.20](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#urgent-upgrade-notes), and the `master` label and taint will be removed in v1.25 (for reference, see the [v1.24 release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#urgent-upgrade-notes)), the next upstream release. It seems like now is a good time to make this transition.

I'm happy to open a PR for this.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.24.2
- Cloud provider or hardware configuration: AWS EC2 (not EKS)
- OS (e.g. from /etc/os-release): Ubuntu 20.04.4
- Kernel (e.g. `uname -a`): `5.13.0-1031-aws`
- Install tools: `kubeadm`
- Others:

/kind bug

olemarkus:
[...]

PR #431, opened 2022-07-18
chore: add k8s 1.24+ taints

frezbo:
Add new `node-role.kubernetes.io/control-plane` taints so that the
cloud controller tolerates them.

Signed-off-by: Noel Georgi <git@frezbo.dev>

Fixes: #430

### DIFF
--- a/charts/aws-cloud-controller-manager/Chart.yaml
+++ b/charts/aws-cloud-controller-manager/Chart.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 name: aws-cloud-controller-manager
 description: Installs Cloud Controller Manager for AWS Cloud Provider
-version: 0.0.6
+version: 0.0.7
 appVersion: v1.23.0-alpha.0
 maintainers:
 - name: Nick Turner
--- a/charts/aws-cloud-controller-manager/templates/daemonset.yaml
+++ b/charts/aws-cloud-controller-manager/templates/daemonset.yaml
@@ -20,7 +20,16 @@ spec:
         k8s-app:  {{ template "cloud-controller-manager.name" . }}
     spec:
       tolerations: {{ toYaml ( .Values.tolerations | default .Values.tolerations ) | nindent 8 }}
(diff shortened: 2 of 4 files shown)


kwohlfahrt:
It would only affect verions <= 1.19, which went EOL October 2021, so I would support just moving forward. However on EKS 1.19 is supported until August 2022.

frezbo:
@kwohlfahrt I have updated the PR and ended up using `nodeAffinity` since it's OR'ed. So this should still work for old k8s. I've also fixed up the helm values file, so it installs properly out of the box.

mikescholl-sysdig:
hI @frezbo @nckturner any chance this is getting pushed forward? Currently running into this in a couple clusters where the `master` role no longer exists.

frezbo:
@mikescholl-sysdig I'm also waiting on the maintainers

nckturner:
Sorry for the delay.  

/triage accepted
/ok-to-test

frezbo:
```bash
 /home/prow/go/src/k8s.io/cloud-provider-aws
./hack/e2e/run.sh: line 149: /home/prow/go/src/k8s.io/cloud-provider-aws/_output/test/20220819181918/kops: No such file or directory 
```

seems like an unrelated failure? :thinking:

nckturner:
@frezbo you'll need to rebase, the above failure was fixed in master.

nckturner:
We should bump the chart version as well, shouldn't we?

frezbo:
> We should bump the chart version as well, shouldn't we?

Does 0.0.7 sound fine?

frezbo:
Updated the chart version to `0.0.7` and also fixed the e2e test failure.

frezbo:
@nckturner this should be good to go

frezbo:
closing in favor of https://github.com/kubernetes/cloud-provider-aws/pull/466

olemarkus:
Sorry I missed this one. I was sort of going through my mailbox latest to oldest and #466 appeared first.

mikescholl-sysdig:
@olemarkus FYI #466 is not an ideal solution since you will now break K8S versions older than 1.24 that do not have the control-plane taints. It is also impossible to clear that setting when its set in the values.yaml

PR #437, opened 2022-08-03
Skip health rule creation when it is a subset of the client rule

sjenning:
**What type of PR is this?**
/kind cleanup

Currently, if a Service of type LoadBalancer is created, and `Spec.LoadBalancerSourceRanges` is not specified, the AWS cloud-provider creates one rule for the client with a source of all addresses (0.0.0.0/0).  It also create a rule per-zone for the health check.  In many cases, these health rules are subsets of the client rule with more specific CIDR blocks.

AWS Quota on `Inbound or outbound rules per security group` is defaulted to 60 with a hard max of 200 (security groups per network interface * inbound or outbound rules per security group <= 1000).

Thus, a Service of type LoadBalancer on a cluster with nodes in 3 zones will result in 4 rules being added to the security group.  This means that, for such a cluster with the default quota, a maximum of 15 LBs can be created.

https://docs.aws.amazon.com/eks/latest/userguide/network-load-balancing.html#network-load-balancer

> For each NLB that you create Amazon EKS adds one inbound rule to the node's security group for client traffic and one rule for each load balancer subnet in the VPC for health checks. Deployment of a service of type LoadBalancer can fail if Amazon EKS attempts to create rules that exceed the quota for the maximum number of rules allowed for a security group.

The over quota condition manifests as a `RulesPerSecurityGroupLimitExceeded` error on the CCM or KCM (legacy).

This PR skips the health rules if they are a subset of the client rule.  This quadruples the number of typical Services of type LoadBalancer (i.e. Services without `Spec.LoadBalancerSourceRanges` specified) that will fit in a 3-zone cluster for a given quota. 

Second commit is just gofmt update since the job uses go1.19 now
https://go.dev/doc/go1.19#go-doc

Reduced number of rules in node security groups

```release-note
Skip creation of health check rules when they are a subset of the client rule
```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -173,7 +173,9 @@ const ServiceAnnotationLoadBalancerSSLNegotiationPolicy = "service.beta.kubernet
 // ServiceAnnotationLoadBalancerBEProtocol is the annotation used on the service
 // to specify the protocol spoken by the backend (pod) behind a listener.
 // If `http` (default) or `https`, an HTTPS listener that terminates the
-//  connection and parses headers is created.
+//
+//	connection and parses headers is created.
+//
 // If set to `ssl` or `tcp`, a "raw" SSL listener is used.
 // If set to `http` and `aws-load-balancer-ssl-cert` is not used then
 // a HTTP listener is used.
@@ -236,9 +238,9 @@ const ServiceAnnotationLoadBalancerTargetNodeLabels = "service.beta.kubernetes.i
 // subnetID or subnetName from different AZs
 // By default, the controller will auto-discover the subnets. If there are multiple subnets per AZ, auto-discovery
 // will break the tie in the following order -
-//   1. prefer the subnet with the correct role tag. kubernetes.io/role/elb for public and kubernetes.io/role/internal-elb for private access
-//   2. prefer the subnet with the cluster tag kubernetes.io/cluster/<Cluster Name>
-//   3. prefer the subnet that is first in lexicographic order
+//  1. prefer the subnet with the correct role tag. kubernetes.io/role/elb for public and kubernetes.io/role/internal-elb for private access
+//  2. prefer the subnet with the cluster tag kubernetes.io/cluster/<Cluster Name>
+//  3. prefer the subnet that is first in lexicographic order
 const ServiceAnnotationLoadBalancerSubnets = "service.beta.kubernetes.io/aws-load-balancer-subnets"
 
 // Event key when a volume is stuck on attaching state when being attached to a volume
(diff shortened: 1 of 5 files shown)


M00nF1sh (approved):
the change looks good to me.
/lgtm
/approve

pkg/providers/v1/aws.go
```diff
@@ -173,7 +173,9 @@ const ServiceAnnotationLoadBalancerSSLNegotiationPolicy = "service.beta.kubernet
 // ServiceAnnotationLoadBalancerBEProtocol is the annotation used on the service
 // to specify the protocol spoken by the backend (pod) behind a listener.
 // If `http` (default) or `https`, an HTTPS listener that terminates the
-//  connection and parses headers is created.
+//
+//	connection and parses headers is created.
```
micahhausler, line 177: Is this doc change intentional?

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -834,8 +834,12 @@ func (c *Cloud) updateInstanceSecurityGroupsForNLB(lbName string, instances map[
 		for sgID, sg := range clusterSGs {
 			sgPerms := NewIPPermissionSet(sg.IpPermissions...).Ungroup()
 			if desiredSGIDs.Has(sgID) {
-				if err := c.updateInstanceSecurityGroupForNLBTraffic(sgID, sgPerms, healthRuleAnnotation, "tcp", healthCheckPorts, subnetCIDRs); err != nil {
-					return err
+				// If the client rule is 1) all addresses 2) tcp and 3) has same ports as the healthcheck,
+				// then the health rules are a subset of the client rule and are not needed.
+				if len(clientCIDRs) != 1 || clientCIDRs[0] != "0.0.0.0/0" || clientProtocol != "tcp" || !healthCheckPorts.Equal(clientPorts) {
```
micahhausler, line 839: Can you add a unit test to cover this new logic?

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -1234,7 +1238,8 @@ func (c *Cloud) ensureLoadBalancer(namespacedName types.NamespacedName, loadBala
 
 // syncElbListeners computes a plan to reconcile the desired vs actual state of the listeners on an ELB
 // NOTE: there exists an O(nlgn) implementation for this function. However, as the default limit of
-//       listeners per elb is 100, this implementation is reduced from O(m*n) => O(n).
+//
```
micahhausler, line 1241: Is this doc change intentional?

pkg/providers/v1/aws.go
sjenning: Allows new gofmt to pass

pkg/providers/v1/aws_loadbalancer.go
sjenning: Allows new gofmt to pass

micahhausler: ah we probably need a separate PR to gofmt the whole repo then so individual PRs don't have to worry about this

PR #443, opened 2022-08-19
Handle InvalidInstanceID.NotFound when tagging resources

saurav-agarwalla:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**: We recently found an issue with Node Lifecycle Controller where it leaks node objects in some cases and as a result, this controller runs into errors when trying to tag the instances underlying those node objects. This PR handles those errors gracefully and ignores those resources.

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -173,7 +173,9 @@ const ServiceAnnotationLoadBalancerSSLNegotiationPolicy = "service.beta.kubernet
 // ServiceAnnotationLoadBalancerBEProtocol is the annotation used on the service
 // to specify the protocol spoken by the backend (pod) behind a listener.
 // If `http` (default) or `https`, an HTTPS listener that terminates the
-//  connection and parses headers is created.
+//
+//	connection and parses headers is created.
+//
 // If set to `ssl` or `tcp`, a "raw" SSL listener is used.
 // If set to `http` and `aws-load-balancer-ssl-cert` is not used then
 // a HTTP listener is used.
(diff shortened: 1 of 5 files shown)


nckturner:
Was the leak in the node lifecycle controller reported upstream as well? If so, can you link the issue here?  Also, (concern brought up by @justinsb) do we need to check the other callers of this tagging API?  Are there cases where we lose retries due to ignoring this error, in the case of calling an eventually consistent API?

saurav-agarwalla:
> Was the leak in the node lifecycle controller reported upstream as well? If so, can you link the issue here? Also, (concern brought up by @justinsb) do we need to check the other callers of this tagging API? Are there cases where we lose retries due to ignoring this error, in the case of calling an eventually consistent API?

> Was the leak in the node lifecycle controller reported upstream as well? If so, can you link the issue here?

Checked with the team on this and it looks like we haven't done it yet. I created one for CCM - https://github.com/kubernetes/cloud-provider-aws/issues/444.

> Also, (concern brought up by @justinsb) do we need to check the other callers of this tagging API? Are there cases where we lose retries due to ignoring this error, in the case of calling an eventually consistent API?

Good point. I checked this already. `TagResource` and `UntagResource` functions are only used by the Tagging Controller. Rest of the controllers use [this](https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/tags.go#L198-L244) which we're not changing.

nckturner:
/retest

Can you re-run gofmt, and/or check your go version?  There seems to be a diff in verify-gofmt.sh.

nckturner:
The first run failed on https://github.com/kubernetes/cloud-provider-aws/issues/445, I'm not sure if it is flaky or always fails.

saurav-agarwalla:
> /retest
> 
> Can you re-run gofmt, and/or check your go version? There seems to be a diff in verify-gofmt.sh.

Works fine for me locally and my Golang version is also 1.18. I'll check to see what's up.

nckturner:
@saurav-agarwalla please rebase and update your go version to 1.19

nckturner:
"merged 0 commits" lol wut

PR #446, opened 2022-08-19
Revert "Ensure kops cluster is always torn down after e2e run"

### LINKED ISSUE #445: Kops cleanup gives error
**What happened**:
When running the e2e tests, the trap attempts to run [kops-finish](https://github.com/kubernetes/cloud-provider-aws/blob/master/hack/e2e/run.sh#L130):

```
trap kops-finish EXIT
```

However, the `${test-run}` directory seems to be missing when its run, so perhaps cleanup happens before trap can execute?

```
 Ginkgo ran 1 suite in 2m57.717745585s
Test Suite Passed
/home/prow/go/src/k8s.io/cloud-provider-aws
./hack/e2e/run.sh: line 125: /home/prow/go/src/k8s.io/cloud-provider-aws/_output/test/20220819152419/kops: No such file or directory
make: *** [Makefile:133: test-e2e] Error 127
+ EXIT_VALUE=2
+ set +o xtrace
Cleaning up after docker in docker. 
```

See https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/cloud-provider-aws/443/pull-cloud-provider-aws-e2e/1560647514955190272

Cleanup to happen successfully.

Run the e2e tests in prow by opening a PR.  

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

/kind bug

nckturner:
This reverts commit 0b078ce702846fb1c9357496dfea2f4f46c54188.

**What type of PR is this?**
>

/kind failing-test

Temporarily revert the trap EXIT/kops cleanup function, which is failing in the prow environment.  We do need to figure out our cleanup story but this is blocking PRs.

Fixes #445

```release-note
NONE
```

### DIFF
--- a/hack/e2e/run.sh
+++ b/hack/e2e/run.sh
@@ -121,14 +121,7 @@ aws ecr get-login-password --region "${REGION}" | docker login --username AWS --
 docker tag "${BUILD_IMAGE}" "${IMAGE_NAME}:${IMAGE_TAG}"
 docker push "${IMAGE_NAME}:${IMAGE_TAG}"
 
-function kops-finish {
-    "${test_run}"/kops delete cluster --name "${CLUSTER_NAME}" --yes
-}
-
 if [[ "${UP}" = "yes" ]]; then
-
-    trap kops-finish EXIT
-
     kubetest2 kops \
       -v 2 \
       --up \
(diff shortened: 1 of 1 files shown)


olemarkus:
I think perhaps adding `|| true` to behind the `kops delete` call is a better temporary fix.

nckturner:
I thought about that, but its failing consistently so I want to make sure clusters are cleaned up as frequently as possible.  This seems to fail (almost?) every time, whereas the previous mechanism fails if a new change is pushed on the PR when the existing job is running (perhaps other situations as well).  Seems slightly better?

olemarkus (approved):
Agreed. It' strange that this is failing ...

olemarkus (approved):

nckturner:
> Agreed. It' strange that this is failing ...

Yeah I couldn't figure out why

PR #470, opened 2022-08-25
Fix broken helm chart

camaeel:
**What type of PR is this?**
/kind bug

Current helm chart in the master branch is broken. When applied to the real cluster it throws an error: `error: error validating "STDIN": error validating data: ValidationError(DaemonSet.spec.template.spec.containers[0].env): invalid type for io.k8s.api.core.v1.Container.env: got "map", expected "array"; if you choose to ignore these errors, turn validation off with --validate=false`

None

```release-note
Fixes helm chart default env in values.yaml
```


olemarkus (approved):

puchongcnxx29 (commented):
A

PR #471, opened 2022-08-25
Use ip address to describeENI for fargate nodes

wangamel:
**What type of PR is this?**
/kind bug

The function `InstanceExistsByProviderID` returns true if the instance with the given provider id. It tries to describeENI with either `privateDNSName` or `privateIpAddress` passed in the `providerID`. For fargate use cases, the instanceID got from providerID will be like either ip-name based privateDNSName or just ip address if enableDnsSupport is set to false in a VPC. However, with resource-name based EC2 resources, privateDNSName is a different format, which has instance id instead of ip address to generate privateDNSName. This causes that `InstanceExistsByProviderID` cannot find corresponding instance. The change here is to use privateIdAddress to describeENI since this format remains same for either ip-name based or resource-name based EC2 resources.

Fixes #

```release-note
Fix fargate nodes cannot join the cluster with resource-name enabled for EC2 hostnames
```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -5090,7 +5090,7 @@ func IsFargateNode(nodeName string) bool {
 }
 
 // extract private ip address from node name
-func nodeNameToIpAddress(nodeName string) string {
+func nodeNameToIPAddress(nodeName string) string {
 	nodeName = strings.TrimPrefix(nodeName, privateDNSNamePrefix)
 	nodeName = strings.Split(nodeName, ".")[0]
 	return strings.ReplaceAll(nodeName, "-", ".")
@@ -5200,7 +5200,7 @@ func (c *Cloud) describeNetworkInterfaces(nodeName string) (*ec2.NetworkInterfac
 	// convert node name to ip address because ip-name based and resource-named EC2 resources
 	// may have different privateDNSName formats but same privateIpAddress format
 	if strings.HasPrefix(eniEndpoint, privateDNSNamePrefix) {
-		eniEndpoint = nodeNameToIpAddress(eniEndpoint)
+		eniEndpoint = nodeNameToIPAddress(eniEndpoint)
 	}
 
 	filters = append(filters, newEc2Filter("private-ip-address", eniEndpoint))
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -5089,6 +5089,13 @@ func IsFargateNode(nodeName string) bool {
 	return strings.HasPrefix(nodeName, fargateNodeNamePrefix)
 }
 
+// extract private ip address from node name
+func nodeNameToIpAddress(nodeName string) string {
+	nodeName = strings.TrimPrefix(nodeName, privateDNSNamePrefix)
+	nodeName = strings.Split(nodeName, ".")[0]
+	return strings.ReplaceAll(nodeName, "-", ".")
+}
+
 func (c *Cloud) nodeNameToInstanceID(nodeName types.NodeName) (InstanceID, error) {
 	if strings.HasPrefix(string(nodeName), rbnNamePrefix) {
 		return InstanceID(nodeName), nil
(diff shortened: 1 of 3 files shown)


pkg/providers/v1/aws_test.go
```diff
@@ -3796,6 +3796,15 @@ func TestInstanceExistsByProviderIDForFargate(t *testing.T) {
 	assert.True(t, instanceExist)
 }
 
+func TestInstanceExistsByProviderIDWithNodeNameForFargate(t *testing.T) {
+	awsServices := newMockedFakeAWSServices(TestClusterID)
+	c, _ := newAWSCloud(CloudConfig{}, awsServices)
+
+	instanceExist, err := c.InstanceExistsByProviderID(context.TODO(), "aws:///us-west-2c/1abc-2def/fargate-ip-192-168-164-88.us-west-2.compute.internal")
```
wongma7, line 3803: should there be a test case for `privateDNSName is a different format, which has instance id instead of ip address to generate privateDNSName `?

could you help me fill in this table to make sure my understanding is correct

| | enableDnsSupport true | enableDnsSupport false |
| ------------- | ------------- |------------- |
| RBN true  | ? | ? |
| RBN false | fargate-ip-192-168-164-88.us-west-2.compute.internal | ? |

wongma7: thank you so much I got it now

pkg/providers/v1/aws.go
```diff
@@ -5089,6 +5089,13 @@ func IsFargateNode(nodeName string) bool {
 	return strings.HasPrefix(nodeName, fargateNodeNamePrefix)
 }
 
+// extract private ip address from node name
+func nodeNameToIpAddress(nodeName string) string {
+	nodeName = strings.TrimPrefix(nodeName, privateDNSNamePrefix)
+	nodeName = strings.Split(nodeName, ".")[0]
```
wongma7, line 5095: this breaks for `fargate-192.168.164.88`
see  https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/cloud-provider-aws/471/pull-cloud-provider-aws-test/1562936995968192512 

```
--- FAIL: TestNodeAddressesForFargate (0.00s)
panic: runtime error: index out of range [1] with length 1 [recovered]
	panic: runtime error: index out of range [1] with length 1 
```

pkg/providers/v1/aws.go
```diff
@@ -5089,6 +5089,13 @@ func IsFargateNode(nodeName string) bool {
 	return strings.HasPrefix(nodeName, fargateNodeNamePrefix)
 }
 
+// extract private ip address from node name
+func nodeNameToIpAddress(nodeName string) string {
```
wongma7, line 5093: linter is complaining about the func name:
```
nodeNameToIpAddress should be nodeNameToIPAddress 

 golint -set_exit_status k8s.io/cloud-provider-aws/cmd/aws-cloud-controller-manager k8s.io/cloud-provider-aws/cmd/ecr-credential-provider k8s.io/cloud-provider-aws/pkg/apis/config k8s.io/cloud-provider-aws/pkg/apis/config/v1alpha1 k8s.io/cloud-provider-aws/pkg/controllers/options k8s.io/cloud-provider-aws/pkg/controllers/tagging k8s.io/cloud-provider-aws/pkg/providers/v1 k8s.io/cloud-provider-aws/pkg/providers/v2 k8s.io/cloud-provider-aws/pkg/providers/v2/mocks
/home/prow/go/src/k8s.io/cloud-provider-aws/pkg/providers/v1/aws.go:5093:6: func nodeNameToIpAddress should be nodeNameToIPAddress 
```
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/cloud-provider-aws/471/pull-cloud-provider-aws-check/1562936995934638080

pkg/providers/v1/aws.go
wongma7: or rather fargate-ip-192.168.164.88 ? but how could such a name exist. the test might be wrong, I'm not sure.

wongma7: Maybe with dhcp options sets or something if no domain is provided at all. Just guessing.

wangamel: `nodeAddresses, _ := c.NodeAddressesByProviderID(context.TODO(), "aws:///us-west-2c/1abc-2def/fargate-ip-192.168.164.88")` the original test is wrong, it should `fargate-192.168.164.88`

PR #472, opened 2022-08-26
Release version v1.25.0

olemarkus:
```release-note
NONE
```


olemarkus:
/cc @hakman @nckturner 

We'd like a quick alpha image with the new k8s.io deps updated and see how it fares in the kops e2e grid.

Issue #444, opened 2022-08-19
Node Lifecycle Controller leaks node objects


saurav-agarwalla:
**What happened**: The node lifecycle controller can leak node objects in cases where the controller is down for some time. This is because there are two scenarios that can happen when an underlying instance is terminated - for some time, EC2 returns the status of that instance as terminated in the `DescribeInstances` API call which is correctly handled [here](https://github.com/kubernetes/cloud-provider-aws/blob/1c6194aad0122ab44504de64187e3d1a7415b198/pkg/providers/v1/aws.go#L1708-L1712). But after some time has elapsed, the same API call will start returning `InvalidInstanceID.NotFound` which is where the bug surfaces. It seems, we wrap the error from EC2 [here](https://github.com/kubernetes/cloud-provider-aws/blob/1c6194aad0122ab44504de64187e3d1a7415b198/pkg/providers/v1/aws.go#L1007) which causes the check [here](https://github.com/kubernetes/cloud-provider-aws/blob/1c6194aad0122ab44504de64187e3d1a7415b198/pkg/providers/v1/aws.go#L1696-L1698) to fail since [this cast](https://github.com/kubernetes/cloud-provider-aws/blob/1c6194aad0122ab44504de64187e3d1a7415b198/pkg/providers/v1/aws.go#L1940) will never be true. Not wrapping the error should fix this issue.

**What you expected to happen**: I expect this code to handle `InvalidInstanceID.NotFound` gracefully.

**How to reproduce it (as minimally and precisely as possible)**: Create a dummy node object using `kubectl apply` with a non-existent instance ID. CCM logs would show an error instead of properly handling this case and deleting the node object.

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

/kind bug

saurav-agarwalla:
The legacy provider also has the same bug - https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L974.

olemarkus:
Nice find

/triage accepted

saurav-agarwalla:
> Nice find
> 
> /triage accepted

Credit to @jyotimahapatra for finding the bug. 🙌

PR #484, opened 2022-08-31
Handle wrapped error for InvalidInstanceID.NotFound in InstanceExistsByProviderID

saurav-agarwalla:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**: Fixes https://github.com/kubernetes/cloud-provider-aws/issues/444 and hence prevents leak of node objects.

Fixes https://github.com/kubernetes/cloud-provider-aws/issues/444

**Special notes for your reviewer**: Ideally, this fix should be done [here](https://github.com/kubernetes/cloud-provider-aws/blob/1c6194aad0122ab44504de64187e3d1a7415b198/pkg/providers/v1/aws.go#L1007) by not wrapping the error but that function is exported and used in multiple places in this package as well as probably outside the package as well. Changing it there might have unintended consequences.

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -1941,6 +1941,10 @@ func isAWSErrorInstanceNotFound(err error) bool {
 		if awsError.Code() == ec2.UnsuccessfulInstanceCreditSpecificationErrorCodeInvalidInstanceIdNotFound {
 			return true
 		}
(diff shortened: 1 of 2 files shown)


jyotimahapatra:
I think the usage of string comparison is not the best course, but given the usage/impact of DescribeInstances would be far reaching, it will be a tradeoff.

PR #502, opened 2022-09-26
Single goroutine access to the node cache

kon-angelo:
**What type of PR is this?**
/kind bug

This PR restricts the access to the `describeAllInstancesUncached` to a single worker. For AWS accounts with thousand of EC2 instances the call to fetch the instances from the AWS API is costly in resources. On startup of the CCM with `concurrent-service-syncs` > 1 because the cache is empty all the workers will try to hit the AWS API which creates very large spikes in resources used.

With this PR only one worker will access describeAllInstancesUncached while the rest will wait for the mutex. 

Fixes #

```release-note
Fix a bug that leads to multiple workers making concurrent `DescribeInstance` requests to AWS.
```

### DIFF
--- a/pkg/providers/v1/instances.go
+++ b/pkg/providers/v1/instances.go
@@ -149,7 +149,8 @@ type instanceCache struct {
 	snapshot *allInstancesSnapshot
 }
 
-// Gets the full information about these instance from the EC2 API
+// Gets the full information about these instance from the EC2 API. Caller must have acquired c.mutex before
+// calling describeAllInstancesUncached.
 func (c *instanceCache) describeAllInstancesUncached() (*allInstancesSnapshot, error) {
 	now := time.Now()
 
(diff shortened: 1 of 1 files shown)


jaypipes (approved):
Seems like a small and reasonable fix to me, :+1:

pkg/providers/v1/instances.go
```diff
@@ -195,22 +192,14 @@ type cacheCriteria struct {
...
+	if c.snapshot != nil && c.snapshot.MeetsCriteria(criteria) {
 		klog.V(6).Infof("EC2 DescribeInstances - using cached results")
+		return c.snapshot, nil
 	}
 
-	return snapshot, nil
+	return c.describeAllInstancesUncached()
 }
 
 // getSnapshot returns a snapshot if one exists
```
prateekgogia, line 205: Should we remove the `getSnapshot` method since its not being used anywhere else?

dims:
/approve
/lgtm

looks like comments have been addressed and Jay seems happy with it. It looks good to me as well.

PR #522, opened 2022-10-14
Bump K8s libs to 0.25.3

lobziik:
/kind bug
/kind cleanup

Bug described in https://github.com/kubernetes/kubernetes/issues/112793  was introduced in released versions of `k8s.io/cloud-provider`, namely in v0.25.(0, 1, 2). This was fixed in the most recent `k8s.io/cloud-provider` version `v0.25.3`. This PR bumps k8s libs to bring this fix into cloud-provider-aws.

Also, this patch bumps k8s libs version for the e2e test suite.

Fixes #

```release-note
NONE
```


olemarkus (approved):

olemarkus:
Doesn't look like the version bumps pass tests just like that, unfortunately. It errors out in a panic.

lobziik:
@olemarkus I rolled back bump for e2e suite, `k8s.io/kubernetes/test/e2e/framework` introduces dependency on ginkgo v2 which caused this panic as far as i understand.

I suggest to address dependency bump for e2e suite in follow-up PR, this shouldn't affect this test essence i think.

olemarkus (approved):
I thought we already bumped to ginkgo v2. I've seen this error before when one used mixed versions though. We'll have a look in a followup PR.

PR #529, opened 2022-11-09
Enable booting up kubelet with aws provider when it restarts

vpineda1996:
**What type of PR is this?**
/kind feature

Enables the cloud provider to boot up whenever the CPI is disconnected from the region.

```release-note
Optional configuration introduced to cache the self-describe call. It introduces 
new parameter in cloud config to specify the location of the cache.
```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -17,11 +17,14 @@ limitations under the License.
 package aws
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
+	"os"
 	"regexp"
 	"sort"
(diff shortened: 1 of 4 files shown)


olemarkus:
I believe these code paths only run when using the in-tree cloud provider. This repo still contain quite a few of those because it's been directly split from there. I.e this code does not run within the kubelet.

vpineda1996:
Oh I see. Are you saying that the code path that I am modifying does not run when the AWS cloud provider is run as an external cloud provider. Would it make sense to remove this then?

I was baffled when I was testing this because the patch worked when I was using the in-tree cloud provider but the code didn't seem to run when I used the external cloud provider.

olemarkus:
Yeah. Those code paths can be removed.

PR #540, opened 2022-12-03
Release 1.26.0-rc.1

olemarkus:
Release 1.26-rc.1

/release-notes-none


hakman:
@olemarkus Are you sure you don't want a beta for now?

olemarkus:
Uh I just followed the convention of the upstream libs. That's the big change here anyway. We can call it beta if that matters, but I imagine the only change before the GA release would be to bump the k8s deps to GA.

hakman (approved):

PR #565, opened 2023-01-27
Add support for IPv6 Fargate nodes

HusainZafar:
**What type of PR is this?**
/kind feature

This PR adds support for IPv6 Fargate nodes by assigning the right NodeInternalIP based on IP family

NONE

This code doesn't change anything for EC2 worker nodes

No

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -1668,7 +1668,26 @@ func (c *Cloud) NodeAddressesByProviderID(ctx context.Context, providerID string
 		if eni == nil || err != nil {
 			return nil, err
 		}
-		return getNodeAddressesForFargateNode(aws.StringValue(eni.PrivateDnsName), aws.StringValue(eni.PrivateIpAddress)), nil
+
+		var addresses []v1.NodeAddress
+
+		// Assign NodeInternalIP based on IP family
+		for _, family := range c.cfg.Global.NodeIPFamilies {
+			switch family {
+			case "ipv4":
+				nodeAddresses := getNodeAddressesForFargateNode(aws.StringValue(eni.PrivateDnsName), aws.StringValue(eni.PrivateIpAddress))
+				addresses = append(addresses, nodeAddresses...)
+			case "ipv6":
+				if eni.Ipv6Addresses == nil || len(eni.Ipv6Addresses) == 0 {
+					klog.Errorf("no Ipv6Addresses associated with the eni")
+					continue
+				}
+				internalIPv6Address := eni.Ipv6Addresses[0].Ipv6Address
+				nodeAddresses := getNodeAddressesForFargateNode(aws.StringValue(eni.PrivateDnsName), aws.StringValue(internalIPv6Address))
+				addresses = append(addresses, nodeAddresses...)
+			}
+		}
+		return addresses, nil
 	}
 
 	instance, err := describeInstance(c.ec2, instanceID)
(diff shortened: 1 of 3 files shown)


pkg/providers/v1/aws.go
```diff
@@ -1668,7 +1668,16 @@ func (c *Cloud) NodeAddressesByProviderID(ctx context.Context, providerID string
 		if eni == nil || err != nil {
 			return nil, err
 		}
-		return getNodeAddressesForFargateNode(aws.StringValue(eni.PrivateDnsName), aws.StringValue(eni.PrivateIpAddress)), nil
+
+		// Assign NodeInternalIP based on IP family
+		if eni.Ipv6Addresses != nil {
+			internalIPv6Address := eni.Ipv6Addresses[0].Ipv6Address
+			klog.Infof("Processing IPV6 fargate node with instanceID: %s", string(instanceID))
```
olemarkus, line 1675: I don't think we want these on info level. Can imagine this would get quite noisy.

pkg/providers/v1/aws.go
```diff
@@ -1668,7 +1668,16 @@ func (c *Cloud) NodeAddressesByProviderID(ctx context.Context, providerID string
 		if eni == nil || err != nil {
 			return nil, err
 		}
-		return getNodeAddressesForFargateNode(aws.StringValue(eni.PrivateDnsName), aws.StringValue(eni.PrivateIpAddress)), nil
+
+		// Assign NodeInternalIP based on IP family
+		if eni.Ipv6Addresses != nil {
```
olemarkus, line 1673: Does this mean node IPs are either/or? I think this needs to respect `c.cfg.Global.NodeIPFamilies`

olemarkus (changes requested):
Fargate is dualstack isn't it? So I think I'd like to see more cases for that behaving consistently with EC2.

pkg/providers/v1/aws.go
HusainZafar: Ack, removed the added log lines

pkg/providers/v1/aws.go
HusainZafar: Updated the PR to set the IP based on NodeIPFamilies consistent with EC2 nodes and verified that the addresses field with cloud provider is consistent with addresses field without cloud provider enabled

pkg/providers/v1/aws.go
```diff
@@ -1668,7 +1668,22 @@ func (c *Cloud) NodeAddressesByProviderID(ctx context.Context, providerID string
...
+		var addresses []v1.NodeAddress
+
+		// Assign NodeInternalIP based on IP family
+		for _, family := range c.cfg.Global.NodeIPFamilies {
+			switch family {
+			case "ipv4":
+				nodeAddresses := getNodeAddressesForFargateNode(aws.StringValue(eni.PrivateDnsName), aws.StringValue(eni.PrivateIpAddress))
+				addresses = append(addresses, nodeAddresses...)
+			case "ipv6":
+				internalIPv6Address := eni.Ipv6Addresses[0].Ipv6Address
```
nckturner, line 1681: Even if the NodeIPFamily is set to ipv6, I wonder if its possible for the ENI to not have IPv6 addresses due to a misconfiguration or something.  Can we check to make sure the addresses exist and handle a failure case before referencing the first item in the list here?

HusainZafar: Ack, added a check to make sure Ipv6Addresses is not missing or empty

HusainZafar: Valid point for that use case, I have updated to log the error and return the existing addresses

nckturner:
@HusainZafar will you squash you commits please?

PR #581, opened 2023-03-08
Don't update latest tag when uploading artifacts

hakman:
/cc @olemarkus @dims


hakman:
Closing in favor of #582

Issue #596, opened 2023-04-16
ecr-credential-provider fails with Kubelet 1.27


ikolomiyets:
**What happened**:
ecr-credentials-provider deployed to managed k8s cluster fails to produce credentials.
Following errors appeared in the logs:
`E0415 17:54:20.388462   20666 plugin.go:233] Failed getting credential from external registry credential provider: error execing credential provider plugin ecr-credential-provider for image xxxxxxx.dkr.ecr.eu-west-1.amazonaws.com/xxxxxxxxx: exit status 1`

**What you expected to happen**:
We expected kubelet to be able to get ECR credentials and pull required image

**How to reproduce it (as minimally and precisely as possible)**:
The ecr-credential-provider binary installed on the worker node in /usr/local/bin directory.

Credential provider config file /var/lib/kubernetes/ecr-credential-provider-config.yaml created with the following content:
```
kind: CredentialProviderConfig
apiVersion: kubelet.config.k8s.io/v1
providers:
  - name: ecr-credential-provider
    apiVersion: credentialprovider.kubelet.k8s.io/v1
    matchImages:
      - "*.dkr.ecr.*.amazonaws.com"
    defaultCacheDuration: 1m
```
Kubelet configured with the following parameters:
```
  --image-credential-provider-bin-dir=/usr/local/bin \
  --image-credential-provider-config=/var/lib/kubernetes/ecr-credential-provider-config.yaml \
```

**Anything else we need to know?**:
Further investigation shown that it fails validating the GroupName. The following code accepts only the `credentialprovider.kubelet.k8s.io/v1alpha1` apiVersion.
```
	if gvk.GroupVersion() != v1alpha1.SchemeGroupVersion {
		return fmt.Errorf("group version %s is not supported", gvk.GroupVersion())
	}
```

However, since kubernetes 1.26 APIs [were updated to v1](https://github.com/kubernetes/kubernetes/pull/111616) (although with no API changes).

**Environment**:
- Kubernetes version (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.1", GitCommit:"4c9411232e10168d7b050c49a1b59f6df9d7ea4b", GitTreeState:"clean", BuildDate:"2023-04-14T13:21:19Z", GoVersion:"go1.20.3", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v5.0.1
[...]

PR #597, opened 2023-04-16
Kubelet 1.27.x failing to get credentials when credentialprovider.kubelet.k8s.io/v1 api is used

ikolomiyets:
**What type of PR is this?**
>
> /kind bug

CredentialProvider API has been updated to v1 since 1.26 however the ecr-credential-provider still expects `credentialprovider.kubelet.k8s.io/v1alpha1` version.

Fixes # 596

```release-note

- Credential Provider Config file should be modified in the existing installations to reflect changes in the API version

```

### DIFF
--- a/cmd/ecr-credential-provider/main.go
+++ b/cmd/ecr-credential-provider/main.go
@@ -21,6 +21,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"k8s.io/kubelet/pkg/apis/credentialprovider/v1"
 	"net/url"
 	"os"
 	"regexp"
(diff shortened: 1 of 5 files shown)


cmd/ecr-credential-provider/main.go
```diff
@@ -21,6 +21,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"k8s.io/kubelet/pkg/apis/credentialprovider/v1"
```
dims, line 24: Can you please move this line (24) to line 36?

dims:
xref: https://github.com/kubernetes/cloud-provider-aws/issues/596

cartermckinnon (approved):
Looks fine to me. Ideally we'd continue supporting the other API versions, as Kubelet does, but it'd muck up the code significantly and this is easy enough to handle when upgrading to 1.27.

nckturner:
> Looks fine to me. Ideally we'd continue supporting the other API versions, as Kubelet does, but it'd muck up the code significantly and this is easy enough to handle when upgrading to 1.27.

Yeah it forces an upgrade as is, but ... I think that's ok?  We do need to call it out in release notes though.

PR #598, opened 2023-04-17
fix: handle subnets with resource based naming more flexibly

tzneal:
**What type of PR is this?**
/kind bug

This adds support for RHEL based distributions that typically use a FQDN.

Fixes #

See cloud-init for RHEL at https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/rhel.py#L46
and https://github.com/canonical/cloud-init/blob/f146fe71733e72b94fad525b8cc9988b1405e760/cloudinit/distros/__init__.py#L343

```release-note
NONE
```

### DIFF
--- a/docs/prerequisites.md
+++ b/docs/prerequisites.md
@@ -6,7 +6,7 @@ AWS supports two naming conventions: [IP-based or resource-based naming](https:/
 
 When _IP-based naming_ is used, the nodes must be named after the instance followed by the regional domain name (`ip-xxx-xxx-xxx-xxx.ec2.<region>.internal`). If you have custom domain name set in the DHCP options, you must set `--hostname-override` on kube-proxy and kubelet to match the above-mentioned naming convention.
 
-When _resource based naming_ is used, the node must be named after the instance without any domain name (`i-1234567890abcdefg`). Custom domain name may be used as long as the output of `hostname` does not include the domain name. `--hostname-override` should not be set on any components when using resource-based naming.
+When _resource based naming_ is used, the node must be named after the instance either with or without a domain name (`i-1234567890abcdefg` or `i-1234567890abcdefg.<region>.compute.internal`). A custom domain name, configured through DHCP options, may also be used. 
 
 ## IAM Policies
 For the `aws-cloud-controller-manager` to be able to communicate to AWS APIs, you will need to create a few IAM policies for your EC2 instances. The control plane (formerly master) policy is a bit open and can be scaled back depending on the use case. Adjust these based on your needs.
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -5124,6 +5124,12 @@ func nodeNameToIPAddress(nodeName string) string {
 
 func (c *Cloud) nodeNameToInstanceID(nodeName types.NodeName) (InstanceID, error) {
 	if strings.HasPrefix(string(nodeName), rbnNamePrefix) {
+		// depending on if you use a RHEL (e.g. AL2) or Debian (e.g. standard Ubuntu) based distribution, the
+		// hostname on the machine may be either i-00000000000000001 or i-00000000000000001.region.compute.internal.
+		// This handles both scenarios by returning anything before the first '.' in the node name if it has an RBN prefix.
+		if idx := strings.IndexByte(string(nodeName), '.'); idx != -1 {
+			return InstanceID(nodeName[0:idx]), nil
+		}
 		return InstanceID(nodeName), nil
 	}
 	if len(nodeName) == 0 {
--- a/pkg/providers/v1/aws_test.go
+++ b/pkg/providers/v1/aws_test.go
@@ -560,14 +560,14 @@ func testHasNodeAddress(t *testing.T, addrs []v1.NodeAddress, addressType v1.Nod
 	t.Errorf("Did not find expected address: %s:%s in %v", addressType, address, addrs)
 }
 
-func makeInstance(num string, privateIP, publicIP, privateDNSName, publicDNSName string, ipv6s []string, setNetInterface bool) ec2.Instance {
+func makeInstance(instanceID string, privateIP, publicIP, privateDNSName, publicDNSName string, ipv6s []string, setNetInterface bool) ec2.Instance {
 	var tag ec2.Tag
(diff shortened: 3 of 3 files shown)


tzneal:
CC @nckturner

olemarkus:
Looks good. But perhaps update the prerequisite docs to reflect this?

tzneal:
> Looks good. But perhaps update the prerequisite docs to reflect this?

Updated, I tried to clarify it a bit.  Unfortunately between DHCP Options, RBN and OS config, I think there may be some situations where the hostname override is still needed.

pkg/providers/v1/aws_test.go
```diff
@@ -604,10 +604,10 @@ func makeInstance(num int, privateIP, publicIP, privateDNSName, publicDNSName st
 func TestNodeAddressesByProviderID(t *testing.T) {
 	// Note instance0 and instance1 have the same name
 	// (we test that this produces an error)
-	instance0 := makeInstance(0, "192.168.0.1", "1.2.3.4", "instance-same.ec2.internal", "instance-same.ec2.external", nil, true)
-	instance1 := makeInstance(1, "192.168.0.2", "", "instance-same.ec2.internal", "", nil, false)
-	instance2 := makeInstance(2, "192.168.0.1", "1.2.3.4", "instance-other.ec2.internal", "", nil, false)
-	instance3 := makeInstance(3, "192.168.0.3", "", "instance-ipv6.ec2.internal", "", []string{"2a05:d014:aa7:911:fc7e:1600:fc4d:ab2", "2a05:d014:aa7:911:9f44:e737:1aa0:6489"}, true)
+	instance0 := makeInstance("00000000000000000", "192.168.0.1", "1.2.3.4", "instance-same.ec2.internal", "instance-same.ec2.external", nil, true)
```
nckturner, line 607: This would be a good opportunity to convert this test to be table driven, which would make it much more readable.

pkg/providers/v1/aws.go
```diff
@@ -5124,6 +5124,12 @@ func nodeNameToIPAddress(nodeName string) string {
 
 func (c *Cloud) nodeNameToInstanceID(nodeName types.NodeName) (InstanceID, error) {
 	if strings.HasPrefix(string(nodeName), rbnNamePrefix) {
+		// depending on if you use a RHEL or Debian based distribution, the hostname on the machine may be either
```
nckturner, line 5127: In the comment, can you make it clear which distro has which name format?  I think its useful information.

nckturner:
Some nits, otherwise LGTM.

docs/prerequisites.md
```diff
@@ -6,7 +6,7 @@ AWS supports two naming conventions: [IP-based or resource-based naming](https:/
 
 When _IP-based naming_ is used, the nodes must be named after the instance followed by the regional domain name (`ip-xxx-xxx-xxx-xxx.ec2.<region>.internal`). If you have custom domain name set in the DHCP options, you must set `--hostname-override` on kube-proxy and kubelet to match the above-mentioned naming convention.
```
cartermckinnon, line 7: Maybe revisit these docs in a separate PR, but this isn't accurate. The `--hostname-override` shouldn't always be set to match the `$REGION.compute.internal` domain (example here is also wrong btw), it should be set to whatever convention `PrivateDnsName` is using, depending the VPC's DNS attributes.

tzneal: That's a question I had, how does RBN on regular Ubuntu work currently where you get a node name of `i-01234567890`, which isn't the `PrivateDnsName`? Agree it can wait until we get a really good handle on documenting this carefully.

pkg/providers/v1/aws.go
tzneal: Its complicated as it can vary. Upstream Ubuntu uses a non-FQDN, but EKS Ubuntu has a FQDN. I tried to clarify more.

pkg/providers/v1/aws_test.go
tzneal: Done, review carefully, especially the IPv6 part.  The test seemed odd to me.

docs/prerequisites.md
cartermckinnon: That would work if your `configmap/aws-auth` had a mapping from the node role to `system:node:{{SessionName}}`, because the session name is the instance ID. With the standard `system:node:{{EC2PrivateDnsName}}`, it wouldn't work (FWICT)

olemarkus: I think this is all about CCM supporting the distros own hostname convention and not what AWS may put in the metadata API or similar. There may be edge cases, although I cannot really think of one, but `--hostname-override` is already discouraged and it complicates node bootstrapping.

olemarkus (changes requested):

docs/prerequisites.md
tzneal: I think it's more than just the hostname convention.  e.g. EKS creates an aws-auth configmap of:

```
apiVersion: v1
data:
  mapRoles: |
    - groups:
      - system:bootstrappers
      - system:nodes
      rolearn: arn:aws:iam::12345678901:role/role-name
      username: system:node:{{EC2PrivateDNSName}}
```
So the hostname must match the EC2PrivateDNSName using the default config.

olemarkus: Using the name from the EC2 API is one of the things we want to explicitly avoid. So it sounds like this is something that may be changed on the EKS side to match what we are doing here.

pkg/providers/v1/aws_test.go
```diff
@@ -560,14 +560,14 @@ func testHasNodeAddress(t *testing.T, addrs []v1.NodeAddress, addressType v1.Nod
 	t.Errorf("Did not find expected address: %s:%s in %v", addressType, address, addrs)
 }
 
-func makeInstance(num int, privateIP, publicIP, privateDNSName, publicDNSName string, ipv6s []string, setNetInterface bool) ec2.Instance {
+func makeInstance(num string, privateIP, publicIP, privateDNSName, publicDNSName string, ipv6s []string, setNetInterface bool) ec2.Instance {
```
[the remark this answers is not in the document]
tzneal: SGTM

tzneal:
I believe that the comments are addressed, please review at your convenience.

PR #607, opened 2023-05-19
Add version flag to ecr-credential-provider

cartermckinnon:
**What type of PR is this?**

/kind feature

The `ecr-credential-provider` has recently changed the API version it supports (#597). Currently, there is no straightforward way to distinguish two `ecr-credential-provider` binaries, which makes changes like #597 difficult to handle. This PR adds a `--version` flag that returns the standard `gitVersion` as in `k8s.io/component-base/version`, to help in such situations.

~~I ran `go mod vendor`, which did...a lot of things. Should I back those changes out?~~ Removed.

Yes.
```release-note
A `--version` flag has been added to the `ecr-credential-provider`.
```

### DIFF
--- a/cmd/ecr-credential-provider/main.go
+++ b/cmd/ecr-credential-provider/main.go
@@ -31,10 +31,12 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ecr"
 	"github.com/aws/aws-sdk-go/service/ecrpublic"
+	"github.com/spf13/cobra"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/component-base/logs"
 	"k8s.io/klog/v2"
-	"k8s.io/kubelet/pkg/apis/credentialprovider/v1"
+	v1 "k8s.io/kubelet/pkg/apis/credentialprovider/v1"
 )
 
 const ecrPublicRegion string = "us-east-1"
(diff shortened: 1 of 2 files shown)


dims:
@cartermckinnon love the idea. please drop the `vendor/` directory?

dims:
/ok-to-test

can you please paste output of both `ecr-credential-provider` and `ecr-credential-provider --version` so we can see how it looks?

cartermckinnon:
Example output:
```
> ecr-credential-provider --help
ECR credential provider for kubelet

Usage:
  ecr-credential-provider [flags]

Flags:
  -h, --help      help for ecr-credential-provider
  -v, --version   version for ecr-credential-provider
```

```
> ecr-credential-provider --version
ecr-credential-provider version v1.27.1-8-gb80e8ef-dirty
```

dims:
/approve

thanks @cartermckinnon

cmd/ecr-credential-provider/main.go
```diff
@@ -238,9 +240,28 @@ func parseRepoURL(image string) (string, string, string, error) {
...
 	}
 }
+
+var gitVersion string
+
+func newCredentialProviderCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "ecr-credential-provider",
+		Short:   "ECR credential provider for kubelet",
+		Version: gitVersion,
```
cartermckinnon, line 257: I let cobra handle the management of the flag instead of using `k8s.io/component-base/verflag`. The flag registered by that package:
1. Creates output like `Kubernetes version v1.27.0-xxxxxx` which I found strange for this binary.
2. Registers version formats we don't currently use/populate (e.g. `raw`).

I also like that cobra will omit registering the flag if `gitVersion` is an empty string (unset in ldflags), so there's no need to have a weird default `gitVersion` like `v0.0.0-abcdefg`.

dims: @cartermckinnon yep, makes sense. +1

PR #609, opened 2023-06-01
Release 1.27.2

kmala:
Release 1.27.2

```release-note
NONE
```


mmerkes:
Tx for putting this together. This does include [adds support for authenticating public ECR registries in the credential provider](https://github.com/kubernetes/cloud-provider-aws/pull/603). Not sure if you should call that out or not under user facing changes.

Otherwise, LGTM.

PR #629, opened 2023-07-19
Update CCM 1.23 k8s deps to 1.23.17

cheeseandcereal:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Updating the kubernetes and related transitive dependencies to latest released 1.23 kubernetes version

Fixes #

```release-note
NONE
```


M00nF1sh (approved):
/lgtm
/approve

Dockerfile
```diff
@@ -14,15 +14,15 @@
 ##                               BUILD ARGS                                   ##
 ################################################################################
 # This build arg allows the specification of a custom Golang image.
-ARG GOLANG_IMAGE=golang:1.17.5
+ARG GOLANG_IMAGE=golang:1.19.6
```
M00nF1sh, line 17: should we compile with the latest golang version like v1.19.11? otherwise there might be cves by Golang

cheeseandcereal: I didn't want to use a different image here than what we use for the DISTROLESS_IMAGE which uses `go-runner`

It looks like go-runner doesn't have the latest go versions - https://github.com/kubernetes/k8s.io/blob/main/k8s.gcr.io/images/k8s-staging-build-image/images.yaml#L438-L525

I can update this to 1.19.7 for now, at least, since go-runner has an image for that, but not 1.19.11 unless diverging this image, and the DISTROLESS_IMAGE is fine

PR #631, opened 2023-07-19
Update CCM 1.26 k8s deps to 1.26.6

cheeseandcereal:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Updating the kubernetes and related transitive dependencies to latest released 1.26 kubernetes version

Fixes #

```release-note
NONE
```


.ko.yaml
```diff
@@ -1 +1 @@
-defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.19.4-bullseye.0
+defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.19.7-bullseye.0
```
cartermckinnon, line 1: 1.26 is on an older go version than 1.25?

cheeseandcereal: This particular patch version is. We're also discussing maybe moving away from this go-runner image since it seems to be quite far behind in availability for new golang versions. It only exists for golang 1.19 up to 1.19.7 and 1.20 up to 1.20.2 today, which are both older.

cartermckinnon: Gotcha, makes sense.

PR #639, opened 2023-08-14
Add IPv6 addresses to node addresses list

### LINKED ISSUE #638: Support for IPv6/dualstack

I'd like to start using dualstack in our kubernetes cluster via the `CloudDualStackNodeIPs` feature gate.
Trying to do so I get errors such as:
```
I0814 03:40:03.764807       1 node_controller.go:427] Initializing node i-0a11a57aeffb69cf7 with cloud provider
E0814 03:40:04.070450       1 node_controller.go:236] error syncing 'i-0a11a57aeffb69cf7': failed to get node modifiers from cloud provider: provided node ip for node "i-0a11a57aeffb69cf7" is not valid: failed to get node address from cloud provider that matches ip: 2600:1f10:45a5:a900:33fc:a923:65e5:9414, requeuing
```

Trying to debug the issue, I think it's because the code at
https://github.com/kubernetes/cloud-provider-aws/blob/d0551093673e8c355db17249b8f069767c014748/pkg/providers/v2/instances.go#L216C46-L216C64 doesn't look at `Ipv6Addresses`. It only iterates over the IPv4 addresses in `PrivateIpAddresses`.

**Why is this needed**:

The EC2 api returns IPv6 and IPv4 addresses in different fields.

/kind feature

james-callahan:
**What type of PR is this?**
/kind bug

Add IPv6 addresses to node addresses list in the v2 provider

Fixes #638

I'm not sure if this should be considered a bugfix or a feature.

```release-note
fix support for IPv6 addresses in v2 provider
```

### DIFF
--- a/pkg/providers/v2/instances.go
+++ b/pkg/providers/v2/instances.go
@@ -226,6 +226,20 @@ func nodeAddressesForInstance(instance *ec2.Instance) ([]v1.NodeAddress, error) 
 				})
 			}
 		}
+
+		for _, privateIP := range networkInterface.Ipv6Addresses {
+			if ipAddress := aws.StringValue(privateIP.Ipv6Address); ipAddress != "" {
+				ip := net.ParseIP(ipAddress)
+				if ip == nil {
+					return nil, fmt.Errorf("invalid IP address %q from instance %q", ipAddress, aws.StringValue(instance.InstanceId))
+				}
+
+				addresses = append(addresses, v1.NodeAddress{
+					Type:    v1.NodeInternalIP,
+					Address: ip.String(),
+				})
+			}
+		}
 	}
 
 	return addresses, nil


olemarkus:
I see this PR goes against v2, which has sort of been abandoned. This feature is already available in v1.

james-callahan:
> I see this PR goes against v2, which has sort of been abandoned. This feature is already available in v1.

I've been using v2 happily for some time now.
Last time I tried out v1 there was something that didn't work well... I can't remember what it was off the top of my head.

In any case, this PR should fix a missing piece of functionality from v2.

pkg/providers/v2/instances.go
```diff
@@ -226,6 +226,20 @@ func nodeAddressesForInstance(instance *ec2.Instance) ([]v1.NodeAddress, error)
 				})
 			}
 		}
+
+		for _, privateIP := range networkInterface.Ipv6Addresses {
```
cartermckinnon, line 230: The behavior in the v1 provider is a bit different, we only utilize the first entry in this array: https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws.go#L1574

I don't have all the context as to why that was done.

james-callahan: I'm happy to swap to that behaviour instead if that'll get the PR landed.


> I don't have the context as to why that was done.

My guess is that because the first IP cannot be modified (I vaguely recall this from somewhere?); but the others can at runtime by modifying the ENI; and the author of v1 was trying to keep it as a static list?

james-callahan:
@olemarkus is there something I can do to help get this merged?

olemarkus:
Sorry, I'm not approving PRs against v2. I only maintain v1.

pkg/providers/v2/instances.go
cartermckinnon: This is kind of related to: https://github.com/kubernetes/cloud-provider-aws/issues/656

The VPC CNI will allocate and utilize additional ENI's for pod IP's. The pod IP's are filtered out from the Node's addresses when the `provided-node-ip` annotation is present: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/node/helpers/address.go#L115-L120

But ideally we wouldn't add Pod IP's in the first place. I **think** we should only be adding the first address on the first ENI

@tzneal WDYT?

tzneal: I was wondering if we should just use the PrivateDnsName/PrivateIpAddress from the describe-instances output and ignore ENIs.  Anything we can do will break someone, maybe a flag to control what we do?

cartermckinnon: Hm yeah I think you're right; that's what we end up with in practice on EKS anyway. But I'm not sure how this would impact other CNI's, cluster addons, etc; so making it configurable is a safe bet.

cartermckinnon:
Going to close this in light of #677.

/close

PR #643, opened 2023-08-16
Update tagging controller's alias

chewong:
https://github.com/kubernetes/cloud-provider-aws/pull/641 introduced a bug where the controller alias map has the wrong value for the tagging controller:
```
"tagging" is not in the list of known controllers
Error: "tagging" is not in the list of known controllers
```

This PR will fix the issue. I have tested it in my cluster and it was able to start the cloud controller manager successfully.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/aws-cloud-controller-manager/main.go
+++ b/cmd/aws-cloud-controller-manager/main.go
@@ -78,7 +78,7 @@ func main() {
 	app.ControllersDisabledByDefault.Insert(tagging.TaggingControllerKey)
 
 	controllerAliases := names.CCMControllerAliases()
-	controllerAliases[tagging.TaggingControllerKey] = tagging.TaggingControllerClientName
+	controllerAliases[tagging.TaggingControllerKey] = tagging.TaggingControllerKey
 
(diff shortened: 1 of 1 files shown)


cmd/aws-cloud-controller-manager/main.go
```diff
@@ -78,7 +78,7 @@ func main() {
 	app.ControllersDisabledByDefault.Insert(tagging.TaggingControllerKey)
 
 	controllerAliases := names.CCMControllerAliases()
-	controllerAliases[tagging.TaggingControllerKey] = tagging.TaggingControllerClientName
+	controllerAliases[tagging.TaggingControllerKey] = tagging.TaggingControllerKey
```
wyike, line 81: Hi @chewong , regarding 1.28 KCM and CCM new feature on controller alias,  if we have some customized controller implemented in the ccm provider, should we also update its name convention to same pattern? I guess no but still want to double confirm: If we don't do this work, will it impact any functionality like logging etc.,?  Thanks.

PR #655, opened 2023-09-11
Add code to generate junit xml in the right directory

dims:
/kind bug

junit xml(s) were not being generated at all, so the test grid showed just 2 entries:
image

With this change, the following directories get created in the correct directory for testgrid to pick up from:
image


dims:
you can see the testgrid looks better now:
https://testgrid.k8s.io/provider-aws-cloud-provider-aws#e2e&width=20

PR #673, opened 2023-10-10
Remove ecr pattern restriction

### LINKED ISSUE #668: Allow `ecr-credential-provider` to generate credentials for any registry
**What would you like to be added**: Allow `ecr-credential-provider` to generate credentials for any registry (only base restrictions on `CredentialProviderConfig`)

**Why is this needed**:
Today `ecr-credential-provider`, is limited to `public.ecr.aws` , and registry url compliant with [ecrPattern](https://github.com/kubernetes/cloud-provider-aws/blob/master/cmd/ecr-credential-provider/main.go#L45). 

In order to achieve this functionality we need to remove this restriction.

**End user use case** : 
Using an EKS Cluster, we need to mirror public registry where SLA aren't garrantee or with some IP Rate limits restrictions. 

In order to achieve that, I've setuped the followings : 
- ECR Pull through cache to `public.ecr.aws` , `quay.io` and `registry.k8s.io`
- Containerd configured to mirror `public.ecr.aws` , `quay.io` and `registry.k8s.io` to my private ECR registry
- Private registry is hosted on another account (cross account scenario), on the same aws region
- aws-ecr-credential-provider configured to generate credentials for : `public.ecr.aws` , `quay.io` and `registry.k8s.io` with a config file like : 
  ```json
  {
    "apiVersion": "kubelet.config.k8s.io/v1alpha1",
    "kind": "CredentialProviderConfig",
    "providers": [
      {
        "name": "ecr-credential-provider",
        "matchImages": [
          "*.dkr.ecr.*.amazonaws.com",
          "*.dkr.ecr.*.amazonaws.com/*",
          "*.dkr.ecr.*.amazonaws.com.cn",
          "*.dkr.ecr-fips.*.amazonaws.com",
          "*.dkr.ecr.*.c2s.ic.gov",
          "*.dkr.ecr.*.sc2s.sgov.gov",
          "quay.io",
          "public.ecr.aws",
          "registry.k8s.io"
        ],
        "defaultCacheDuration": "12h",
        "apiVersion": "credentialprovider.kubelet.k8s.io/v1alpha1"
      }
    ]
   }
  ```

* Current status : No credentials has been generated, and pull through cache aren't use 
* Expected status : Credentials should be generated (for the current node , on the same aws region, or an explicit configuration on `CredentialProviderConfig`) , to allow my node to use my private registry mirror for public registry.

**Potential solution**:

I wonder if it's really needed to _generate an error here_ and not relying on SDK region behavior (and potential user configuration in provider configuration - `AWS_REGION` env variable in `CredentialProviderConfig` for exemple).

By removing program error [here](https://github.com/kubernetes/cloud-provider-aws/pull/667/files#diff-61e97abee8e39e3bb484acf33a377325f9514a670e1a541c15a8a61eb9343f79R232) , with something like : 
```suggestion
if len(splitHost) != 6 {
  return "", nil
}
```
(further code updates will be probably needed ;-) ) 
this provider will rely on AWS SDK default behaviour, and let opportunity to users to handle more uses cases.
Relying on SDK default behavior will still throw an error when no region are configured (Eg. : `E0922 13:04:35.558288   57050 main.go:255] Error running credential provider plugin: MissingRegion: could not find region configuration` ) 

Ref.: https://github.com/kubernetes/cloud-provider-aws/pull/667#discussion_r1334174043

Sample working code modification : https://github.com/jBouyoud/cloud-provider-aws/commit/5a167d0ae233981db3a992dd767b1ff1f60fdcda

I'll be glad to help on this topic, please let me known how.
Regards

/kind feature

jBouyoud:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

Remove ecr pattern restriction and forward region responsability to AWS SDK default behavior. This allow users to use this binary with any registries even they configure properly their AWS SDK (`AWS_REGION` environment variable or any other configuration handled in AWS SDK)  

Fixes #668

```release-note
[ecr-credential-provider] An ECR token is returned for the default AWS region when the specified image reference doesn't match a known ECR pattern.
```

### DIFF
--- a/cmd/ecr-credential-provider/main.go
+++ b/cmd/ecr-credential-provider/main.go
@@ -62,6 +62,7 @@ type ecrPlugin struct {
 func defaultECRProvider(region string) (*ecr.ECR, error) {
 	cfg := aws.Config{}
 	if region != "" {
+		klog.Warningf("No region found in the image reference, the default region will be used. Please refer to AWS SDK documentation for configuration purpose.")
 		cfg.Region = aws.String(region)
 	}
 	sess, err := session.NewSessionWithOptions(session.Options{
--- a/cmd/ecr-credential-provider/main.go
+++ b/cmd/ecr-credential-provider/main.go
@@ -60,8 +60,12 @@ type ecrPlugin struct {
 }
 
 func defaultECRProvider(region string) (*ecr.ECR, error) {
+	cfg := aws.Config{}
+	if region != "" {
+		cfg.Region = aws.String(region)
+	}
 	sess, err := session.NewSessionWithOptions(session.Options{
-		Config:            aws.Config{Region: aws.String(region)},
+		Config:            cfg,
 		SharedConfigState: session.SharedConfigEnable,
 	})
 	if err != nil {
@@ -122,11 +126,10 @@ func (e *ecrPlugin) getPublicCredsData() (*credsData, error) {
 
 func (e *ecrPlugin) getPrivateCredsData(imageHost string, image string) (*credsData, error) {
 	klog.Infof("Getting creds for private image %s", image)
-	region, err := parseRegionFromECRPrivateHost(imageHost)
-	if err != nil {
-		return nil, err
-	}
+	var err error
+
 	if e.ecr == nil {
+		region := parseRegionFromECRPrivateHost(imageHost)
 		e.ecr, err = defaultECRProvider(region)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


cmd/ecr-credential-provider/main.go
```diff
@@ -60,8 +60,12 @@ type ecrPlugin struct {
 }
 
 func defaultECRProvider(region string) (*ecr.ECR, error) {
+	cfg := aws.Config{}
+	if region != "" {
+		cfg.Region = aws.String(region)
+	}
```
cartermckinnon, line 66: I think we need a log warning that no region was found in the image reference/the default region will be used/etc.

cartermckinnon (commented):
Change looks fine to me. I don't think the `ecr-credential-provider` needs to strictly validate the image references passed to it, because it will only be passed what the user has defined in their `--image-credential-provider-config` file.

cartermckinnon:
@jBouyoud we need a release note on this as well.

cartermckinnon:
/assign @tzneal 

can you give this a look?

olemarkus (approved):

PR #448, opened 2022-08-19
Handle InvalidInstanceID.NotFound when tagging resources

saurav-agarwalla:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**: We recently found an issue with Node Lifecycle Controller where it leaks node objects in some cases and as a result, this controller runs into errors when trying to tag the instances underlying those node objects. This PR handles those errors gracefully and ignores those resources.

**Which issue(s) this PR fixes**: N/A

**Special notes for your reviewer**: N/A

**Does this PR introduce a user-facing change?**: NONE

PR #674, opened 2023-10-13
Do not allow EC2 instance ID NotFound to succeed tagging

ndbaker1:
**What type of PR is this?**

/kind bug

Removes the graceful handling of `InvalidInstanceID.NotFound` error when attempting to `tag` an ec2 instance that has not fully come up. This has caused an issue where we've seen the tagging controller misleadingly exit successfully, not actually tagging the instance, and does not re-queue the item to (ideally) execute again once the instance becomes visible.

example log feed:
```log
tags.go:326] Couldn't find resource when trying to tag it hence skipping it, InvalidInstanceID.NotFound: The instance ID 'i-***' does not exist status code: 400, request id: ***
tagging_controller.go:299] Successfully tagged i-*** with map[aws:eks:cluster-name:***]. Labeling the nodes with tagging controller labels now.
tagging_controller.go:305] Successfully labeled node ip-***.compute.internal with map[k8s.io/cloud-provider-aws:***].
```

This behavior *does* satisfy the `untag` action, since removing the tag from a non-existing instance is a no-op, so no changes need to be made there.

Its worth mentioning the initial PR to gracefully handle this (https://github.com/kubernetes/cloud-provider-aws/pull/448) aimed to fix all cases discussed in issue https://github.com/kubernetes/cloud-provider-aws/issues/444 where the untracked `InvalidInstanceID.NotFound` errors were valid failure modes in the context of instance termination. 

### DIFF
--- a/pkg/providers/v1/tags.go
+++ b/pkg/providers/v1/tags.go
@@ -322,10 +322,6 @@ func (c *Cloud) TagResource(resourceID string, tags map[string]string) error {
 	output, err := c.ec2.CreateTags(request)
 
 	if err != nil {
-		if isAWSErrorInstanceNotFound(err) {
-			klog.Infof("Couldn't find resource when trying to tag it hence skipping it, %v", err)
-			return nil
-		}
 		klog.Errorf("Error occurred trying to tag resources, %v", err)
 		return err
 	}
@@ -346,6 +342,8 @@ func (c *Cloud) UntagResource(resourceID string, tags map[string]string) error {
 	output, err := c.ec2.DeleteTags(request)
 
 	if err != nil {
+		// An instance not found should not fail the untagging workflow as it
+		// would for tagging, since the target state is already reached.
 		if isAWSErrorInstanceNotFound(err) {
 			klog.Infof("Couldn't find resource when trying to untag it hence skipping it, %v", err)
 			return nil
--- a/pkg/providers/v1/tags_test.go
+++ b/pkg/providers/v1/tags_test.go
@@ -24,6 +24,7 @@ import (
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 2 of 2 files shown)


pkg/providers/v1/tags.go
```diff
@@ -322,10 +322,6 @@ func (c *Cloud) TagResource(resourceID string, tags map[string]string) error {
 	output, err := c.ec2.CreateTags(request)
 
 	if err != nil {
-		if isAWSErrorInstanceNotFound(err) {
-			klog.Infof("Couldn't find resource when trying to tag it hence skipping it, %v", err)
-			return nil
-		}
```
cartermckinnon, line 328: So if we return this error instead of silencing it, the workItem will be re-queued and presumably we'll successfully tag the instance after the API becomes consistent?

How long did we observe that to take in this scenario?

cartermckinnon:
Change looks fine to me. IIUC, silencing this error in #448 was just an optimization; if there is an errant `Node` in the API, we'll try to tag it `n` times, but there's no correctness issue per se?

pkg/providers/v1/tags.go
ndbaker1: yep, that's the intention 👍 

from the logs we have for this event, CCM executes the tagging work item almost exactly when the instance launches. given that when experimenting through cli 
```bash
ID=$(aws ec2 run-instances --image-id ami-07d07d65c47e5aa90 --instance-type t2.micro --query Instances[0].InstanceId --output text)
aws ec2 create-tags --resources $ID --tags Key=test,Value=value
aws ec2 describe-tags --filters Name=resource-id,Values=$ID
```
you pretty much can't encounter the issue, i think any amount of retry in place would fix the issue

hakman (approved):
Looks like a good change for better reliability. Thanks @ndbaker1!

Issue #608, opened 2023-05-23
Prefix delegation based IPAM support in cc provider extension for IPv6


philsbln:
**What would you like to be added**:

We would like to integrate an IP address management (IPAM) controller with the AWS cloud-controller-manager to assign POD addresses based on prefix delegations. Our primary use-case are IPv6 only clusters, but the concept would also work with some limitations for IPv4.

The desired process for assigning IPv6 addresses to pods is as follows:
- The cloud provider assigns globally unique and routable IPv6 addresses to VPCs/Networks, e.g., using IPAM pools in case of AWS
- When creating nodes, we delegates an IPv6 prefixes to each node, e.g., a /80 in case in case of AWS.
- The `IPAM controller` within the `cloud-controller-manager` should picks up this prefix delegation through the cloud provider API and writes the delegated prefix in the nodes `PodCIDR ` attribute. This can be implemented analogous to the GCP implementation as a custom `CloudAllocator`.
- The standard host-local CNI on the node picks up the PodCIDR and assigns addresses to pods.
 
**Why is this needed**:

While adding IPv6 support to the [gardener project](https://github.com/gardener/), we strive to get globally unique IPv6 addresses across all our clusters.
We would prefer to integrate the IPv6 IPAM functionality with the cloud provider's as much as possible as using provider managed IPv6 space also eliminates the need for NAT or routing hacks. 

The functionality to use delegated prefixes as PodCIDR is already implemented in [amazon-vpc-cni-k8s](https://github.com/aws/amazon-vpc-cni-k8s/tree/master), which requires api keys to read/add prefix delegations to be present on the nodes. This implementation was reasonable for IPv4, where nodes needed dynamically add multiple prefix delegations in order to preserve precious address space. For IPv6, we only need a single prefix delegation and can add this one at the time we create the node, thus, eliminating the risk to expose API keys through a compromised node and the need to deploy the vpc-cni to the nodes.

/kind feature

olemarkus:
This sounds identical to what kOps does, allowing IPv6 prefix support regardless of which CNI is used. We have been discussing if the kOps address controller should be moved to CCM before.

philsbln:
> This sounds identical to what kOps does, allowing IPv6 prefix support regardless of which CNI is used. We have been discussing if the kOps address controller should be moved to CCM before.

[...]

PR #679, opened 2023-10-17
Add/nodeipamcontroller

DockToFuture:
**What type of PR is this?**
 /kind feature

This change adds a `nodeipam controller` to the aws cloud-controller-manager for IPv6 prefix delegation. The node ipam controller assigns POD addresses to kubernetes nodes based on prefix delegations. The controller supports plain IPv6 (`--dualstack=false`) and dual stack (`--dualstack=true`). In dual stack mode the IPAM for IPv4 is done in the same way it was done in the kube-controller-manager.

Fixes https://github.com/kubernetes/cloud-provider-aws/issues/608

```release-note
This change adds a `nodeipam controller` to the aws cloud-controller-manager for IPv6 prefix delegation. The node ipam controller assigns POD addresses to kubernetes nodes based on prefix delegations. The controller supports plain IPv6 (`--dualstack=false`) mode and dual stack (`--dualstack=true`) mode. 
```


dims:
@DockToFuture how did you test this? I can't spot any tests in this PR, are parts of this controller here borrowed from other places? could we borrow the tests that exist there as well?

DockToFuture:
Hello @dims I've deployed the controller into our IPv6 Kubernetes clusters and test it with the following controllers(`--controllers=cloud-node,cloud-node-lifecycle,nodeipam,service,route`).
The code is similar to https://github.com/kubernetes/cloud-provider-gcp/blob/master/cmd/cloud-controller-manager/nodeipamcontroller.go . What for tests do you expect? Can you elaborate on it a little bit?

dims:
@DockToFuture so what i was thinking was if we were to enhance one of the existing harness(es) to enable this control, for example `make test-e2e` uses `kops` to setup a cluster and run a bunch of tests
- https://github.com/kubernetes/cloud-provider-aws/blob/master/Makefile#L177-L184
- https://github.com/kubernetes/cloud-provider-aws/blob/master/hack/e2e/run.sh#L124-L134

if we take this harness (switch on ipv6!), add a way to ensure that this new controller is activated/enabled that would be the first step. then we could add some basic golang based tests in test/e2e folder to ensure we don't break this controller going forward. You called out `The node ipam controller assigns POD addresses to kubernetes nodes based on prefix delegations` in your PR body as the feature, so that's what we would test, launch pods and test if the ip addresses look right.

Does that make sense?

dims:
@DockToFuture also there are some unit tests there in that other repo you pointed to, some of it may be ported over here as well? 
https://cs.k8s.io/?q=package&i=nope&files=.*nodeipam.*_test%5C.go&excludeFiles=&repos=kubernetes/cloud-provider-gcp

olemarkus:
It should be possible to use this as a drop-in replacement to the ipam controller we implemented in kOps. 
I expect kOps would still handle assigning the actual prefixes to instances.

kOps would not support dualstack though, so it may be an idea to mark that option experimental unless there is another way of getting that e2e-tested.

DockToFuture:
> @DockToFuture also there are some unit tests there in that other repo you pointed to, some of it may be ported over here as well? https://cs.k8s.io/?q=func%20Test&i=nope&files=.*nodeipam.*_test%5C.go&excludeFiles=&repos=kubernetes/cloud-provider-gcp

I've added all relevant unit tests and adapted them where necessary. I've also added some unit tests for the new functionality.

DockToFuture:
> It should be possible to use this as a drop-in replacement to the ipam controller we implemented in kOps. I expect kOps would still handle assigning the actual prefixes to instances.
> 
> kOps would not support dualstack though, so it may be an idea to mark that option experimental unless there is another way of getting that e2e-tested.

I've marked the flag as experimental.

DockToFuture:
> * https://github.com/kubernetes/cloud-provider-aws/blob/master/Makefile#L177-L184

What is the best way to do that? Create a cluster with kops and afterwards edit the deployment of the cloud-controller-manager in the e2e test or is there a way to directly set the flags for the cloud-controller-manager in the kops cluster creation call? I didn't find any information how this could be done. Any ideas?

DockToFuture:
> @DockToFuture so what i was thinking was if we were to enhance one of the existing harness(es) to enable this control, for example `make test-e2e` uses `kops` to setup a cluster and run a bunch of tests
> 
> * https://github.com/kubernetes/cloud-provider-aws/blob/master/Makefile#L177-L184
> * https://github.com/kubernetes/cloud-provider-aws/blob/master/hack/e2e/run.sh#L124-L134
> 
> if we take this harness (switch on ipv6!), add a way to ensure that this new controller is activated/enabled that would be the first step. then we could add some basic golang based tests in test/e2e folder to ensure we don't break this controller going forward. You called out `The node ipam controller assigns POD addresses to kubernetes nodes based on prefix delegations` in your PR body as the feature, so that's what we would test, launch pods and test if the ip addresses look right.
> 
> Does that make sense?


What is the best way to do that? Create a cluster with kops and afterwards edit the deployment of the cloud-controller-manager in the e2e test or is there a way to directly set the flags for the cloud-controller-manager in the kops cluster creation call? I didn't find any information how this could be done. Any ideas?

DockToFuture:
When creating a kops cluster with `--ipv6` flag, the networking CIDR is set automatically to the IPv4 CIDR of the VPC`172.20.0.0/16`.
This leads to worker nodes not joining the cluster because IPv4 addresses are none routable in the private IPv6 subnets and thus the bootstrapping fails.
```
Dec 08 13:08:45 i-xxxxxxxxxxxxxxx nodeup[1347]: W1208 13:08:45.148005    1347 main.go:133] got error running nodeup (will retry in 30s): failed to get node config from server: Post "https://172.20.1.200:3988/bootstrap": dial tcp 172.20.1.200:3988: connect: network is unreachable; Post "https://172.20.6.164:3988/bootstrap": dial tcp 172.20.6.164:3988: connect: network is unreachable; Post "https://172.20.9.26:3988/bootstrap": dial tcp 172.20.9.26:3988: connect: network is unreachable
```
When editing the cluster config and setting it to the IPv6 CIDR of the VPC, kops throws an error:
```unexpected IP address type: xxxx:xxxx:xxxx:xx00::/56```.
How can I get kops to use the correct IPv6 CIDR range of the VPC? Did this even ever work?

**Comment by DockToFuture** [comment]
I've update the PR with an e2e test. A valid domain needs to be provided to run the e2e tests successfully. It's required to get the kops cluster up and running as the used gossip implementation does not support IPv6 yet. It can be set via`DNS_ZONE` environment variable. I've used a route53 hosted zone domain for testing.

**Comment by DockToFuture** [comment]
PR is now ready for review! :)

File: hack/e2e/run.sh [hunk 1]
```diff
@@ -143,10 +145,62 @@ fi
 
 set -x
 pushd ./tests/e2e
-ginkgo . -v -p --nodes="${GINKGO_NODES}" --focus="${GINKGO_FOCUS}" --skip="${GINKGO_SKIP}" --report-dir="${ARTIFACTS}"
+ginkgo --focus="${GINKGO_FOCUS}" --skip="${GINKGO_SKIP}" --label-filter="${GINKGO_LABEL_FILTER}" . -v -p --nodes="${GINKGO_NODES}"  --report-dir="${ARTIFACTS}"
```
M00nF1sh, line 148: why we added a filter for loadbalancer here

pkg/controllers/nodeipam/config/types.go
```diff
@@ -0,0 +1,30 @@
...
+limitations under the License.
+*/
+
+package config
+
+import "net"
+
+// NodeIPAMControllerConfiguration contains elements describing NodeIPAMController.
+type NodeIPAMControllerConfiguration struct {
+	RateLimit    float64
```
M00nF1sh, line 23: we need some documentation & better name for those fields.
e.g. what's the `RateLimit` for? could there be other "RateLimit" in the future than this one that need to be configured

hack/e2e/run.sh
DockToFuture: Because I only want to run the `loadbalancer` test here and not the `ipv6 prefix` test. In general we could run both tests here and they will succeed but we will get nothing out of it.

DockToFuture:
Any more feedback @M00nF1sh?

dims:
@DockToFuture please rebase and i'll merge. thanks!

DockToFuture:
Hello @dims, I've rebased the PR.

dims:
looks like failure in `pull-cloud-provider-aws-e2e` is at cleanup phase and the test passed

```
[38;5;10m[1mSUCCESS![0m -- [38;5;10m[1m1 Passed[0m | [38;5;9m[1m0 Failed[0m | [38;5;11m[1m0 Pending[0m | [38;5;14m[1m1 Skipped[0m
PASS

Ginkgo ran 1 suite in 1m28.931164668s
Test Suite Passed
```

DockToFuture:
@dims a valid dns-zone needs to be set in the environment variable "DNS_ZONE" in order to create the kops cluster successfully, see my earlier comment: https://github.com/kubernetes/cloud-provider-aws/pull/679#issuecomment-1881073904 for more information.

```
Error: cannot find DNS Zone "example.com".  Please pre-create the zone and set up NS records so that it resolves
Error: exit status 1
make: *** [Makefile:182: test-e2e] Error 1
```

dims:
@M00nF1sh any suggestions on how to make this work?

hack/e2e/run.sh
M00nF1sh: seems we now have a node test suit: https://github.com/kubernetes/cloud-provider-aws/blob/master/tests/e2e/nodes.go

this will be a regression that we skipped the "nodes" test suit when run e2e.
i think we shouldn't have this label-filter, and should run all tests we have.

pkg/controllers/nodeipam/ipam/cidr_allocator.go
```diff
@@ -0,0 +1,480 @@
...
+	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "cidrAllocator"})
+	eventBroadcaster.StartStructuredLogging(0)
+	klog.V(0).Infof("Sending events to api server.")
+	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: client.CoreV1().Events("")})
+
+	// create a cidrSet for each CIDR we operate on.
+	// cidrSet are mapped to clusterCIDR by index
+	cidrSets := make([]*cidrset.CidrSet, len(allocatorParams.ClusterCIDRs))
+	for idx, cidr := range allocatorParams.ClusterCIDRs {
+		cidrSet, err := cidrset.NewCIDRSet(cidr, allocatorParams.NodeCIDRMaskSizes[idx])
```
M00nF1sh, line 170: nit, this can cause crash if the len(allocatorParams.NodeCIDRMaskSizes) <  len(allocatorParams.ClusterCIDRs).
It's a nit comment given this shall be supplied by cluster admins so lt's low risk, but good to have validations with clear error message when len(allocatorParams.NodeCIDRMaskSizes) != len(allocatorParams.ClusterCIDRs)

pkg/controllers/nodeipam/ipam/cidr_allocator.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+	for idx, cidr := range node.Spec.PodCIDRs {
+		_, podCIDR, err := net.ParseCIDR(cidr)
+		if err != nil {
+			return fmt.Errorf("failed to parse node %s, CIDR %s", node.Name, node.Spec.PodCIDR)
+		}
+		// If node has a pre allocate cidr that does not exist in our cidrs.
+		// This will happen if cluster went from dualstack(multi cidrs) to non-dualstack
+		// then we have now way of locking it
+		if idx >= len(r.cidrSets) {
```
M00nF1sh, line 304: nit, we can optimize the performance and readability by doing the check once before the loop
e.g. 
```
if len(node.Spec.PodCIDRs) != len(r.cidrSets) {
   return "some error"
}
for idx, cidr := range node.Spec.PodCIDRs {
 .....original logic without the idx >= len(r.cidrSets) check
}
```

pkg/controllers/nodeipam/ipam/cidr_allocator.go
```diff
@@ -0,0 +1,480 @@
...
+	for idx, cidr := range node.Spec.PodCIDRs {
+		_, podCIDR, err := net.ParseCIDR(cidr)
+		if err != nil {
+			return fmt.Errorf("failed to parse CIDR %s on Node %v: %v", cidr, node.Name, err)
+		}
+
+		// If node has a pre allocate cidr that does not exist in our cidrs.
+		// This will happen if cluster went from dualstack(multi cidrs) to non-dualstack
+		// then we have now way of locking it
+		if idx >= len(r.cidrSets) {
```
M00nF1sh, line 367: same as previous comment, i'd like to move this check before the loop for performance and readability.
the len(node.spec.PodCIDRs)(when it's not 0) must always **equal to**(instead of `<=`)  len(r.cidrSets)

pkg/controllers/nodeipam/ipam/cidr_allocator.go
```diff
@@ -0,0 +1,480 @@
...
+		}
+	}
+	return nil
+}
+
+// WARNING: If you're adding any return calls or defer any more work from this
+// function you have to make sure to update nodesInProcessing properly with the
+// disposition of the node when the work is done.
+func (r *rangeAllocator) AllocateOrOccupyCIDR(node *v1.Node) error {
+	if node == nil {
```
M00nF1sh, line 319: just curious, when will node been nil? or is this sort of Defensive programming(if so, let's add a comment like this should never happen for readability) or remove this check if we can ensure it's non-nil.

pkg/controllers/nodeipam/ipam/cidr_allocator.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+	// allocate and queue the assignment
+	allocated := NodeReservedCIDRs{
+		nodeName:       node.Name,
+		allocatedCIDRs: make([]*net.IPNet, len(r.cidrSets)),
+	}
+
+	for idx := range r.cidrSets {
+		podCIDR, err := r.cidrSets[idx].AllocateNext()
+		if err != nil {
```
M00nF1sh, line 338: if this happens, seems the allocations for other index(< current idx) will be leaked

pkg/controllers/nodeipam/ipam/cidr_allocator.go
```diff
@@ -0,0 +1,480 @@
...
+
+		// If node has a pre allocate cidr that does not exist in our cidrs.
+		// This will happen if cluster went from dualstack(multi cidrs) to non-dualstack
+		// then we have now way of locking it
+		if idx >= len(r.cidrSets) {
+			return fmt.Errorf("node:%s has an allocated cidr: %v at index:%v that does not exist in cluster cidrs configuration", node.Name, cidr, idx)
+		}
+
+		klog.V(4).Infof("release CIDR %s for node:%v", cidr, node.Name)
+		if err = r.cidrSets[idx].Release(podCIDR); err != nil {
```
M00nF1sh, line 372: shall we do a best effort by try release for each cidrSet and return aggregated error?

pkg/controllers/nodeipam/ipam/cidr_allocator.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+
+	//  fetch ipv6 cidr address
+	if node.Spec.ProviderID == "" {
+		klog.Infof("Node %q has empty provider ID", node.Name)
+		return nil
+	}
+
+	// aws:///eu-central-1a/i-07577a7bcf3e576f2
+	instanceID, _ := awsv1.KubernetesInstanceID(node.Spec.ProviderID).MapToAWSInstanceID()
```
M00nF1sh, line 426: the error is ignored..what id the instance id is a non-valid instanceid? like a fargate task?

pkg/controllers/nodeipam/ipam/ipv6_allocator.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+func (ra6 *IPv6RangeAllocator) prefixNodeResource(node *v1.Node) error {
+	if node.Spec.ProviderID == "" {
+		klog.Infof("Node %q has empty provider ID", node.Name)
+		return nil
+	}
+
+	// aws:///eu-central-1a/i-07577a7bcf3e576f2
+	instanceID, _ := awsv1.KubernetesInstanceID(node.Spec.ProviderID).MapToAWSInstanceID()
```
M00nF1sh, line 227: the error is ignored.

pkg/controllers/nodeipam/nodeipam_controller.go
```diff
@@ -0,0 +1,218 @@
...
+			clusterCIDRs:          allocatorParams.ClusterCIDRs,
+			cloud:                 awsCloud,
+			cidrSets:              []*cidrset.CidrSet{},
+			nodeLister:            nodeInformer.Lister(),
+			nodesSynced:           nodeInformer.Informer().HasSynced,
+			nodeCIDRUpdateChannel: make(chan ipam.NodeReservedCIDRs, ipam.CidrUpdateQueueSize),
+			recorder:              recorder,
+			nodesInProcessing:     sets.NewString(),
+		}
+		nodeList, err := listNodes(kubeClient)
```
M00nF1sh, line 163: why we need to use listNodes? shouldn't the informer handles all existing nodes?

M00nF1sh (commented):
Another high level comment:
1. why we choose to have different implementation pattern for the "dualstack" and "ipv6only" allocator? Having different patterns is hard to maintain since technically they are doing the same work..
  * with the dualstack one, the major work is handled directly during informer's event handler, while uses a worker queue to defer the cidr updates. 
  * with the ipv6only one, the informer's event handler merely distribute the node event into a queue, and workers then handles it.


2. the cidr_allocator seems to shall be named like dualstack_allocator(or more specifically cidr_plus_ipv6prefix_allocator). it's not only allocate from preconfigured list of cidrs but also picks up allocates from existing IPv6 prefixes on node). I'm confused by the use case, why would people use both ip from configured list & ipv6 prefixes together?

DockToFuture:
I will close this PR as we have decided to implement an own IPAM controller as suggested by @M00nF1sh to keep the CloudProvider AWS as slim as possible (analogous to the AWS [loadbalancer controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.7/)). Furthermore it gives us more degrees of freedom how the tests for the IPAM controller are run and executed.

dims:
thanks @DockToFuture !

PR #709, opened 2023-10-31
update the go version

kmala:


hakman:
/hold

kmala:

 this change is not in 1.26

cartermckinnon:
>  i don't think we want to update the minor version go for older versions.

You've updated from 1.19 to 1.20 here, though; so why not just take everything to 1.21?

PR #710, opened 2023-10-31
[release-1.27] update go to 1.20.10

kmala:

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,1 +1,1 @@
-defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.20.3-bullseye.0
+defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.20.10-bullseye.0
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 ##                               BUILD ARGS                                   ##
 ################################################################################
 # This build arg allows the specification of a custom Golang image.
-ARG GOLANG_IMAGE=golang:1.20.3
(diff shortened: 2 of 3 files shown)


hakman:
/hold

kmala:

Fixed it

hakman:
> > The correct way to do this is to create a PR in the main branch and then cherry-pick it.
> 
> Yes, i understand it but the master is not having issue as it is updated to 1.21 version and i don't think we want to update the minor version go for older versions. Even the upstream update is done in similar way [kubernetes/kubernetes#121151](https://github.com/kubernetes/kubernetes/pull/121151) https://github.com/kubernetes/cloud-provider-aws/blob/master/Dockerfile#L17
> 
> > Also, there is another place to change:
> 
> Fixed it

Maybe it would help to be on go1.21 on all releases. It should be backwards compatible to my knowledge.
In any case, the idea is that we are discussing the same issue in N PRs. It should be 1 and then cherry-pick. It's pretty annoying and error prone like this.

kmala:
> Maybe it would help to be on go1.21 on all releases

yes, this could be done but the scope of this PR is to fix the CVE and keep it simple. Usually for non master branches patch releases are picked as they are less prone to errors than a minor version.

> In any case, the idea is that we are discussing the same issue in N PRs. It should be 1 and then cherry-pick. It's pretty annoying and error prone like this.

i agree about with what you mentioned and should be followed for most cases but i don't think it should be a rule as per me because, the fix here can't be made on the master. Hence i shared the k/k release branch update  https://github.com/kubernetes/kubernetes/pull/121151 where similar approach is used.

PR #711, opened 2023-10-31
[release-1.28] update go to 1.20.10

kmala:

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,1 +1,1 @@
-defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.20.3-bullseye.0
+defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.20.10-bullseye.0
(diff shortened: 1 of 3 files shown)


hakman:
/hold

kmala:

Fixed it

cartermckinnon:
@kmala you can cherry-pick from release-x to release-x as well, with this script: https://github.com/kubernetes/kubernetes/blob/master/hack/cherry_pick_pull.sh. Just usually faster than opening the PR's by hand.

/retitle [release-1.28] update go to 1.20.10
/lgtm
/approve

cartermckinnon:
After reviewing the other PR's, I think I agree with @hakman 's suggestion to just take everything to go 1.21

/remove-approve

kmala:
> After reviewing the other PR's, I think I agree with @hakman 's suggestion to just take everything to go 1.21

i had updated the master with 1.21 for go runner https://github.com/kubernetes/cloud-provider-aws/pull/720
Once its merged i will cherry-pick this and other changes to update the go version for  build  https://github.com/kubernetes/cloud-provider-aws/pull/675/files.

PR #720, opened 2023-11-01
update the go version to 1.21.3

kmala:
Bump the go version and go runner version

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,1 +1,1 @@
-defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.20.3-bullseye.0
+defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.21.3-bookworm.0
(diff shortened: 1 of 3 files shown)


.ko.yaml
```diff
@@ -1 +1 @@
-defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.20.3-bullseye.0
+defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.21.3-bullseye.0
```
cartermckinnon, line 1: Does this image exist? Looks like it should be the `bookworm` variant: https://github.com/kubernetes/k8s.io/blob/d0222fb7613e8ea893025d92b1f93d22d10b88fa/registry.k8s.io/images/k8s-staging-build-image/images.yaml#L646

kmala: My bad, updated the tags to be proper ones

hakman (approved):

PR #753, opened 2023-11-27
Drop Node event when EC2 instance does not exist

cartermckinnon:
**What type of PR is this?**

/kind bug

This avoids unnecessary retries for invalid Node events (the corresponding EC2 instance does not exist). When many Nodes are created and deleted, the event queue of the tagging controller can grow, causing delays in handling valid events. Every node is "retried" on each update event -- this PR tries to avoid repeated attempts to process a single event when possible.

If the Node is newly-created, we requeue the event and retry, to handle eventual-consistency of the EC2 CreateTags API. If the Node is not newly-created, we drop the event.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/tagging/tagging_controller.go
+++ b/pkg/controllers/tagging/tagging_controller.go
@@ -63,6 +63,9 @@ const (
 
 	// The label for depicting total number of errors a work item encounter and fail
 	errorsAfterRetriesExhaustedWorkItemErrorMetric = "errors_after_retries_exhausted"
+
+	// The period of time after Node creation to retry tagging due to eventual consistency of the CreateTags API.
+	newNodeEventualConsistencyGracePeriod = time.Minute * 5
 )
 
 // Controller is the controller implementation for tagging cluster resources.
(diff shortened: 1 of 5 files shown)


cartermckinnon:
Something seems wonky with the CI, I'll look into it.

cartermckinnon:
/retest

All failures are related to persistent volumes, doesn't seem related to this change.

tzneal (commented):
It look ok to me, but can we add a test for this new behavior in tagging_controller_test.go?

cartermckinnon:
The CI is definitely hosed, same cases have been red in k/k since ~11/24: https://testgrid.k8s.io/presubmits-ec2#pull-kubernetes-e2e-ec2

cartermckinnon:
@tzneal added a couple unit test cases 👍

cartermckinnon:
Haven't had a chance to go down the rabbit hole. Looks like things broke when the `kubekins` image was bumped in the test spec, I assume a change in cloud-provider-aws-test-infra is the issue but I don't see anything obvious in the commit log.

@dims do you have a guess?

cartermckinnon:
@dims I'll try to get a fix up 👍

mmerkes (approved):
LGTM

ndbaker1 (changes requested):
Thanks for picking this up! did we also want to make the queue size visible with some logging around [here](https://github.com/kubernetes/cloud-provider-aws/blob/11088788f267d02cf32ad177467312a11c5692ec/pkg/controllers/tagging/tagging_controller.go#L205)? (since dequeuing latency isn't the most direct metric)

cartermckinnon:
> did we also want to make the queue size visible

I plan to add a metric for this in a separate PR, because it'd be helpful to debug in the future; but I think dequeue latency is still the more important metric to track and alarm on. There can be many events in the queue that are no-ops, and that doesn't necessarily have an impact e.g. how quickly a new Node is tagged.

ndbaker1 (approved):

PR #754, opened 2023-12-01
Enable prometheus provider for workqueue metrics

cartermckinnon:
**What type of PR is this?**

/kind feature

This enables all the internal workqueue metrics. They're registered to the `legacyregistry` as the existing metrics in the tagging controller are.

Fixes #

I changed the name of the `workqueue` instance. Before this change, I don't think this is really being exposed in any way, so shouldn't matter.

```release-note
The tagging controller's `workqueue`now registers its internal metrics.
```

### DIFF
--- a/pkg/controllers/tagging/tagging_controller.go
+++ b/pkg/controllers/tagging/tagging_controller.go
@@ -32,8 +32,13 @@ import (
 	opt "k8s.io/cloud-provider-aws/pkg/controllers/options"
 	awsv1 "k8s.io/cloud-provider-aws/pkg/providers/v1"
 	nodehelpers "k8s.io/cloud-provider/node/helpers"
+	_ "k8s.io/component-base/metrics/prometheus/workqueue" // enable prometheus provider for workqueue metrics
 	"k8s.io/klog/v2"
 )
+
+func init() {
+	registerMetrics()
+}
 
 // workItem contains the node and an action for that node
 type workItem struct {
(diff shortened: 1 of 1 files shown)


pkg/controllers/tagging/tagging_controller.go
```diff
@@ -126,14 +131,13 @@ func NewTaggingController(
 		rateLimitEnabled = false
 	}
 
-	registerMetrics()
```
cartermckinnon, line 129: I noticed this was here, it belongs in `init` IMO.

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -32,9 +32,14 @@ import (
 	opt "k8s.io/cloud-provider-aws/pkg/controllers/options"
 	awsv1 "k8s.io/cloud-provider-aws/pkg/providers/v1"
 	nodehelpers "k8s.io/cloud-provider/node/helpers"
+	_ "k8s.io/component-base/metrics/prometheus/workqueue" // enable prometheus provider for workqueue metrics
```
ndbaker1, line 35: oh cool, didn't realize this was all done with side effects via `init`

ndbaker1 (approved):

pkg/controllers/tagging/tagging_controller.go
cartermckinnon: Yeah assuming these built-in metrics do the job, I'd like to get rid of the other ones we have after 1.29 is cut. The `workqueue` should track processing time (dequeue latency), errors, size, etc.

cartermckinnon:
/assign @hakman 

Can you take a look at this? I'd like to get it into the first 1.29 release

hakman (approved):

PR #773, opened 2023-12-15
Ci: Add support for generating SBOM on release

Rohansjamadagni:
**What type of PR is this?**

/kind documentation

This PR adds Software Bill of Materials (SBOM) Generation to the release action. This change is part of the Kubernetes Security Slam Event, where we improve the security documentation of the project.   

This change has not been tested, and must be verified by the maintainers to ensure that the action is generating the right SBOM. The action is part of the [Kubernetes Release SIG's github actions](https://github.com/kubernetes-sigs/release-actions).

Tagging @puerco to take a look

NONE

### DIFF
--- a/.github/workflows/helm_chart_release.yaml
+++ b/.github/workflows/helm_chart_release.yaml
@@ -20,23 +20,12 @@ jobs:
         run: |
           git config user.name "$GITHUB_ACTOR"
           git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install bom
+        uses: kubernetes-sigs/release-actions/setup-bom@dd08496c83441d6477114cc0555b96d404dacff7 # v0.1.2
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.1.0
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
         with:
           config: .github/cr.yaml
-
-  generate-sbom:
-    runs-on: ubuntu-latest
-
-    permissions: {}
-
-    name: Install bom and test presence in path
-    steps:
-      - name: Install bom
-        uses: puerco/release-actions/setup-bom@main
-        with:
-          bom-release: '0.5.1'
-      - name: Check install!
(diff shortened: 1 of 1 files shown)


.github/workflows/helm_chart_release.yaml
```diff
@@ -20,6 +20,9 @@ jobs:
         run: |
           git config user.name "$GITHUB_ACTOR"
           git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install bom
+        uses: kubernetes-sigs/release-actions/setup-bom@dd08496c83441d6477114cc0555b96d404dacff7 # v0.1.2
```
cartermckinnon, line 25: Awesome that this exists!

This release workflow is only for the helm chart, we haven't figured out the right way to automate releases for the primary artifacts of this project (`cloud-controller-manager-aws` and `ecr-credential-provider`). The binaries and container images are built after a new tag is created, and need to be promoted to the prod registry via 2 PR's, examples:
- https://github.com/kubernetes/k8s.io/pull/6195
- https://github.com/kubernetes/k8s.io/pull/6111

We shouldn't create a GitHub Release until after those PR's merge (in my opinion). And the SBOM should be attached to that release, not the helm chart's (IIUC). If we could:
1. Create a blank GitHub Release for an existing tag after image+artifact promo is complete.
2. Trigger a workflow that would update the Release with the SBOM and references to the prod images/artifacts.

cartermckinnon:
@Rohansjamadagni thanks for this!

.github/workflows/helm_chart_release.yaml
Rohansjamadagni: I see, sorry about this, I don't have the full context around where the artifacts are built. Do you think it would be better/possible if it could be added to  `cloudbuild.yaml`, or where the artifacts are built instead of adding them later?

cartermckinnon: If we could generate the SBOM in the cloudbuild job and include that in the artifacts, that'd be great!

Issue #775, opened 2023-12-15
Token Permissions Check fails in CLOMonitor


pi1814:
CLOMonitor is a tool that runs security scans to verify if the project meets the best security practices. 

While running the CLOMonitor locally for Cloud Provider AWS, we (@Rohansjamadagni, @pi1814) have found the following check fail that must be addressed:
- Token Permissions

Added the complete output file in issue. 
[cpa_formmatted.json](https://github.com/kubernetes/cloud-provider-aws/files/13689383/cpa_formmatted.json)

Steps to reproduce:
Run CLOMonitor CLI locally by running the following commands

```text
$ export GITHUB_TOKEN=<your token>

$ docker run -it \
  --volume $PWD:/repo \
  --env GITHUB_TOKEN=$GITHUB_TOKEN \
  public.ecr.aws/clomonitor/linter clomonitor-linter \
    --path /repo \
    --url https://github.com/<org>/<repo>
```

PR #779, opened 2023-12-15
Changes to set minimum token permissions for the GITHUB_TOKEN in yaml files

pi1814:
This PR is made under Kubernetes Security Slam.
This resolve the Token Permissions Check fail issue found in CLOMonitor for the project
**What type of PR is this?**

> /kind bug

#775

Fixes #

NONE

### DIFF
--- a/.github/workflows/helm_chart_release.yaml
+++ b/.github/workflows/helm_chart_release.yaml
@@ -8,8 +8,13 @@ on:
     paths:
       - "charts/**"
 
+permissions:
+  contents: read
+
 jobs:
   release:
+    permissions:
+      contents: write  # for helm/chart-releaser-action to push chart release and create a release
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -8,6 +8,9 @@ on:
     paths:
       - version.txt
 
+permissions:
+  contents: read
+
 jobs:
   tag:
     if: ${{ github.repository == 'kubernetes/cloud-provider-aws' }}


.github/workflows/tag.yml
```diff
@@ -8,6 +8,9 @@ on:
     paths:
       - version.txt
 
+permissions:
+  contents: read
```
cartermckinnon, line 12: This one needs `write`, because it has to `git push` the new tag

.github/workflows/helm_chart_release.yaml
```diff
@@ -8,8 +8,13 @@ on:
     paths:
       - "charts/**"
 
+permissions:
+  contents: read
```
cartermckinnon, line 12: Hm what's the effect of this if the only job bumps its permissions to `write`? Just covering our bases when/if another `job` is added?

.github/workflows/tag.yml
pi1814: This change is just to set the minimum default access as read for all jobs in the workflow. I think it will not affect the existing job since we already have a job specific write permission for that.

.github/workflows/helm_chart_release.yaml
pi1814: Yes, the top-level permissions tag will only provide minimum access which can be overridden by job-level permissions tag. It is placed as a security measure.
Adding reference for this: [Assigning permissions to jobs](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs)

.github/workflows/tag.yml
cartermckinnon: Oh I didn't notice the job's config, thanks.

Issue #823, opened 2024-01-11
Publish the ecr-credential-provider as a container image


kon-angelo:
**What would you like to be added**:
Requesting to publish the ecr-credential-provider as a container image.

**Why is this needed**:
 The use cases are similar to  https://github.com/kubernetes/cloud-provider-aws/issues/512 and to be able to consume the binary without needing to build it ourselves. 

The request for the container instead of just the binary is because the containers more or less provide an established way of distribution, and because it makes it easy to fulfil compliance requirements e.g. vulnerability scannings etc.

Note: I can take over the task to provide the changes (makefile, dockerfile etc.), but it would be nice if the maintainers provide a buy-in with regards to idea, and the publishing channels (official k8s.io registry, awslabs).

/kind feature

cartermckinnon:
I don't think a container image is the optimal way to distribute this tool. This binary is available on `artifacts.k8s.io`, we just need to automate/generate the release notes here to point to it, e.g.:

https://artifacts.k8s.io/binaries/cloud-provider-aws/v1.29.0/linux/amd64/ecr-credential-provider-linux-amd64

kon-angelo:
> This binary is available on artifacts.k8s.io

Thank you! I was not aware.

> I don't think a container image is the optimal way to distribute this tool.

The container approach does have some advantages. A proper multi-arch build image basically abstracts the issue of picking the correct version and it also streamlines the consumption e.g. some binaries may be compressed or not. In an environment like a k8s node where you are guaranteed to have a container runtime, the CR can offload part of the process. It's a convenience so to say.

cartermckinnon:
I do like the multi-arch magic, but having to grab the binary out of the container image is pretty awkward IMO. I can imagine some scenarios when you might want to grab the latest `ecr-credential-provider` at runtime on each node, but baking it in during the node image build process has it's benefits (the node image can be tested as a package deal, you know you're using a consistent version of the binary across all your nodes, etc.).

If you want to add the `ecr-credential-provider` to the CCM container image, I'm not opposed; but I think the preferred source should be artifacts.k8s.io (and if you're up for making that easier to discover, it'd be appreciated!)

kon-angelo:
@cartermckinnon I very much liked your idea and created a pull request that adds the ecr binary to the cloud-provider image. 

With regards to the discovery/visibility, can you please outline some more concrete ideas (here, or you can reach in slack) and let me see what I can contribute.

[...]

PR #825, opened 2024-01-13
Add ecr-credential-helper to docker image

kon-angelo:
**What type of PR is this?**
/kind feature

Adds ecr-credential-provider to the aws-cloud-controller-manager image.

Fixes https://github.com/kubernetes/cloud-provider-aws/issues/823

```release-note

```

### DIFF
--- a/cmd/ecr-credential-provider/plugin/plugin.go
+++ b/cmd/ecr-credential-provider/plugin/plugin.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package plugin
 
 import (
 	"bufio"
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,12 @@ RUN GO111MODULE=on CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} GOPROXY=$
 		-ldflags="-w -s -X k8s.io/component-base/version.gitVersion=${VERSION}" \
 		-o=aws-cloud-controller-manager \
 		cmd/aws-cloud-controller-manager/main.go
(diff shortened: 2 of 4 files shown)


cmd/ecr-credential-provider/plugin/plugin.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package plugin
```
cartermckinnon, line 17: This change seems unrelated, can you revert it if it's not necessary for this PR?

cartermckinnon:
/hold

While we explore other options. I don't think this is totally unreasonable, but I don't think it'll be used very often; and it has the downsides of increased image size and surface area for security scanners.

cmd/ecr-credential-provider/plugin/plugin.go
kon-angelo: This solves a very peculiar thing I saw. Without it, compilation (aka `go build`) would fail if not performed inside `cmd/ecr-credential-provider` directory (requiring you to `cd`) and the error was that the `NewCredentialProvider` could not be found.

It all goes away when you refactor the `plugin.go` file out of the `main` package though. You can "script" around the issue but refactoring to a proper package seemed the better way to go. We can argue about the package location of course.

hakman:
> While we explore other options. I don't think this is totally unreasonable, but I don't think it'll be used very often; and it has the downsides of increased image size and surface area for security scanners.

@cartermckinnon I suggest creating a separate image for it, in case this is something needed in the future. I don't see any value in bundling it in a single image.

cartermckinnon:
> I suggest creating a separate image for it, in case this is something needed in the future. I don't see any value in bundling it in a single image.

Agreed; I don't think we're going to go this route. A container image just isn't the right way to distribute this.

/close

PR #898, opened 2024-04-18
(cleanup): remove unused volume handling

cartermckinnon:
**What type of PR is this?**

/kind cleanup

This removes dead code associated with EBS volumes. This is old code from the legacy in-tree AWS provider, and has never been used in the AWS CCM. I sanity checked that it is not being used by the EBS or EFS CSI driver projects, and I don't see any usage searching around on grep.app.

Fixes #

```release-note
Legacy volume handling has been removed.
```


mmerkes:
/lgtm
/approve

Tx for cleaning this up! Can believe we had so much dead code in here.

PR #909, opened 2024-04-30
Ensure that addresses are added in network device index order

### LINKED ISSUE #911: DeviceIndex is not respected when processing multiple network interfaces on a node
**What happened**:

In cases where a node has multiple network interfaces, the order in which they are listed from AWS is applied directly to the interfaces list that the CCM produces. Depending on the ordering from AWS, this may or may not be accurate for the node being deployed. Even when `DeviceIndex` is specified, it is not respected.

For example, given this data from AWS:
```json
{
  "Reservations": [
    {
      "Instances": [
        {
          "NetworkInterfaces": [
            {
              "Attachment": {
                "DeviceIndex": 1,
              },
              "MacAddress": "02:0f:d9:6d:bf:eb",
              "PrivateIpAddress": "100.64.2.164",
            },
            {
              "Attachment": {
                "DeviceIndex": 0,
              },
              "MacAddress": "02:5c:60:51:b1:29",
              "PrivateIpAddress": "10.175.4.157",
            }
          ],
        }
      ],
    }
  ]
}
```

The interface with MAC `02:5c:60:51:b1:29` should be the first interface by its `DeviceIndex`, yet it is the second interface due to the ordering from AWS.

When `DeviceIndex` is specified, I expect it to be honored when ordering the network interfaces.

1. create a node with multiple network interfaces, and device indexes specified
2. ensure that the ordering from AWS is wrong from device index, or set the device indexes to be wrong based on the MAC address ordering
3. inspect the primary network interface IP on the node object created

this was solved in the in-tree cloud controller but appears to have not been ported to the out-of-tree controller, see https://github.com/kubernetes/kubernetes/pull/80747

this is the current code in the external ccm, https://github.com/openshift/cloud-provider-aws/blob/release-4.14/pkg/providers/v1/aws.go#L1515

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

/kind bug

javanthropus:
**What type of PR is this?**
/kind bug

This ensures that the addresses associated with network devices attached to the host are added to a Node resource's address list in order of device index.  For some unknown reason AWS returns the list of network devices for just some of our EC2 instances where the primary device is not first in the list.  Without this code change, the addresses of the secondary devices are listed first in the addresses for a Node, and this breaks the ability to interact with pods on the node, such as fetching logs and creating port forwards, because the apiserver always uses the first address of the Node resource to reach the kubelet but is unable to reach kubelet on any of these other addresses.

Fixes #911.

None

```release-note
Addresses associated with the Node resource will be sorted in order of the index of the network device to which they're attached.  In cases where a VPC CNI, such as aws-cni, is used and the order of devices returned for a node is unpredictable, this ensures that Kubernetes uses the right address to interact with the kubelet on the node.
```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -928,6 +928,19 @@ func extractIPv4NodeAddresses(instance *ec2.Instance) ([]v1.NodeAddress, error) 
 
 	addresses := []v1.NodeAddress{}
 
+	// sort by device index so that the first address added to the addresses list is from the first (primary) device
+	sort.Slice(instance.NetworkInterfaces, func(i, j int) bool {
+		// These nil checks should cause interfaces with non-nil attachments to sort before those with nil attachments
+		if instance.NetworkInterfaces[i].Attachment == nil {
+			return false
+		}
+		if instance.NetworkInterfaces[j].Attachment == nil {
+			return true
+		}
+
+		return aws.Int64Value(instance.NetworkInterfaces[i].Attachment.DeviceIndex) < aws.Int64Value(instance.NetworkInterfaces[j].Attachment.DeviceIndex)
+	})
+
 	// handle internal network interfaces
 	for _, networkInterface := range instance.NetworkInterfaces {
 		// skip network interfaces that are not currently in use
--- a/pkg/providers/v1/aws_fakes.go
+++ b/pkg/providers/v1/aws_fakes.go
@@ -77,6 +77,45 @@ func NewFakeAWSServices(clusterID string) *FakeAWSServices {
 	s.selfInstance = selfInstance
 	s.instances = []*ec2.Instance{selfInstance}
 
+	selfInstance.NetworkInterfaces = []*ec2.InstanceNetworkInterface{
+		{
+			Attachment: &ec2.InstanceNetworkInterfaceAttachment{
+				DeviceIndex: aws.Int64(1),
+			},
+			PrivateIpAddresses: []*ec2.InstancePrivateIpAddress{
+				{
+					Primary:          aws.Bool(true),
+					PrivateDnsName:   aws.String("ip-172-20-1-100.ec2.internal"),
+					PrivateIpAddress: aws.String("172.20.1.1"),
+				},
+				{
+					Primary:          aws.Bool(false),
+					PrivateDnsName:   aws.String("ip-172-20-1-101.ec2.internal"),
+					PrivateIpAddress: aws.String("172.20.1.2"),
+				},
+			},
+			Status: aws.String(ec2.NetworkInterfaceStatusInUse),
+		},
+		{
+			Attachment: &ec2.InstanceNetworkInterfaceAttachment{
+				DeviceIndex: aws.Int64(0),
+			},
+			PrivateIpAddresses: []*ec2.InstancePrivateIpAddress{
+				{
+					Primary:          aws.Bool(true),
+					PrivateDnsName:   aws.String("ip-172-20-0-100.ec2.internal"),
+					PrivateIpAddress: aws.String("172.20.0.100"),
+				},
+				{
+					Primary:          aws.Bool(false),
+					PrivateDnsName:   aws.String("ip-172-20-0-101.ec2.internal"),
+					PrivateIpAddress: aws.String("172.20.0.101"),
+				},
+			},
+			Status: aws.String(ec2.NetworkInterfaceStatusInUse),
+		},
+	}
+
 	var tag ec2.Tag
 	tag.Key = aws.String(TagNameKubernetesClusterLegacy)
 	tag.Value = aws.String(clusterID)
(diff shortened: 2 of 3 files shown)


pkg/providers/v1/aws.go
```diff
@@ -1543,6 +1543,11 @@ func extractIPv4NodeAddresses(instance *ec2.Instance) ([]v1.NodeAddress, error)
 
 	addresses := []v1.NodeAddress{}
 
+	// sort by device index so that the first address added to the addresses list is from the first (primary) device
+	sort.Slice(instance.NetworkInterfaces, func(i, j int) bool {
+		return *instance.NetworkInterfaces[i].Attachment.DeviceIndex < *instance.NetworkInterfaces[j].Attachment.DeviceIndex
```
cartermckinnon, line 1548: Can you add checks for a `nil` attachment and use `aws.IntValue` to dereference the index? The EC2 API should return sane values but we shouldn't crash if it doesn't

javanthropus: Should I filter out interfaces where the attachment is `nil`, or should I just sort them to the end of the list?  I feel like filtering them out would be the right solution since they should have addresses associated if the interface isn't actually attached to an instance.

cartermckinnon: The check here: https://github.com/kubernetes/cloud-provider-aws/pull/909/files/40202804f2189a7e3958261c8a1732bd8fd7686c#diff-97a31d0f193ea0b4e2c538595daf93405351054377458053a5010f6fdb97f4dcR1554-R1556

Should filter those out, so you just need to handle the potential `nil` pointer here. Start or end of the slice doesn't matter to me 😄

javanthropus: I believe the updated code should just cause them to be sorted to the end of the list.  It avoids allocating another slice, and the code that pulls the addresses out should ignore them anyway.

javanthropus:
What's the process for back porting changes?  My understanding is that we need to use the 1.27 release branch since our clusters are still running k8s 1.27.

cartermckinnon:
Ah I didn't notice you had this open against `release-1.30`. I changed the base to `master`, and after merge you can open cherrypick PR against older release branches as needed. I use this script to fire them off quickly: https://github.com/kubernetes/kubernetes/blob/master/hack/cherry_pick_pull.sh

example here: https://kops.sigs.k8s.io/contributing/proposing-a-cherry-pick/

javanthropus:
> Ah I didn't notice you had this open against `release-1.30`. I changed the base to `master` so you'll need to rebase.

Sorry about that.  It wasn't clear what the process is for changes like this.  Should I go ahead and squash my commits when I rebase?

cartermckinnon:
Sure you can go ahead and squash! Thx

cartermckinnon:
/approve

/assign @mmerkes

PTAL

pkg/providers/v1/aws.go
```diff
@@ -928,6 +928,19 @@ func extractIPv4NodeAddresses(instance *ec2.Instance) ([]v1.NodeAddress, error)
 
 	addresses := []v1.NodeAddress{}
 
+	// sort by device index so that the first address added to the addresses list is from the first (primary) device
+	sort.Slice(instance.NetworkInterfaces, func(i, j int) bool {
```
mmerkes, line 932: Can we write a unit test for this to confirm that the ordering is what we intend?

javanthropus: I could give it a try, but I didn't do so already because I couldn't find much in the way of existing tests here.  As a result, I don't have any good examples to base my work on.  Did I overlook something?

mmerkes: That's good information to have. I don't know that it necessarily needs to be a strict blocker, but I don't think it would be too crazy.

I don't see any asynchronous calls here, so it should act entirely on the inputs. You could separate this out into another method to really narrow down the testing, but shouldn't be much more scope to leave as is. [Here's a somewhat similar sort of test](https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws_test.go#L3120-L3177).

javanthropus: Thanks for the link.  I was looking in the wrong place for tests.  That's not surprising since I'm very green with golang.  I'll try to add some tests tomorrow and push them up.

javanthropus: I finally managed to understand how the fakes work here, but I don't see any tests for this functionality.  I see some tests for Fargate instances, but they don't use the NetworkInterfaces list and rely solely on the PrivateDnsName and PrivateIpAddress fields of the instance.  Shouldn't that be sufficient for non-Fargate instances as well?  In other words, why does this controller attempt to list all addresses for all network interfaces for a node rather than just the primary address and DNS name?

javanthropus: @mmerkes, do you know if we can simplify this whole thing and just use the primary address and DNS name of the nodes rather than iterate over all attached addresses?

javanthropus: A test has been added as requested.

javanthropus:
The last couple of failed test runs don't appear to be related to my change.  @mmerkes or @cartermckinnon, can one of you PTAL?

cartermckinnon:
@javanthropus sorry about that, the CI is busted. If we can't get a fix in shortly I'll override.

hakman:
@javanthropus Please rebase instead of merge. Thanks!

cartermckinnon:
Thanks for fixing this @javanthropus!

PR #917, opened 2024-05-17
[WIP] extract fargate related code into a standalone file

dims:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


hakman:
@dims You will probably need to rebase to fix `govulncheck`.

dims:
> @dims You will probably need to rebase to fix govulncheck.

thanks @hakman Done!

pkg/providers/v1/iface/types.go
```diff
@@ -0,0 +1,36 @@
+package iface
```
cartermckinnon, line 1: Should `EC2Metadata` be in here too?

pkg/providers/v1/variant/variant.go
```diff
@@ -0,0 +1,88 @@
...
+	v1 "k8s.io/api/core/v1"
+	cloudprovider "k8s.io/cloud-provider"
+
+	"github.com/aws/aws-sdk-go/aws/credentials"
+
+	"k8s.io/cloud-provider-aws/pkg/providers/v1/config"
+	"k8s.io/cloud-provider-aws/pkg/providers/v1/iface"
+)
+
+var variantsLock sync.Mutex
```
cartermckinnon, line 16: I think we need a https://pkg.go.dev/sync#RWMutex since we'll have multiple routines calling `IsVariantNode` while writes will only happen during init-s

pkg/providers/v1/variant/fargate/fargate.go
```diff
@@ -0,0 +1,170 @@
...
+	return eni.NetworkInterfaces[0], nil
+}
+
+func init() {
+	v := &fargateVariant{}
+	variant.RegisterVariant(
+		"fargate",
+		v,
+	)
+}
```
cartermckinnon, line 160: Something connected to `main()` has to `import` this package for the init to have any effect, right? (maybe I missed it)

We could do this registration in `variant.go` instead, but seems like the goal is to allow other variants to be dropped-in without patching other bits?

cartermckinnon: Ah I missed it: https://github.com/kubernetes/cloud-provider-aws/pull/917/files#diff-97a31d0f193ea0b4e2c538595daf93405351054377458053a5010f6fdb97f4dcR63

I think it makes more sense for that to be in `variants.go`, but I'd ship it

rsumukha:
Ran a prow test suite against to validate Fargate tests, it was mostly successful with an unrelated test failure. gtg from Fargate side

pkg/providers/v1/variant/fargate/fargate.go
dims: we can always iterate. thanks!

pkg/providers/v1/variant/variant.go
dims: only during `init`

pkg/providers/v1/iface/types.go
dims: we could in a later PR.

pkg/providers/v1/variant/variant.go
dims: i believe all init's are called in a specific order

PR #938, opened 2024-06-03
refactor: allow variants to implement their own k8s providerID parsing logic

sanjams2:
== Motivation ==

Allow further variant specific customization

== Details ==

This change adds the ability for variants to implement their own logic to parse a k8s providerID into an identifier specific to that variant. This is done by adding a new method 'NodeId' on each variant that takes the providerID (in pre-parsed url format) and returns a respective NodeId. That said, the behavior of the existing Fargate variant and EC2 node are left unchanged. 

NodeId is a refactor of the previous InstanceId which was inadequately named for variants other than EC2 instances (e.g. fargate). I have also taken this opportunity to add a bit more strong typing to the variant methods to make these methods a bit easier to grok and safer to implement. I have also renamed the KubernetesInstanceId to KubernetesProviderId which felt a more adequate name as well. Finally, in order to prevent circular depedencies and to maintain (what I felt) was a more logical concept of a NodeId, we have created a new submodule in the v1 package named 'awsnode' which contains the NodeId type. This allows both the base v1 module and the variant submodule to use this NodeId type withou circular dependencies.

I have opted not to change variables named `instanceId` or similar for the time being given the bloat that would cause on this CR size. 

Im pushing this change now to get feedback quickly. Can work on adding the necessary testing steps/release notes as required. Based on https://github.com/kubernetes/cloud-provider-aws/pull/917, it seems this may not be required for this change.

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"k8s.io/cloud-provider-aws/pkg/providers/v1/awsnode"
 	"net"
 	"regexp"
 	"sort"
@@ -418,7 +419,7 @@ func InstanceIDIndexFunc(obj interface{}) ([]string, error) {
 		// provider ID hasn't been populated yet
 		return []string{""}, nil
 	}
-	instanceID, err := KubernetesInstanceID(node.Spec.ProviderID).MapToAWSInstanceID()
+	instanceID, err := ParseProviderID(node.Spec.ProviderID)
 	if err != nil {
 		//logging the error as warning as Informer.AddIndexers would panic if there is an error
 		klog.Warningf("error mapping node %q's provider ID %q to instance ID: %v", node.Name, node.Spec.ProviderID, err)
@@ -832,16 +833,16 @@ func extractIPv6NodeAddresses(instance *ec2.Instance) ([]v1.NodeAddress, error) 
 // This method will not be called from the node that is requesting this ID. i.e. metadata service
 // and other local methods cannot be used here
 func (c *Cloud) NodeAddressesByProviderID(ctx context.Context, providerID string) ([]v1.NodeAddress, error) {
-	instanceID, err := KubernetesInstanceID(providerID).MapToAWSInstanceID()
+	instanceID, err := ParseProviderID(providerID)
 	if err != nil {
 		return nil, err
 	}
 
-	if v := variant.GetVariant(string(instanceID)); v != nil {
-		return v.NodeAddresses(string(instanceID), c.vpcID)
+	if v := variant.GetVariant(instanceID); v != nil {
+		return v.NodeAddresses(instanceID, c.vpcID)
 	}
 
-	instance, err := describeInstance(c.ec2, instanceID)
+	instance, err := describeInstance(c.ec2, string(instanceID))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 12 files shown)


sanjams2:
@cartermckinnon any way you could give this a quick glance and see if there is anything glaring? thanks in advance!

pkg/providers/v1/awsnode/identifier.go
```diff
@@ -0,0 +1,11 @@
+package awsnode
+
+import "github.com/aws/aws-sdk-go/aws"
+
+// NodeID is the ID of a node used to uniquely identify that node within an AWS service
+type NodeID string
```
cartermckinnon, line 6: I think this could be more useful as an interface, we have too many string alias types in here already 😛 

Something like:
```
type AWSProviderID interface {
    ComputeID() string
}
```

Then we can implement an `EC2ProviderID`, `FargateProviderID`, etc.

The func-s in the `Variant` interface can accept an `AWSProviderID` and cast to the relevant type as necessary, `providerID.(EC2ProviderID)`

pkg/providers/v1/instances.go
```diff
@@ -36,27 +37,16 @@ import (
...
-
-// KubernetesInstanceID represents the id for an instance in the kubernetes API;
+// KubernetesProviderID represents the id for an instance in the kubernetes API;
 // the following form
 //   - aws:///<zone>/<awsInstanceId>
 //   - aws:////<awsInstanceId>
 //   - aws:///<zone>/fargate-<eni-ip-address>
 //   - <awsInstanceId>
-type KubernetesInstanceID string
+type KubernetesProviderID string
```
cartermckinnon, line 46: This type doesn't seem very useful to me, can we get rid of it in favor of something like:
```
func ParseProviderID(providerID string) (AWSProviderID, error)
```
?

cartermckinnon (commented):
I like where this is headed, and I think we can do a bit more compaction of the various types involved

pkg/providers/v1/instances.go
sanjams2: Agree and makes sense. Will fix

pkg/providers/v1/awsnode/identifier.go
sanjams2: hmm I actually am not quite following the need for an interface here. While the NodeID might have different formats, I dont see why there would be different implementations required for those different formats. Along that same line, while it sounds potentially useful to have strongly typed NodeID's, there is currently not a need for it — the casting you mention wouldnt actually be necessary anywhere. That to me signals that an interface may be overkill, but perhaps Im not thinking about a future use-case where having an interface would be ideal. 

So as it stands now, I actually think the ergonomics of the string alias type is fits the needs quite well. Let me know your thoughts.

sanjams2: also, much less of a concern, but moving to an interface would substantially increase the size of the changes in this commit.

sanjams2:
The test errors seem unrelated to my change. I also cannot find an existing issue noting a similar symptom to these failures.

pkg/providers/v1/aws.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"k8s.io/cloud-provider-aws/pkg/providers/v1/awsnode"
```
kmala, line 23: nit: can we group the import correctly ?

pkg/providers/v1/instances.go
```diff
@@ -36,40 +37,29 @@ import (
...
 //   - aws:////<awsInstanceId>
 //   - aws:///<zone>/fargate-<eni-ip-address>
 //   - <awsInstanceId>
-type KubernetesInstanceID string
-
-// MapToAWSInstanceID extracts the InstanceID from the KubernetesInstanceID
-func (name KubernetesInstanceID) MapToAWSInstanceID() (InstanceID, error) {
-	s := string(name)
-
+func ParseProviderID(providerID string) (awsnode.NodeID, error) {
```
kmala, line 46: i observed that we are making call to get variant followed by this method in most places. Can we return the variant also such that the caller can get both at the same time ?

PR #997, opened 2024-07-17
Handle error while registering/deregistering target during load balancer update calls

jaydeokar:
**What type of PR is this?**
/kind bug

If the RegisterInstancesWithLoadBalancer fails for any reason during an LoadBalancer Update call, we do not throw the error and return nil. So we will never know if the target group registration passes or fails

For testing - 
I tried to mock the ELB API as I wasn't able to hit the throttling every time. 

I introduced a 50% chance of ELB API failing to register the target using Mocked return message 
```
		val := rand.Int()
		if val%2 != 0 {
			klog.V(1).Infof("Not using Mocked call %s", loadBalancerName)
			_, err = c.elb.RegisterInstancesWithLoadBalancer(registerRequest)
		} else {
			klog.V(1).Infof("Using Mocked call %s", loadBalancerName)
			err = fmt.Errorf("mocked throttled call")
		}

		if err != nil {
			return err
		}
		klog.V(1).Infof("Instances added to load-balancer %s", loadBalancerName)
```

The events on the service 

```
Warning  UnAvailableLoadBalancer   6m31s                 service-controller  There are no available nodes for LoadBalancer
  Warning  UpdateLoadBalancerFailed  2m34s                 service-controller  Error updating load balancer with new hosts [i-035ae0c5cd612d191 i-05560870005eef14b i-0661ad74a5ae13914 i-0a48dd4251b3e24d3] [node names limited, total number of nodes: 4], error: mocked throttled call
  Normal   UpdatedLoadBalancer       2m6s (x6 over 6m58s)  service-controller  Updated load balancer with new hosts
  Warning  UpdateLoadBalancerFailed  53s                   service-controller  Error updating load balancer with new hosts [i-00153b6992defc5d0 i-035ae0c5cd612d191 i-05560870005eef14b i-0661ad74a5ae13914 i-0a48dd4251b3e24d3 i-0cc2ee458bb7db682] [node names limited, total number of nodes: 6], error: mocked throttled call
  Warning  SyncLoadBalancerFailed    28s (x6 over 2m30s)   service-controller  Error syncing load balancer: failed to ensure load balancer: mocked throttled call
  Normal   EnsuringLoadBalancer      8s (x9 over 13m)      service-controller  Ensuring load balancer
  Normal   EnsuredLoadBalancer       8s (x3 over 13m)      service-controller  Ensured load balancer
``` 

This shows that the controller retries and reconciles again in case of failures. 

The logs in the controller shows the new warn message that was added 

```
W0729 21:20:19.689788       1 aws.go:3016] Error registering/deregistering instances with the load balancer: "mocked throttled call"
I0729 21:20:19.747638       1 event.go:389] "Event occurred" object="ui/ui-nlb-11" fieldPath="" kind="Service" apiVersion="v1" type="Warning" reason="UpdateLoadBalancerFailed" message="Error updating load balancer with new hosts [i-00153b6992defc5d0 i-035ae0c5cd612d191 i-05560870005eef14b i-0661ad74a5ae13914 i-0a48dd4251b3e24d3 i-0cc2ee458bb7db682] [node names limited, total number of nodes: 6], error: mocked throttled call"
```

N/A

Fixes #

N.A

N/A

```release-note
Fixes an bug with UpdateLoadBalancer call where any failure in RegisterInstances API to ELB is always returned as success
```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -3013,7 +3013,8 @@ func (c *Cloud) UpdateLoadBalancer(ctx context.Context, clusterName string, serv
 
 	err = c.ensureLoadBalancerInstances(aws.StringValue(lb.LoadBalancerName), lb.Instances, instances)
 	if err != nil {
-		return nil
+		klog.Warningf("Error registering/deregistering instances with the load balancer: %q", err)
+		return err
 	}
 
 	err = c.updateInstanceSecurityGroupsForLoadBalancer(lb, instances, service.Annotations)
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,1 +1,1 @@
-defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.22.4-bookworm.0
+defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.22.5-bookworm.0
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 ##                               BUILD ARGS                                   ##
 ################################################################################
 # This build arg allows the specification of a custom Golang image.
-ARG GOLANG_IMAGE=golang:1.22.4
+ARG GOLANG_IMAGE=golang:1.22.5
 
 # The distroless image on which the CPI manager image is built.
 #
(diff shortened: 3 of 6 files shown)


pkg/providers/v1/aws.go
```diff
@@ -3013,7 +3013,8 @@ func (c *Cloud) UpdateLoadBalancer(ctx context.Context, clusterName string, serv
 
 	err = c.ensureLoadBalancerInstances(aws.StringValue(lb.LoadBalancerName), lb.Instances, instances)
 	if err != nil {
-		return nil
+		klog.Warningf("Error registering/deregistring instances with the load balancer: %q", err)
+		return err
```
kmala, line 3017: this would cause an event and users can now know about the failures correct? If so , can we update the release notes about the same on change is behavior.

kmala: That event won't happen before correct as we are not returning error. After this change the event will be shown during errors which would help in debugging correct?

jaydeokar: Yes, but that shouldn't be a event which introduces a "behavior" change. It should have been the expected behavior from the start which was not handled. I can call it out as a. bug in the release notes

kmala: yes, that's correct. It's a bug which we are fixing and it would be nice to mention it in the release notes about the fixing it.

jaydeokar: Updated

PR #1024, opened 2024-09-17
Assume instance exists within eventual-consistency grace period

cartermckinnon:
**What type of PR is this?**

/kind bug

For a short period of time after an EC2 instance is launched, the `ec2:DescribeInstances` API may not return details for the instance. When this happens, the node lifecycle controller may delete the `Node` erroneously, believing the corresponding EC2 instance does not exist.

This PR adds a configurable "grace period" after `Node` creation, during which the EC2 instance is assumed to exist.

Fixes #

This is possible in our `InstancesV2` implementation, because the controller will pass us the `Node`: https://github.com/kubernetes/cloud-provider/blob/912e64449ce4cb3645436a768d4a8d5c834652ed/controllers/nodelifecycle/node_lifecycle_controller.go#L253

⚠️ **it will not be possible to cherry-pick this back to older release branches that do not implement `InstancesV2`.**

```release-note
NodeEventualConsistencyGracePeriod is a new configuration option to account for propogation delays in EC2 APIs. InstanceExists will now return an error for newly-created nodes until the grace period has elapsed.
```

### DIFF
--- a/pkg/controllers/tagging/metrics.go
+++ b/pkg/controllers/tagging/metrics.go
@@ -14,23 +14,15 @@ limitations under the License.
 package tagging
 
 import (
+	"sync"
+
 	"k8s.io/component-base/metrics"
 	"k8s.io/component-base/metrics/legacyregistry"
-	"sync"
 )
 
 var register sync.Once
 
 var (
-	workItemDuration = metrics.NewHistogramVec(
-		&metrics.HistogramOpts{
-			Name:           "cloudprovider_aws_tagging_controller_work_item_duration_seconds",
-			Help:           "workitem latency of workitem being in the queue and time it takes to process",
-			StabilityLevel: metrics.ALPHA,
-			Buckets:        metrics.ExponentialBuckets(0.5, 1.5, 20),
-		},
-		[]string{"latency_type"})
-
 	workItemError = metrics.NewCounterVec(
 		&metrics.CounterOpts{
 			Name:           "cloudprovider_aws_tagging_controller_work_item_errors_total",
@@ -43,13 +35,8 @@ var (
 // registerMetrics registers tagging-controller metrics.
 func registerMetrics() {
 	register.Do(func() {
-		legacyregistry.MustRegister(workItemDuration)
 		legacyregistry.MustRegister(workItemError)
 	})
-}
-
-func recordWorkItemLatencyMetrics(latencyType string, timeTaken float64) {
-	workItemDuration.With(metrics.Labels{"latency_type": latencyType}).Observe(timeTaken)
 }
 
 func recordWorkItemErrorMetrics(errorType string, instanceID string) {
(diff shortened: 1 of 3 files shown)


pkg/providers/v1/instances_v2.go
```diff
@@ -44,6 +49,11 @@ func (c *Cloud) getProviderID(ctx context.Context, node *v1.Node) (string, error
 // InstanceExists returns true if the instance for the given node exists according to the cloud provider.
 // Use the node.name or node.spec.providerID field to find the node in the cloud provider.
 func (c *Cloud) InstanceExists(ctx context.Context, node *v1.Node) (bool, error) {
+	if time.Since(node.CreationTimestamp.Time) < instanceExistsGracePeriod {
```
kmala, line 52: should we check only if the node is not getting deleted by checking if the deletion timestamp exists?

pkg/providers/v1/instances_v2.go
```diff
@@ -22,12 +22,17 @@ package aws
...
 	"context"
+	"time"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	cloudprovider "k8s.io/cloud-provider"
 )
 
+const (
+	instanceExistsGracePeriod = 2 * time.Minute
```
kmala, line 33: how did we come with this time period?

cartermckinnon:
Closing this in favor of https://github.com/kubernetes/kubernetes/pull/127424

cartermckinnon:
Rebooting this.

pkg/providers/v1/instances_v2.go
```diff
@@ -45,6 +47,12 @@ func (c *Cloud) getProviderID(ctx context.Context, node *v1.Node) (string, error
 // InstanceExists returns true if the instance for the given node exists according to the cloud provider.
 // Use the node.name or node.spec.providerID field to find the node in the cloud provider.
 func (c *Cloud) InstanceExists(ctx context.Context, node *v1.Node) (bool, error) {
+	if time.Since(node.CreationTimestamp.Time) < c.nodeEventualConsistencyGracePeriod {
+		// recently-launched EC2 instances may not appear in `ec2:DescribeInstances`
+		// we return an error to cause the cloud-node-lifecycle-controller to ignore this node
```
mmerkes, line 52: Is there any downside to just forcing it to fail? Could increase the error logs. Can we just move this check down into code so that we only return an error if it's within the grace period AND it doesn't have the eventually consistent fields set? That would drastically reduce the noise.

cartermckinnon: Yep had the same thought. We have to handle `InvalidInstanceId.NotFound` differently in the `Instances` v1/v2 impls, since we only have the context to apply this grace period in v2

PR #1025, opened 2024-09-19
Add network topology information to nodes

### LINKED ISSUE #998: Add network topology information to nodes

I'd like information about the network topology of an instance to be added to the node. This information can be dynamically retrieved via the [EC2 DescribeInstanceTopology API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceTopology.html) and be applied onto the node object similar to `topology.k8s.aws/zone-id`, whose mechanism was introduced [in this PR](https://github.com/kubernetes/kubernetes/pull/123223). 

**Why is this needed**:
Information about an instance's topology is greatly beneficial when scheduling AI/ML workloads, where pods need to be in close proximity of one another. 

Mapping [the EC2 InstanceTopology to K8s labels](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/how-ec2-instance-topology-works.html) might not be straightforward given it returns a list of network nodes, each sorted by the number of layers. We could probably choose something like - 
```
topology.k8s.aws/network-node-1: nn1
topology.k8s.aws/network-node-2: nn2
topology.k8s.aws/network-node-3: nn3
```

/kind feature

wwvela:
**What type of PR is this?**
/kind feature

This PR is to Add network topology information to nodes for Issue #998.

Fixes #998 

NONE

### DIFF
--- a/hack/e2e/run.sh
+++ b/hack/e2e/run.sh
@@ -128,7 +128,7 @@ if [[ "${UP}" = "yes" ]]; then
       --run-id="${test_run_id}" \
       --cloud-provider=aws \
       --cluster-name="${CLUSTER_NAME}" \
-      --create-args="--dns=none --zones=${ZONES} --node-size=p4d.24xlarge --master-size=m5.large --override=cluster.spec.cloudControllerManager.cloudProvider=aws --override=cluster.spec.cloudControllerManager.image=${IMAGE_NAME}:${IMAGE_TAG}" \
+      --create-args="--dns=none --zones=${ZONES} --node-size=m5.large --master-size=m5.large --override=cluster.spec.cloudControllerManager.cloudProvider=aws --override=cluster.spec.cloudControllerManager.image=${IMAGE_NAME}:${IMAGE_TAG}" \
       --admin-access="0.0.0.0/0" \
       --kubernetes-version="${KUBERNETES_VERSION}" \
       --kops-version-marker=https://storage.googleapis.com/k8s-staging-kops/kops/releases/markers/master/latest-ci-updown-green.txt \
--- a/hack/e2e/run.sh
+++ b/hack/e2e/run.sh
@@ -128,7 +128,7 @@ if [[ "${UP}" = "yes" ]]; then
       --run-id="${test_run_id}" \
       --cloud-provider=aws \
       --cluster-name="${CLUSTER_NAME}" \
-      --create-args="--dns=none --zones=${ZONES} --node-size=m5.large --master-size=m5.large --override=cluster.spec.cloudControllerManager.cloudProvider=aws --override=cluster.spec.cloudControllerManager.image=${IMAGE_NAME}:${IMAGE_TAG}" \
+      --create-args="--dns=none --zones=${ZONES} --node-size=p4d.24xlarge --master-size=m5.large --override=cluster.spec.cloudControllerManager.cloudProvider=aws --override=cluster.spec.cloudControllerManager.image=${IMAGE_NAME}:${IMAGE_TAG}" \
       --admin-access="0.0.0.0/0" \
       --kubernetes-version="${KUBERNETES_VERSION}" \
       --kops-version-marker=https://storage.googleapis.com/k8s-staging-kops/kops/releases/markers/master/latest-ci-updown-green.txt \
--- a/pkg/providers/v1/aws_ec2.go
+++ b/pkg/providers/v1/aws_ec2.go
@@ -30,6 +30,16 @@ type awsSdkEC2 struct {
 	ec2 ec2iface.EC2API
 }
 
+func (s *awsSdkEC2) DescribeInstanceTopology(request *ec2.DescribeInstanceTopologyInput) (*ec2.InstanceTopology, error) {
+	resp, err := s.ec2.DescribeInstanceTopology(request)
+	if err != nil {
+		return nil, fmt.Errorf("error describe AWS Instance Topology: %q", err)
+	} else if len(resp.Instances) == 0 {
+		return nil, nil
+	}
+	return resp.Instances[0], err
+}
+
 // Implementation of EC2.Instances
 func (s *awsSdkEC2) DescribeInstances(request *ec2.DescribeInstancesInput) ([]*ec2.Instance, error) {
 	// Instances are paged
--- a/pkg/providers/v1/aws_ec2.go
+++ b/pkg/providers/v1/aws_ec2.go
@@ -35,7 +35,7 @@ func (s *awsSdkEC2) DescribeInstanceTopology(request *ec2.DescribeInstanceTopolo
 	if err != nil {
 		return nil, fmt.Errorf("error describe AWS Instance Topology: %q", err)
 	} else if len(resp.Instances) == 0 {
-		return nil, nil
+		return []*ec2.InstanceTopology{}, nil
 	}
 	return resp.Instances, err
 }
--- a/pkg/providers/v1/aws_ec2.go
+++ b/pkg/providers/v1/aws_ec2.go
@@ -30,14 +30,14 @@ type awsSdkEC2 struct {
 	ec2 ec2iface.EC2API
 }
 
-func (s *awsSdkEC2) DescribeInstanceTopology(request *ec2.DescribeInstanceTopologyInput) (*ec2.InstanceTopology, error) {
+func (s *awsSdkEC2) DescribeInstanceTopology(request *ec2.DescribeInstanceTopologyInput) ([]*ec2.InstanceTopology, error) {
 	resp, err := s.ec2.DescribeInstanceTopology(request)
 	if err != nil {
 		return nil, fmt.Errorf("error describe AWS Instance Topology: %q", err)
 	} else if len(resp.Instances) == 0 {
 		return nil, nil
 	}
-	return resp.Instances[0], err
+	return resp.Instances, err
 }
 
 // Implementation of EC2.Instances
--- a/pkg/providers/v1/iface/types.go
+++ b/pkg/providers/v1/iface/types.go
@@ -8,6 +8,7 @@ import "github.com/aws/aws-sdk-go/service/ec2"
 type EC2 interface {
 	// Query EC2 for instances matching the filter
 	DescribeInstances(request *ec2.DescribeInstancesInput) ([]*ec2.Instance, error)
+	DescribeInstanceTopology(request *ec2.DescribeInstanceTopologyInput) (*ec2.InstanceTopology, error)
 
 	DescribeSecurityGroups(request *ec2.DescribeSecurityGroupsInput) ([]*ec2.SecurityGroup, error)
 
--- a/pkg/providers/v1/iface/types.go
+++ b/pkg/providers/v1/iface/types.go
@@ -8,7 +8,7 @@ import "github.com/aws/aws-sdk-go/service/ec2"
 type EC2 interface {
 	// Query EC2 for instances matching the filter
 	DescribeInstances(request *ec2.DescribeInstancesInput) ([]*ec2.Instance, error)
-	DescribeInstanceTopology(request *ec2.DescribeInstanceTopologyInput) (*ec2.InstanceTopology, error)
+	DescribeInstanceTopology(request *ec2.DescribeInstanceTopologyInput) ([]*ec2.InstanceTopology, error)
 
 	DescribeSecurityGroups(request *ec2.DescribeSecurityGroupsInput) ([]*ec2.SecurityGroup, error)
 
--- a/pkg/providers/v1/instances_v2.go
+++ b/pkg/providers/v1/instances_v2.go
@@ -22,11 +22,9 @@ package aws
 
 import (
 	"context"
-	"github.com/aws/aws-sdk-go/service/ec2"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	cloudprovider "k8s.io/cloud-provider"
-	"k8s.io/utils/strings/slices"
 	"strconv"
 )
 
@@ -76,26 +74,18 @@ func (c *Cloud) getAdditionalLabels(zoneName string, instanceID string, instance
 
 	additionalLabels[LabelZoneID] = zoneID
 
-	if c.isTopologySupported(instanceType, region) {
-		topologyRequest := &ec2.DescribeInstanceTopologyInput{InstanceIds: []*string{&instanceID}}
-		topology, err := c.ec2.DescribeInstanceTopology(topologyRequest)
-		if err != nil || len(topology) == 0 {
-			return nil, err
-		} else {
-			for index, networkNode := range topology[0].NetworkNodes {
-				layer := index + 1
-				label := LabelNetworkNode + strconv.Itoa(layer)
-				additionalLabels[label] = *networkNode
-			}
+	nodeTopology, err := c.topologyCache.getNodeTopology(instanceType, region, instanceID)
+	if err != nil {
+		return nil, err
+	} else if nodeTopology != nil {
+		for index, networkNode := range nodeTopology.NetworkNodes {
+			layer := index + 1
+			label := LabelNetworkNode + strconv.Itoa(layer)
+			additionalLabels[label] = *networkNode
 		}
 	}
 
 	return additionalLabels, nil
-}
-
-// TopologySupportedInstancePrefixes is help to filter instance types supported by topology API
-func (c *Cloud) isTopologySupported(instanceType string, region string) bool {
-	return slices.Contains(c.cfg.Global.TopologySupportedInstanceTypes, instanceType) && slices.Contains(c.cfg.Global.TopologySupportedRegions, region)
 }
 
 // InstanceMetadata returns the instance's metadata. The values returned in InstanceMetadata are
--- a/pkg/providers/v1/instances_v2.go
+++ b/pkg/providers/v1/instances_v2.go
@@ -23,14 +23,11 @@ package aws
 import (
 	"context"
 	"github.com/aws/aws-sdk-go/service/ec2"
-	"k8s.io/klog/v2"
-	"k8s.io/utils/strings/slices"
-	"strconv"
-	"strings"
-
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	cloudprovider "k8s.io/cloud-provider"
+	"k8s.io/utils/strings/slices"
+	"strconv"
 )
 
 func (c *Cloud) getProviderID(ctx context.Context, node *v1.Node) (string, error) {
@@ -79,30 +76,26 @@ func (c *Cloud) getAdditionalLabels(zoneName string, instanceID string, instance
 
 	additionalLabels[LabelZoneID] = zoneID
 
-	if c.topologySupportedInstance(instanceType) && slices.Contains(c.cfg.Global.TopologySupportedRegions, region) {
+	if c.isTopologySupported(instanceType, region) {
 		topologyRequest := &ec2.DescribeInstanceTopologyInput{InstanceIds: []*string{&instanceID}}
 		topology, err := c.ec2.DescribeInstanceTopology(topologyRequest)
-		if err != nil || topology == nil {
-			klog.Infof("topology api not supported for instance type: %s or region: %s", instanceType, region)
-			return additionalLabels, nil
-		}
-		for index, networkNode := range topology.NetworkNodes {
-			label := LabelNetworkNode + strconv.Itoa(index)
-			additionalLabels[label] = *networkNode
+		if err != nil || len(topology) == 0 {
+			return nil, err
+		} else {
+			for index, networkNode := range topology[0].NetworkNodes {
+				layer := index + 1
+				label := LabelNetworkNode + strconv.Itoa(layer)
+				additionalLabels[label] = *networkNode
+			}
 		}
 	}
-	
+
 	return additionalLabels, nil
 }
 
 // TopologySupportedInstancePrefixes is help to filter instance types supported by topology API
-func (c *Cloud) topologySupportedInstance(instanceType string) bool {
-	for _, prefix := range c.cfg.Global.TopologySupportedInstancePrefixes {
-		if strings.HasPrefix(instanceType, prefix) {
-			return true
-		}
-	}
-	return false
+func (c *Cloud) isTopologySupported(instanceType string, region string) bool {
+	return slices.Contains(c.cfg.Global.TopologySupportedInstanceTypes, instanceType) && slices.Contains(c.cfg.Global.TopologySupportedRegions, region)
 }
 
 // InstanceMetadata returns the instance's metadata. The values returned in InstanceMetadata are
--- a/pkg/providers/v1/instances_v2.go
+++ b/pkg/providers/v1/instances_v2.go
@@ -22,6 +22,11 @@ package aws
 
 import (
 	"context"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/strings/slices"
+	"strconv"
+	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -63,7 +68,7 @@ func (c *Cloud) InstanceShutdown(ctx context.Context, node *v1.Node) (bool, erro
 	return c.InstanceShutdownByProviderID(ctx, providerID)
 }
 
-func (c *Cloud) getAdditionalLabels(zoneName string) (map[string]string, error) {
+func (c *Cloud) getAdditionalLabels(zoneName string, instanceID string, instanceType string, region string) (map[string]string, error) {
 	additionalLabels := map[string]string{}
 
 	// Add the zone ID to the additional labels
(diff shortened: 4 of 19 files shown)


pkg/providers/v1/instances_v2.go
```diff
@@ -74,9 +79,32 @@ func (c *Cloud) getAdditionalLabels(zoneName string) (map[string]string, error)
...
+		topologyRequest := &ec2.DescribeInstanceTopologyInput{InstanceIds: []*string{&instanceID}}
+		topology, err := c.ec2.DescribeInstanceTopology(topologyRequest)
+		if err != nil || topology == nil {
+			klog.Infof("topology api not supported for instance type: %s or region: %s", instanceType, region)
+			return additionalLabels, nil
+		}
+		for index, networkNode := range topology.NetworkNodes {
+			label := LabelNetworkNode + strconv.Itoa(index)
+			additionalLabels[label] = *networkNode
+		}
```
cartermckinnon, line 92: I'm not clear how you would use this. We want folks to be able to schedule a pod within a "networkNode", but if the label key contains a node-specific index like this, what would the pod's nodeSelector look like?

pkg/providers/v1/aws_ec2.go
```diff
@@ -30,6 +30,16 @@ type awsSdkEC2 struct {
...
 }
 
+func (s *awsSdkEC2) DescribeInstanceTopology(request *ec2.DescribeInstanceTopologyInput) (*ec2.InstanceTopology, error) {
+	resp, err := s.ec2.DescribeInstanceTopology(request)
+	if err != nil {
+		return nil, fmt.Errorf("error describe AWS Instance Topology: %q", err)
+	} else if len(resp.Instances) == 0 {
+		return nil, nil
+	}
+	return resp.Instances[0], err
```
ndbaker1, line 40: this wrapper's logic seems strange, shouldn't we return the list of `[]*ec2.InstanceTopology` instead of just the first item?

pkg/providers/v1/well_known_labels.go
```diff
@@ -20,4 +20,5 @@ const (
 	// LabelZoneID is a topology label that can be applied to any resource
 	// but will be initially applied to nodes.
 	LabelZoneID = "topology.k8s.aws/zone-id"
+	LabelNetworkNode = "topology.k8s.aws/network-node-"
```
ndbaker1, line 23:
```suggestion
	LabelNetworkNode = "topology.k8s.aws/network-node-layer-"
```

IIUC, the info you're embedding into the tag is about the layer of the network node in the node set

pkg/providers/v1/instances_v2.go
ndbaker1: after reading through the [instance topology docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/how-ec2-instance-topology-works.html) i think i have an idea how this works?

Assuming the goal is to run workloads that share the same topology, you'd need to know the network nodes that your capacity sits on first, then use that in a `nodeSelector` for the corresponding layer index. AFAIK there's no way to apply a `nodeSelector` based on some heuristic like "schedule pods to nodes where the label `topology.k8s.aws/network-node-layer-1` matches", so this logic would have to be handled by a high level system. the only guarentee is that there are always at least 3 layers.

However, this scheme breaks down if network nodes can be part of different layers, which i couldn't immediately grok from the docs but i assumed wasn't possible.

cartermckinnon: If we inverted this, so that the label key contains the networkNode and the value is the layer number, that would allow you to use the Exists operator in a nodeAffinity to place the pod within a networkNode, without needing the layer to also match: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/node-selector-requirement/#NodeSelectorRequirement

suket22: > yep, if trees with different roots can overlap (share network nodes).

We can assume that's not the case for now. 

> A primary motivation for this feature was observability, being able to correlate issues with something more granular than AZ

Do you mean the feature to apply arbitrary labels from within the cloud provider? I think observability at the zone-id level is where we started, but I'd like to think we can help with scheduling here too. The next thing I'm thinking is a label that describes whether an instance has a GPU or not, so we don't need to schedule GPU pods (Device Plugin, EFA etc) on nodes that don't need it etc.

> Assuming the goal is to run workloads that share the same topology, you'd need to know the network nodes that your capacity sits on first, then use that in a nodeSelector for the corresponding layer index.

Right that's what I was expecting you'd need to do. You don't know the topology until you launch the instance first, so it forces this ordering of sorts where you launch the instance first, you realize what your network nodes are, and you should ideally see a common network node across your instances. I think you could maybe define weighted node affinities across the network node layers so you pack it as close as possible?

pkg/providers/v1/instances_v2.go
```diff
@@ -74,9 +79,32 @@ func (c *Cloud) getAdditionalLabels(zoneName string) (map[string]string, error)
...
+	}
+	
 	return additionalLabels, nil
 }
 
+// TopologySupportedInstancePrefixes is help to filter instance types supported by topology API
+func (c *Cloud) topologySupportedInstance(instanceType string) bool {
+	for _, prefix := range c.cfg.Global.TopologySupportedInstancePrefixes {
+		if strings.HasPrefix(instanceType, prefix) {
```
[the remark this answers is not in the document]
suket22: Do we want to be updating the cloud-provider each time there's a new AWS instance type, though?

pkg/providers/v1/instances_v2.go
```diff
@@ -74,9 +79,32 @@ func (c *Cloud) getAdditionalLabels(zoneName string) (map[string]string, error)
 
 	additionalLabels[LabelZoneID] = zoneID
 
+	if c.topologySupportedInstance(instanceType) && slices.Contains(c.cfg.Global.TopologySupportedRegions, region) {
+		topologyRequest := &ec2.DescribeInstanceTopologyInput{InstanceIds: []*string{&instanceID}}
+		topology, err := c.ec2.DescribeInstanceTopology(topologyRequest)
+		if err != nil || topology == nil {
```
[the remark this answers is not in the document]
wwvela: I revised the topologySupportedInstance to check the supported instance types instead of the instance prefix, so the DescribeInstanceTopology api should supported here. I will throw error directly

pkg/providers/v1/aws_ec2.go
wwvela: DescribeInstanceTopology can accept a list of instance IDs, if we pass only 1 instance ID, it will return 1 according topology info something like below.
```
"Instances": [
        {
            "InstanceId": "i-xxx",
            "InstanceType": "p5e.48xlarge",
            "NetworkNodes": [
                "nn-xxxxx",
                "nn-xxxxx",
                "nn-xxxxx"
            ],
            "AvailabilityZone": "us-west-2c",
            "ZoneId": "usw2-az3"
        }
    ]
```
But yeah, it seems more reasonable to return all instances in api and get the first instance in where trigger it and get result. Will revised in next

pkg/providers/v1/instances_v2.go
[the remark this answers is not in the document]
wwvela: I discussed this with Carter before. We don't want to trigger this api for all nodes, because some cx might have hundreds or thousands of nodes in their cluster, if we call this api for all nodes, it very likely trigger throttle issue. 

And DescribeInstanceTopology only supported for very limited instances types which lots of cx might not use. check the supported instance type and regions here. https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceTopology.html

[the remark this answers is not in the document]
wwvela: > Do we want to be updating the cloud-provider each time there's a new AWS instance type, though?

We don't need to update cloud-provider, TopologySupportedInstanceTypes and TopologySupportedRegions is added in cloud-config where will accept the input in KCP CCM manifest. We only need to update the manifest with the cloud-config input.

pkg/providers/v1/aws_ec2.go
ndbaker1: agreed, because we still allow the full `ec2.DescribeInstanceTopologyInput` payload which could specify more than one instance id

pkg/providers/v1/well_known_labels.go
wwvela: revised in next version

pkg/providers/v1/aws_ec2.go
wwvela: revised

pkg/providers/v1/instances_v2.go
```diff
@@ -74,9 +79,32 @@ func (c *Cloud) getAdditionalLabels(zoneName string) (map[string]string, error)
...
+			label := LabelNetworkNode + strconv.Itoa(index)
+			additionalLabels[label] = *networkNode
+		}
+	}
+	
 	return additionalLabels, nil
 }
 
+// TopologySupportedInstancePrefixes is help to filter instance types supported by topology API
+func (c *Cloud) topologySupportedInstance(instanceType string) bool {
```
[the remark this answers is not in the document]
wwvela: revised

pkg/providers/v1/instances_v2.go
```diff
@@ -74,9 +79,32 @@ func (c *Cloud) getAdditionalLabels(zoneName string) (map[string]string, error)
 
 	additionalLabels[LabelZoneID] = zoneID
 
+	if c.topologySupportedInstance(instanceType) && slices.Contains(c.cfg.Global.TopologySupportedRegions, region) {
+		topologyRequest := &ec2.DescribeInstanceTopologyInput{InstanceIds: []*string{&instanceID}}
+		topology, err := c.ec2.DescribeInstanceTopology(topologyRequest)
+		if err != nil || topology == nil {
+			klog.Infof("topology api not supported for instance type: %s or region: %s", instanceType, region)
+			return additionalLabels, nil
```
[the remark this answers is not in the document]
wwvela: As we will use the supportedInstanceType instead of instance prefix now, the error throw our there should not because not supported issue. I throw the error directly here in new version

pkg/providers/v1/aws_ec2.go
```diff
@@ -30,6 +30,16 @@ type awsSdkEC2 struct {
 	ec2 ec2iface.EC2API
 }
 
+func (s *awsSdkEC2) DescribeInstanceTopology(request *ec2.DescribeInstanceTopologyInput) ([]*ec2.InstanceTopology, error) {
+	resp, err := s.ec2.DescribeInstanceTopology(request)
```
ndbaker1, line 34: just realized this api is paginated, should we use logic similar to the `DescribeInstances` wrapper below?

pkg/providers/v1/aws_ec2.go
```diff
@@ -30,6 +30,16 @@ type awsSdkEC2 struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("error describe AWS Instance Topology: %q", err)
+	} else if len(resp.Instances) == 0 {
+		return nil, nil
```
ndbaker1, line 38: is this branch necessary? wouldn't it be better to have an empty array as a default with no error?

pkg/providers/v1/topology.go
```diff
@@ -0,0 +1,70 @@
...
+			t.unsupportedInstance = append(t.unsupportedInstance, instanceType)
+		}
+		return topology[0], nil
+	}
+	return nil, nil
+}
+
+func (t *topologyCache) mightSupportTopology(instanceType string, region string) bool {
+	t.mutex.RLock()
+	defer t.mutex.RUnlock()
```
ndbaker1, line 57: this lock might need to belong in the caller `getNodeTopology`? since that's where the possible writes occur to the data

pkg/providers/v1/topology.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"k8s.io/klog/v2"
+)
+
+type topologyCache struct {
+	cloud               *Cloud
+	mutex               sync.RWMutex
+	unsupportedInstance []string
+	unsupportedRegion   []string
+}
```
ndbaker1, line 33: just future proofing here, is there a minimal cache with TTL implementation we could use?

pkg/providers/v1/topology.go
```diff
@@ -0,0 +1,70 @@
...
+				return nil, nil
+			}
+			return nil, err
+		} else if len(topology) == 0 {
+			// instanceType is not support topology info and the result is empty
+			t.unsupportedInstance = append(t.unsupportedInstance, instanceType)
+		}
+		return topology[0], nil
+	}
+	return nil, nil
```
ndbaker1, line 52: maybe a static error to throw here is better? it seems odd to return no error with `nil`

then we could check something like `errors.Is(err, ErrTopologyUnsupported)`

pkg/providers/v1/aws_ec2.go
wwvela: The result of the DescribeInstanceTopology is actually only 1 element in the array because we will not do group describe for multiple instances. And it will only return 1 topology info for 1 instance something like 
```
{
    "Instances": [
        {
            "InstanceId": "i-xxx",
            "InstanceType": "p5e.48xlarge",
            "NetworkNodes": [
                "nn-xxx",
                "nn-xxx",
                "nn-xxx"
            ],
            "AvailabilityZone": "us-west-2c",
            "ZoneId": "usw2-az3"
        }
    ]
}
```

kmala: did we test this with large instances because even though the result can be only 1 instance because of filters, not all filters are applied at the server side for the ec2 and for filter which are applied later, we have to go over multiple pages?

pkg/providers/v1/topology.go
wwvela: Why we need to throw error if the instance not supported by the api? as we will just skip adding the topology label for it

pkg/providers/v1/topology.go
wwvela: I think it is not neccessary for the topology catch to have TTL implementation, as the already supported instance type and region will not changed

pkg/providers/v1/topology.go
wwvela: revised

pkg/providers/v1/aws_ec2.go
wwvela: changed to `return []*ec2.InstanceTopology{}, nil`

pkg/providers/v1/topology.go
```diff
@@ -0,0 +1,70 @@
...
+	unsupportedInstance []string
+	unsupportedRegion   []string
+}
+
+func (t *topologyCache) getNodeTopology(instanceType string, region string, instanceID string) (*ec2.InstanceTopology, error) {
+	t.mutex.RLock()
+	defer t.mutex.RUnlock()
+	if t.mightSupportTopology(instanceType, region) {
+		topologyRequest := &ec2.DescribeInstanceTopologyInput{InstanceIds: []*string{&instanceID}}
+		topology, err := t.cloud.ec2.DescribeInstanceTopology(topologyRequest)
```
mmerkes, line 40: Per offline discussion, we need to gracefully handle access denied when making this API call. Non-EKS customers may not have enough permissions to make their call, and they can update them if they want this functionality.

pkg/providers/v1/iface/types.go
```diff
@@ -8,6 +8,7 @@ import "github.com/aws/aws-sdk-go/service/ec2"
 type EC2 interface {
 	// Query EC2 for instances matching the filter
 	DescribeInstances(request *ec2.DescribeInstancesInput) ([]*ec2.Instance, error)
+	DescribeInstanceTopology(request *ec2.DescribeInstanceTopologyInput) ([]*ec2.InstanceTopology, error)
```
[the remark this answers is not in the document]
wwvela: revised

pkg/providers/v1/instances_v2.go
```diff
@@ -22,10 +22,10 @@ package aws
 
 import (
 	"context"
-
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	cloudprovider "k8s.io/cloud-provider"
+	"strconv"
```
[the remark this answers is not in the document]
wwvela: revised

pkg/providers/v1/instances_v2.go
```diff
@@ -103,7 +114,12 @@ func (c *Cloud) InstanceMetadata(ctx context.Context, node *v1.Node) (*cloudprov
 		return nil, err
 	}
 
-	additionalLabels, err := c.getAdditionalLabels(zone.FailureDomain)
+	instanceID, err := KubernetesInstanceID(providerID).MapToAWSInstanceID()
+	if err != nil {
+		return nil, err
+	}
+
+	additionalLabels, err := c.getAdditionalLabels(zone.FailureDomain, string(instanceID), instanceType, zone.Region)
```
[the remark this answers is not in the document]
wwvela: it should not cause ec2 limit. The topologyCache is designed to avoid this situation, the describeInstanceTopology api will only triggered once for those unsupported regions and nodes with unsupported instance types and will cache it in the topologyCache. So describeInstanceTopology api will only triggered for supported instance type or those not cached unsupported instance types.

hack/e2e/run.sh
```diff
@@ -128,7 +128,7 @@ if [[ "${UP}" = "yes" ]]; then
       --run-id="${test_run_id}" \
       --cloud-provider=aws \
       --cluster-name="${CLUSTER_NAME}" \
-      --create-args="--dns=none --zones=${ZONES} --node-size=m5.large --master-size=m5.large --override=cluster.spec.cloudControllerManager.cloudProvider=aws --override=cluster.spec.cloudControllerManager.image=${IMAGE_NAME}:${IMAGE_TAG}" \
+      --create-args="--dns=none --zones=${ZONES} --node-size=p4d.24xlarge --master-size=m5.large --override=cluster.spec.cloudControllerManager.cloudProvider=aws --override=cluster.spec.cloudControllerManager.image=${IMAGE_NAME}:${IMAGE_TAG}" \
```
mmerkes, line 131: This is not going to work. These instance types are super expensive and not readily available, so it won't be practical to test here.

tests/e2e/nodes.go
```diff
@@ -44,4 +47,62 @@ var _ = Describe("[cloud-provider-aws-e2e] nodes", func() {
...
+
+		if len(nodeList.Items) < 2 {
+			framework.Failf("Conformance requires at least two nodes")
+		}
+		
+		clientConfig, err := framework.LoadConfig()
+    	framework.ExpectNoError(err)
+
+    	client, err := kubernetes.NewForConfig(clientConfig)
+    	framework.ExpectNoError(err)
```
mmerkes, line 64: spacing is off here

tests/e2e/nodes.go
```diff
@@ -44,4 +47,62 @@ var _ = Describe("[cloud-provider-aws-e2e] nodes", func() {
 			gomega.Expect(node.Labels).To(gomega.HaveKey("topology.k8s.aws/zone-id"))
 		}
 	})
+
+	It("should correctly label nodes with instance type p3dn.24xlarge", func(ctx context.Context) {
```
mmerkes, line 51: Test name is out of date.

tests/e2e/nodes.go
mmerkes: Should be able to just run `go fmt` or whatever the k8s dev command is.

pkg/providers/v1/topology.go
wwvela: add the grace handle for permission missing and add the e2e test

hack/e2e/run.sh
wwvela: But the instance type supported by the DescribeInstanceTopology are expensive. So I'm not sure if can still use the e2e to test this change or it will be better to test it by creating a dev cluster and replace the CCM manually

tests/e2e/nodes.go
wwvela: Revised

tests/e2e/nodes.go
wwvela: Revised and verified by `make check`

pkg/providers/v1/aws_ec2.go
mmerkes: For our current usage, a lack of pagination should be fine. The only risk would be if we were re-using for some other use where we were describing multiple instances, which, AFAIK, isn't the type of thing that we ever do in the cloud provider.

I could be wrong, but when you're supplying a primary key like an instance ID, I've never seen EC2 not return the resource in question on the first call.

pkg/providers/v1/topology.go
```diff
@@ -0,0 +1,74 @@
...
+		if err != nil {
+			klog.Errorf("Error describing instance topology: %q", err)
```
mmerkes, line 42: nit: This should be a warn, or have a log for each branch so that there's an explicit log for each scenario.

pkg/providers/v1/topology.go
```diff
@@ -0,0 +1,74 @@
...
+			if strings.Contains(err.Error(), "The functionality you requested is not available in this region") {
+				t.unsupportedRegion = append(t.unsupportedRegion, region)
+				return nil, nil
+			} else if strings.Contains(err.Error(), "You are not authorized to perform this operation") {
+				// gracefully handle the DecribeInstanceTopology access missing error
+				klog.Infof("Not authorized to perform: ec2:DescribeInstanceTopology, permission missing")
+				return nil, nil
+			}
```
mmerkes, line 50: @cartermckinnon Does this cover all of the scenarios that you talked about that should be gracefully handled?

pkg/providers/v1/topology.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"k8s.io/klog/v2"
+)
+
+type topologyCache struct {
+	cloud               *Cloud
+	mutex               sync.RWMutex
+	unsupportedInstance []string
+	unsupportedRegion   []string
```
mmerkes, line 32: nit: Wouldn't a map be more efficient for these?

pkg/providers/v1/topology.go
```diff
@@ -0,0 +1,74 @@
...
+	if t.unsupportedInstance == nil {
+		t.unsupportedInstance = []string{}
+	}
+	if t.unsupportedRegion == nil {
+		t.unsupportedRegion = []string{}
+	}
+	// if both instanceType and region are not in unsupported cache, the instance type and region might be supported
+	// or we haven't check the supportness and cache it yet. If they are unsupported and not cached, we will run
+	// describeTopology api once for them
+	// Initialize the map if it's unset
```
mmerkes, line 72: nit: remove

tests/e2e/nodes.go
```diff
@@ -44,4 +47,60 @@ var _ = Describe("[cloud-provider-aws-e2e] nodes", func() {
...
+					Verb:     "get",
+				},
+			},
+		}
+		result, err := client.AuthorizationV1().SelfSubjectAccessReviews().Create(context.TODO(), ssar, metav1.CreateOptions{})
+		if err != nil {
+			framework.Failf("Error checking EC2 describeInstanceTopology access: %v", err)
+		}
+		allowed := result.Status.Allowed
+		supportedInstanceType := "p4d.24xlarge"
```
mmerkes, line 78: nit: May be brittle if topology information is ever backported to vanilla instances, though would be easy enough to fix.

tests/e2e/nodes.go
```diff
@@ -44,4 +47,60 @@ var _ = Describe("[cloud-provider-aws-e2e] nodes", func() {
...
+					Resource: "describeInstanceTopology",
+					Verb:     "get",
+				},
+			},
+		}
+		result, err := client.AuthorizationV1().SelfSubjectAccessReviews().Create(context.TODO(), ssar, metav1.CreateOptions{})
+		if err != nil {
+			framework.Failf("Error checking EC2 describeInstanceTopology access: %v", err)
+		}
+		allowed := result.Status.Allowed
```
mmerkes, line 77: Is this not something that we can have expectations on what it will be?

pkg/providers/v1/topology.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (t *topologyCache) getNodeTopology(instanceType string, region string, instanceID string) (*ec2.InstanceTopology, error) {
+	t.mutex.RLock()
+	defer t.mutex.RUnlock()
+	if t.mightSupportTopology(instanceType, region) {
+		topologyRequest := &ec2.DescribeInstanceTopologyInput{InstanceIds: []*string{&instanceID}}
+		topology, err := t.cloud.ec2.DescribeInstanceTopology(topologyRequest)
+		if err != nil {
+			klog.Errorf("Error describing instance topology: %q", err)
+			if strings.Contains(err.Error(), "The functionality you requested is not available in this region") {
```
cartermckinnon, line 43: we should be casting these to an `awserr` and then comparing against error code constants, instead of doing string things.

see: https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/handling-errors.html

pkg/providers/v1/topology.go
```diff
@@ -0,0 +1,74 @@
...
+				return nil, nil
+			} else if strings.Contains(err.Error(), "You are not authorized to perform this operation") {
+				// gracefully handle the DecribeInstanceTopology access missing error
+				klog.Infof("Not authorized to perform: ec2:DescribeInstanceTopology, permission missing")
+				return nil, nil
+			}
+			return nil, err
+		} else if len(topology) == 0 {
+			// instanceType is not support topology info and the result is empty
+			t.unsupportedInstance = append(t.unsupportedInstance, instanceType)
```
cartermckinnon, line 54: you haven't acquired a write lock before modifying supportedInstance/supportedRegion

pkg/providers/v1/topology.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (t *topologyCache) mightSupportTopology(instanceType string, region string) bool {
+	// Initialize the map if it's unset
+	if t.unsupportedInstance == nil {
+		t.unsupportedInstance = []string{}
+	}
+	if t.unsupportedRegion == nil {
+		t.unsupportedRegion = []string{}
+	}
```
cartermckinnon, line 68: This can be done when you create the `topologyCache`, and I'd also move the locking down into this function instead of assuming the caller has acquired the correct lock

pkg/providers/v1/topology.go
cartermckinnon: Yep I believe so

mmerkes:
I'm taking over this change from @wwvela in #1038. She's on vacation, so I'm trying to get this over the line. I've done some significant refactoring to address issues discussed in this review, even some I said I wouldn't fix...

1. Paginate DescribeInstanceTopology for completeness.
2. Expire cache after 24 hours when checking of network topology information is available.
3. Optimize adding additional labels: Both the zone ID and network topology labels will only be added if they don't already exist. In the previous implementation, we'd call DescribeInstanceTopology every time a node with a topology was available.
4. Properly handles error codes from AWS API call
5. The caching implementation should solve concurrency issues and remove the need to use a mutex

I don't see much for unit tests here, so I plan on adding those this morning. FYI @cartermckinnon @ndbaker1 @kmala

mmerkes:
/close

Completed in #1038

PR #1028, opened 2024-09-24
Make metadata logging best-effort

### LINKED ISSUE #1027: Fatal error when AWS_EC2_METADATA_DISABLED is true
https://github.com/kubernetes/cloud-provider-aws/pull/990 introduced a new log line that tries to get information from the instance metadata service.  However, when `AWS_EC2_METADATA_DISABLED` is `true`, `GetInstanceIdentityDocument()` returns an fatal error.

```
F0924 09:21:48.998037       1 main.go:104] Cloud provider could not be initialized: could not init cloud provider "aws": error creating AWS metadata client: "unable to get instance identity document: EC2MetadataRequestError: failed to get EC2 instance identity document\ncaused by: RequestCanceled: EC2 IMDS access disabled via AWS_EC2_METADATA_DISABLED env var"
```

This variable is commonly set is situations where the AWS CCM is running on an instance in an AWS account different from the account in which it operates/acts.

There are two fixes that come to mind:
1. Make the logging best-effort by ignoring any error from `GetInstanceIdentityDocument()` or
1. Skip the logging block if the `AWS_EC2_METADATA_DISABLED` is `true`

I'm not that confident that `AWS_EC2_METADATA_DISABLED` being `true` is the only situation in which `GetInstanceIdentityDocument()` fails, so I'm leaning toward option 1.

cc @dims

sjenning:
Fixes #1027

/kind bug

```release-note
Fixes an issue introduced in v1.31.0 where a fatal error occurs when AWS_EC2_METADATA_DISABLED is true`
```

### DIFF
--- a/pkg/providers/v1/aws_sdk.go
+++ b/pkg/providers/v1/aws_sdk.go
@@ -211,18 +211,17 @@ func (p *awsSDKProvider) Metadata() (config.EC2Metadata, error) {
 	p.addAPILoggingHandlers(&client.Handlers)
 
 	identity, err := client.GetInstanceIdentityDocument()
-	if err != nil {
-		return nil, fmt.Errorf("unable to get instance identity document: %v", err)
+	if err == nil {
+		klog.InfoS("instance metadata identity",
+			"region", identity.Region,
+			"availability-zone", identity.AvailabilityZone,
+			"instance-type", identity.InstanceType,
+			"architecture", identity.Architecture,
+			"instance-id", identity.InstanceID,
+			"private-ip", identity.PrivateIP,
+			"account-id", identity.AccountID,
+			"image-id", identity.ImageID)
 	}
-	klog.InfoS("instance metadata identity",
-		"region", identity.Region,
-		"availability-zone", identity.AvailabilityZone,
-		"instance-type", identity.InstanceType,
-		"architecture", identity.Architecture,
-		"instance-id", identity.InstanceID,
-		"private-ip", identity.PrivateIP,
-		"account-id", identity.AccountID,
-		"image-id", identity.ImageID)
 	return client, nil
 }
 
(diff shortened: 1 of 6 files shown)


sjenning:
```release-note
Fixes an issue first included in v1.31.0 where a fatal error occurs when AWS_EC2_METADATA_DISABLED is true`
```

dims:
@sjenning looks good to me! looks like there may be unrelated reasons for the ci failures though :(

sjenning:
@dims tests are green now.  Can you approve/lgtm?

pkg/providers/v1/aws_sdk.go
```diff
@@ -211,18 +211,17 @@ func (p *awsSDKProvider) Metadata() (config.EC2Metadata, error) {
 	p.addAPILoggingHandlers(&client.Handlers)
 
 	identity, err := client.GetInstanceIdentityDocument()
-	if err != nil {
-		return nil, fmt.Errorf("unable to get instance identity document: %v", err)
```
cartermckinnon, line 215: Can you add a log line for this error?

sjenning: Thanks for the review!

The line would be logged for every invocation of the function and it would basically be a log line that says  "I wanted to log something useful but the call failed so I'm logging the fact that I couldn't log something useful".

How often is this function called?

cartermckinnon: Calls to IMDS can fail for reasons other than `AWS_EC2_METADATA_DISABLED`, so I think its worth logging. It's only called once in an `init()`: https://github.com/kubernetes/cloud-provider-aws/blob/d7e05d57709cd46297490b51ce0dd11a54dbea35/pkg/providers/v1/aws.go#L473

kmala:
we also need to update the go version to update to 1.22.7 for the govulncheck to succeed. Can you do it if possible else i can make a new PR for it?

ialidzhikov:
Can we proceed with merging this PR?

PR #1019, opened 2024-08-18
Switch to latest master of kops

dims:
the URL we are using seems to have gotten outdated. let's use kops master latest url.

```
❯ curl https://storage.googleapis.com/kops-ci/bin/latest-ci-updown-green.txt
https://storage.googleapis.com/k8s-staging-kops/kops/releases/1.30.0-beta.2+v1.30.0-beta.1-35-gcd7fba97d5

❯ curl https://storage.googleapis.com/k8s-staging-kops/kops/releases/markers/master/latest-ci-updown-green.txt
https://storage.googleapis.com/k8s-staging-kops/kops/releases/1.30.0-beta.2+v1.30.0-beta.1-110-g5d4d867086
```

/assign @hakuna-matatah

PR #1033, opened 2024-10-12
Bump go mod version

ndbaker1:
**What type of PR is this?**
/kind cleanup

Addresses govulncheck vulnerabilities with go version and project dependencies

Fixes #

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module k8s.io/cloud-provider-aws
 
-go 1.22.5
+go 1.22.7
 
 require (
 	github.com/aws/aws-sdk-go v1.51.1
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module k8s.io/cloud-provider-aws
 
-go 1.22.5
+go 1.22.7
```
kmala, line 3: can you also update the other places like dockerfile, cloud build etc where go version is used.

ndbaker1: ack done, not seeing any other places where the version pops up

kmala: can you also update https://github.com/kubernetes/cloud-provider-aws/blob/release-1.30/.ko.yaml and we should be good

ndbaker1: ah hidden dotfiles 😠
updated, i'll do the same to PRs for other branches after this one's merged

kmala: can you cherry pick https://github.com/kubernetes/cloud-provider-aws/pull/1019/files to fix the build issues

PR #1089, opened 2025-01-14
Use typed client-go constructors for rate-limit related objects

yue9944882:
**What type of PR is this?**
/kind cleanup

Since k8s 1.31, a few client-go types including rate-limitter have onboarded to golang generics which is available for golang 1.18+. This PR switches the controller to use the generic types from k8s.io dependencies.

**Special notes for your reviewer**: Trivial cleanups

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/tagging/tagging_controller.go
+++ b/pkg/controllers/tagging/tagging_controller.go
@@ -116,30 +116,32 @@ func NewTaggingController(
 		return nil, err
 	}
 
-	var rateLimiter workqueue.RateLimiter
+	var rateLimiter workqueue.TypedRateLimiter[any]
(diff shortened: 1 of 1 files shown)


pkg/controllers/tagging/tagging_controller.go
```diff
@@ -116,30 +116,32 @@ func NewTaggingController(
 		return nil, err
 	}
 
-	var rateLimiter workqueue.RateLimiter
+	var rateLimiter workqueue.TypedRateLimiter[any]
```
cartermckinnon, line 119: What's the benefit of this implementation if we're using `any` as the type variable?

PR #1091, opened 2025-01-16
optimize tagging controller workqueue handling

kmala:
**What type of PR is this?**

 /kind bug

The tagging controller can add the same node multiple times to the workqueue since the node is updated frequently by the kubelet if the add event for that node is not processed yet. This is because for the workqueue add we are not adding an object which is different for each event. When a large number of nodes are added at a time, the queue can become so big that it can take hours/days for all the events to be handled.
Removed the time and retry count from the workqueue object as they are provided by default by the workqueue library and made the workqueue object simple such that multiple add/updates to a node will only add one item to the workqueue.

Fixes #

```release-note
cloudprovider_aws_tagging_controller_work_item_duration_seconds is no longer published by the controller and instead the same details can be fetched from the default work queue metrics(https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/metrics/prometheus/workqueue/metrics.go#L40)
```

### DIFF
--- a/pkg/controllers/tagging/metrics.go
+++ b/pkg/controllers/tagging/metrics.go
@@ -14,23 +14,15 @@ limitations under the License.
 package tagging
 
 import (
+	"sync"
+
 	"k8s.io/component-base/metrics"
 	"k8s.io/component-base/metrics/legacyregistry"
-	"sync"
 )
 
 var register sync.Once
 
 var (
-	workItemDuration = metrics.NewHistogramVec(
-		&metrics.HistogramOpts{
-			Name:           "cloudprovider_aws_tagging_controller_work_item_duration_seconds",
-			Help:           "workitem latency of workitem being in the queue and time it takes to process",
-			StabilityLevel: metrics.ALPHA,
-			Buckets:        metrics.ExponentialBuckets(0.5, 1.5, 20),
-		},
-		[]string{"latency_type"})
-
 	workItemError = metrics.NewCounterVec(
 		&metrics.CounterOpts{
 			Name:           "cloudprovider_aws_tagging_controller_work_item_errors_total",
@@ -43,13 +35,8 @@ var (
 // registerMetrics registers tagging-controller metrics.
 func registerMetrics() {
 	register.Do(func() {
-		legacyregistry.MustRegister(workItemDuration)
 		legacyregistry.MustRegister(workItemError)
 	})
-}
-
-func recordWorkItemLatencyMetrics(latencyType string, timeTaken float64) {
-	workItemDuration.With(metrics.Labels{"latency_type": latencyType}).Observe(timeTaken)
 }
 
 func recordWorkItemErrorMetrics(errorType string, instanceID string) {
--- a/pkg/controllers/tagging/tagging_controller.go
+++ b/pkg/controllers/tagging/tagging_controller.go
@@ -22,6 +22,7 @@ import (
 
 	"golang.org/x/time/rate"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	coreinformers "k8s.io/client-go/informers/core/v1"
@@ -42,16 +43,21 @@ func init() {
 	registerMetrics()
 }
 
-// workItem contains the node and an action for that node
+// taggingControllerNode contains the node details required for tag/untag of node resources.
+type taggingControllerNode struct {
+	providerID string
+	name       string
+}
+
+// workItem contains the node name, provider id and an action for that node.
 type workItem struct {
-	node           *v1.Node
-	action         func(node *v1.Node) error
-	requeuingCount int
-	enqueueTime    time.Time
+	name       string
+	providerID string
+	action     string
 }
 
 func (w workItem) String() string {
-	return fmt.Sprintf("[Node: %s, RequeuingCount: %d, EnqueueTime: %s]", w.node.GetName(), w.requeuingCount, w.enqueueTime)
+	return fmt.Sprintf("[Node: %s, Action: %s]", w.name, w.action)
 }
 
 const (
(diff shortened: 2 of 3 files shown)


dims:
/hold

give me a bit please to review this!

dims:
/hold cancel

Looks good to me @kmala

pkg/controllers/tagging/metrics.go
```diff
@@ -14,23 +14,15 @@ limitations under the License.
...
+
 	"k8s.io/component-base/metrics"
 	"k8s.io/component-base/metrics/legacyregistry"
-	"sync"
 )
 
 var register sync.Once
 
 var (
-	workItemDuration = metrics.NewHistogramVec(
```
dims, line 25: are we ok dropping this metric?

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -277,11 +282,19 @@ func (tc *Controller) process() bool {
 
 // tagNodesResources tag node resources
 // If we want to tag more resources, modify this function appropriately
-func (tc *Controller) tagNodesResources(node *v1.Node) error {
+func (tc *Controller) tagNodesResources(node *taggingControllerNode) error {
 	for _, resource := range tc.resources {
 		switch resource {
 		case opt.Instance:
-			err := tc.tagEc2Instance(node)
+			v1node, err := tc.nodeInformer.Lister().Get(node.name)
```
dims, line 289: Do we want to try to capture all necessary information in taggingControllerNode itself to avoid this lookup?

dims: (in a follow up!)

pkg/controllers/tagging/metrics.go
cartermckinnon: We should have an equivalent metric from the workqueue package itself via: https://github.com/kubernetes/cloud-provider-aws/blob/bf18a5157151081039ac8e9a4882df2e923fb3bd/pkg/controllers/tagging/tagging_controller.go#L35

but please make sure we actually scrape that one @kmala !

kmala: work queue metrics are available by default https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/metrics/prometheus/workqueue/metrics.go#L40 and i had verified the same the metrics endpoint of the CCM will provide the tagging controller metrics

pkg/controllers/tagging/tagging_controller.go
kmala: we need to have a lookup if we want to avoid having multiple items in workqueue since the labels can be different on different updates.

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -367,12 +380,13 @@ func (tc *Controller) untagEc2Instance(node *v1.Node) error {
...
-		action:         action,
-		requeuingCount: 0,
-		enqueueTime:    time.Now(),
+func (tc *Controller) enqueueNode(node *v1.Node, action string) {
+	// if the struct has fields which are all comparable then the workqueue add will handle make sure multiple adds of the same object
+	// will only have one item in the workqueue.
+	item := workItem{
+		name:       node.GetName(),
+		providerID: node.Spec.ProviderID,
+		action:     action,
```
cartermckinnon, line 389: The action `func` wasn't comparable, so this change allows multiple workitems for the same node, with different actions. Is that desirable? Why change action to a string?

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -42,16 +43,21 @@ func init() {
...
+	providerID string
+	name       string
+}
+
+// workItem contains the node name, provider id and an action for that node.
 type workItem struct {
-	node           *v1.Node
-	action         func(node *v1.Node) error
-	requeuingCount int
-	enqueueTime    time.Time
```
cartermckinnon, line 50: So these 2 fields were the crux of the issue, right? If a workItem already existed in the queue, it's `enqueueTime` would always be different, even if it's `requeingCount` happened to match (`0`).

pkg/controllers/tagging/tagging_controller.go
cartermckinnon: The `*Node` should point to the same struct though, right? the watcher/lister should only have one copy of the Node in memory

pkg/controllers/tagging/metrics.go
cartermckinnon: Yes we've had those metrics for a while because of the import above (you have to import it for these to be registered: https://github.com/kubernetes/kubernetes/blob/d06398aac3e378da0c95472cd39713998206e9ff/staging/src/k8s.io/component-base/metrics/prometheus/workqueue/metrics.go#L26)

But do we actually scrape those? I'm pretty sure we've been scraping this other metric that we're removing

cartermckinnon:
@ndbaker1 can you review this one?

pkg/controllers/tagging/tagging_controller.go
ndbaker1: I'm thinking the same. Seems we're relying on queuing-order to determine the final state which boils down to a last-write-wins situation in the workqueue. would any edge cases with instance termination would break that assumption

pkg/controllers/tagging/tagging_controller.go
kmala: yes, mostly

pkg/controllers/tagging/tagging_controller.go
kmala: > so this change allows multiple workitems for the same node, with different actions. Is that desirable? 

Before we are adding to workitems for every update if the add hasn't happened and there could be work item for the delete also. With this change we would only have maximum of 2 work items in the queue which should be the same behavior as before as as single or multiple add tag work items would mean the same behavior at the end.

the action to a string is made so that its comparable and will prevent duplicate updates to not add to the workqueue.

pkg/controllers/tagging/metrics.go
kmala: yes, these metrics are available and i had verified the same by querying the metrics endpoint.
what do you mean by scrape those?

pkg/controllers/tagging/tagging_controller.go
Issacwww: Correct me if I am wrong, the `requeuingCount` looks like there is a maximum retry. Are we good to remove it?
Is there a option to keep those fields and only compare the node identity fields when enqueue?

kmala: we can get the requeue count from the workqueue itself and i changed the code to use that. So the functionality remains the same as before.

PR #1109, opened 2025-02-23
Optimize node controller describe instances calls

shiv-amz:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:

Currently, we make multiple ```ec2:DescribeInstances``` API calls to populate various fields in InstanceMetadata. This PR consolidates multiple API calls into a single request per instance, aiming to improve node join latency.

#### Does this PR introduce a user-facing change?

No functional changes; the behavior remains the same, with fewer ec2:DescribeInstances API calls.

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -857,6 +857,10 @@ func (c *Cloud) NodeAddressesByProviderID(ctx context.Context, providerID string
 		return nil, err
 	}
 
+	return c.getInstanceNodeAddress(instance)
+}
+
+func (c *Cloud) getInstanceNodeAddress(instance *ec2.Instance) ([]v1.NodeAddress, error) {
 	var addresses []v1.NodeAddress
 
 	for _, family := range c.cfg.Global.NodeIPFamilies {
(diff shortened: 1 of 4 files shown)


pkg/providers/v1/aws.go
```diff
@@ -857,6 +857,10 @@ func (c *Cloud) NodeAddressesByProviderID(ctx context.Context, providerID string
 		return nil, err
 	}
 
+	return c.getInstanceNodeAddress(instance)
```
mmerkes, line 860: Do we actually need the above method anymore? Some of these methods may actually be obsolete after migrating to the Instances V2 API. I would confirm.

pkg/providers/v1/instances_v2.go
```diff
@@ -118,25 +120,43 @@ func (c *Cloud) InstanceMetadata(ctx context.Context, node *v1.Node) (*cloudprov
...
-	zone, err := c.GetZoneByProviderID(ctx, providerID)
-	if err != nil {
-		return nil, err
-	}
+	var (
+		instanceType  string
+		zone          cloudprovider.Zone
+		nodeAddresses []v1.NodeAddress
+	)
+	if variant.IsVariantNode(string(instanceID)) {
```
mmerkes, line 133: You might consider refactoring to pull all of this stuff out because we often get burned from the Fargate-specific stuff. We'll need to be sure to verify that there's no regression there. Rather than an if/else, probably best to just pull all of that out one or another.

cartermckinnon (commented):
One nit, but LGTM.

pkg/providers/v1/instances_v2.go
shiv-amz: InstanceMetadata is called for both fargate and ec2 by controllers. Currently variant logic is being handled by API's (InstanceTypeByProviderID, GetZoneByProviderID, NodeAddressesByProviderID) which branch out for fargate versus ec2 variants. However these API's are also directly called by controllers, for instances V1 implementation.

I doubt we can remove the variant check without broader refactoring.

PR #1112, opened 2025-02-28
Adding sagemaker instance format parse logic

yue9944882:
This PR allows CCM to parse AWS sagemaker instance.


cartermckinnon:
/hold

SageMaker nodes should be handled via the `variant` path (that's what it was added for). Following up internally about this...

PR #1116, opened 2025-03-14
feat: add tagging controller delays and work queue size metrics

shvbsle:
**What type of PR is this?**
>
> /kind feature
> /kind flake

Adds tagging controller delays metric. This measure the delay between node-creation and tagging of the EC2 Instance. The metrics should show up like so:

```
inital_node_tagging_delay_seconds [ALPHA] Latency (in seconds) between node creation and its first successful tagging by TaggingController.
# TYPE tagging_controller_inital_node_tagging_delay_seconds histogram
tagging_controller_inital_node_tagging_delay_seconds_bucket{le="1"} 0
tagging_controller_inital_node_tagging_delay_seconds_bucket{le="4"} 0
tagging_controller_inital_node_tagging_delay_seconds_bucket{le="16"} 0
tagging_controller_inital_node_tagging_delay_seconds_bucket{le="64"} 0
tagging_controller_inital_node_tagging_delay_seconds_bucket{le="256"} 0
tagging_controller_inital_node_tagging_delay_seconds_bucket{le="1024"} 0
tagging_controller_inital_node_tagging_delay_seconds_bucket{le="+Inf"} 0
tagging_controller_inital_node_tagging_delay_seconds_sum 0
tagging_controller_inital_node_tagging_delay_seconds_count 0
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/tagging/metrics.go
+++ b/pkg/controllers/tagging/metrics.go
@@ -44,20 +44,13 @@ var (
 			StabilityLevel: metrics.ALPHA,
 		},
 	)
-	workQueueSize = metrics.NewGauge(
-		&metrics.GaugeOpts{
-			Subsystem:      metricsSubsystem,
-			Name:           "work_queue_size",
-			Help:           "Current size of the work queue.",
-			StabilityLevel: metrics.ALPHA,
-		},
-	)
 )
 
 // registerMetrics registers tagging-controller metrics.
 func registerMetrics() {
 	register.Do(func() {
 		legacyregistry.MustRegister(workItemError)
+		legacyregistry.MustRegister(nodeTaggingDelay)
 	})
 }
 
--- a/pkg/controllers/tagging/metrics.go
+++ b/pkg/controllers/tagging/metrics.go
@@ -20,6 +20,10 @@ import (
 	"k8s.io/component-base/metrics/legacyregistry"
 )
 
+const (
+	metricsSubsystem = "tagging_controller"
+)
+
 var register sync.Once
 
 var (
@@ -30,6 +34,24 @@ var (
 			StabilityLevel: metrics.ALPHA,
 		},
 		[]string{"error_type", "instance_id"})
+
+	nodeTaggingDelay = metrics.NewHistogram(
+		&metrics.HistogramOpts{
+			Subsystem:      metricsSubsystem,
+			Name:           "node_tagging_delay_seconds",
+			Help:           "Number of seconds after node creation when TaggingController successfully tagged or untagged the node resources.",
+			Buckets:        metrics.ExponentialBuckets(1, 4, 6), // 1s -> ~17m
+			StabilityLevel: metrics.ALPHA,
+		},
+	)
+	workQueueSize = metrics.NewGauge(
+		&metrics.GaugeOpts{
+			Subsystem:      metricsSubsystem,
+			Name:           "work_queue_size",
+			Help:           "Current size of the work queue.",
+			StabilityLevel: metrics.ALPHA,
+		},
+	)
 )
 
 // registerMetrics registers tagging-controller metrics.
--- a/pkg/controllers/tagging/metrics.go
+++ b/pkg/controllers/tagging/metrics.go
@@ -35,11 +35,11 @@ var (
 		},
 		[]string{"error_type", "instance_id"})
 
-	nodeTaggingDelay = metrics.NewHistogram(
+	initialNodeTaggingDelay = metrics.NewHistogram(
 		&metrics.HistogramOpts{
 			Subsystem:      metricsSubsystem,
-			Name:           "node_tagging_delay_seconds",
-			Help:           "Number of seconds after node creation when TaggingController successfully tagged or untagged the node resources.",
+			Name:           "inital_node_tagging_delay_seconds",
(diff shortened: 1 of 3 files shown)


pkg/controllers/tagging/metrics.go
```diff
@@ -30,6 +34,24 @@ var (
...
+	nodeTaggingDelay = metrics.NewHistogram(
+		&metrics.HistogramOpts{
+			Subsystem:      metricsSubsystem,
+			Name:           "node_tagging_delay_seconds",
+			Help:           "Number of seconds after node creation when TaggingController successfully tagged or untagged the node resources.",
+			Buckets:        metrics.ExponentialBuckets(1, 4, 6), // 1s -> ~17m
+			StabilityLevel: metrics.ALPHA,
+		},
+	)
+	workQueueSize = metrics.NewGauge(
```
cartermckinnon, line 47: I *think* we should already have an equivalent metric from the `workqueue` itself: https://github.com/kubernetes/cloud-provider-aws/blob/9358055c45bfad50f7d972571c1efbe5fa447f9f/pkg/controllers/tagging/tagging_controller.go#L36

they're defined here: https://github.com/kubernetes/client-go/blob/master/util/workqueue/metrics.go

and the metrics should be emitted with the `Name` of the workqueue (probably a prefix?): https://github.com/kubernetes/cloud-provider-aws/blob/9358055c45bfad50f7d972571c1efbe5fa447f9f/pkg/controllers/tagging/tagging_controller.go#L147

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -269,6 +274,7 @@ func (tc *Controller) process() bool {
 		}
 
 		tc.workqueue.Forget(obj)
+		nodeTaggingDelay.Observe(time.Since(currentNode.CreationTimestamp.Time).Seconds())
```
cartermckinnon, line 277: I think we want this within tagEc2Resource, around: https://github.com/kubernetes/cloud-provider-aws/blob/9358055c45bfad50f7d972571c1efbe5fa447f9f/pkg/controllers/tagging/tagging_controller.go#L343

Since we only want to record it when we're tagging (not un-tagging) and we don't want to record it unless tags are actually applied

pkg/controllers/tagging/metrics.go
shvbsle: Confirmed that we have equivalent metric for worker queue size already. They show up under the label `tagging-controller`. It is this: 
```
workqueue_depth{name="tagging-controller"} 0
```
Will remove `workQueueSize`

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -342,6 +342,7 @@ func (tc *Controller) tagEc2Instance(node *v1.Node) error {
 
 	klog.Infof("Successfully labeled node %s with %v.", node.GetName(), labels)
 
+	nodeTaggingDelay.Observe(time.Since(node.CreationTimestamp.Time).Seconds())
```
kmala, line 345: why node creation time? there can also be an update to the tags which would cause retag right?
If we want to know just for the current iteration we already have work queue metrics

shvbsle: Intent is to observe tagging delays especially during the node-startup and in clusters with a large number of nodes. Since the emitted metric is a Histogram, a re-tag event would fall in Inf+ buckets. The histogram would still allow us to get a reliable p90 metric to observe delays during node-startup

kmala: How would this be different from the work queue metrics as the only additional thing this might add is the time it takes to add to the workqueue which should be mostly quick/immediate. Otherwise this is mostly sum of work queue latency and work duration metrics right?

shvbsle: Currently we are only interested in tagging delays. IIUC, `workqueue_queue_duration_seconds_bucket` does not provide a way to distinguish between a tagging, an untagging or an error event, which makes it an unreliable proxy for what we want to measure.

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -398,6 +402,21 @@ func (tc *Controller) enqueueNode(node *v1.Node, action string) {
...
+
+	if _, ok := node.Labels[taggingControllerLabelKey]; ok {
+		return false
+	}
+
+	// If the code reaches this section then the node has some labels but not tagging controller's labels
+	// This is a first-time tagging scenario
+	return true
+
+}
```
cartermckinnon, line 418:
```suggestion
func (tc *Controller) isInitialTag(node *v1.Node) bool {
	_, ok := node.Labels[taggingControllerLabelKey]
	return !ok
}
```

cartermckinnon:
disregard comment about placement of isInitialTag, misread the code. We don't ever set the labels on this copy of the object

pkg/controllers/tagging/tagging_controller.go
shvbsle: Added test-cases in latest revision.
```
·> go test ./pkg/controllers/tagging -run Test_isInitialTag
ok      k8s.io/cloud-provider-aws/pkg/controllers/tagging       0.029s
```

PR #1117, opened 2025-03-19
Adds a TTL Cache for DescribeInstanceTopology API call response

shvbsle:
**What type of PR is this?**
>
> /kind bug
> /kind feature

cloud-provider's node-controller attempts to update node-status every 5 minutes. If an instance that supports topology labels doesn't already have it keeps calling the DescribeInstanceTopology API repeatedly. This happens because, `UpdateNodeStatus` in cloud-provider does not actually apply `additionalLabels`.

This change caches the DescribeInstanceTopology API call response for an instance-ID and thus reduces the volume of DescribeInstanceTopology API calls made. If a node is terminated, then the TTL cache would evict the cached response eventually thus ensuring that the cache does not grow unbounded. 

Fixes #

```release-note

```

### DIFF
--- a/pkg/resourcemanagers/topology.go
+++ b/pkg/resourcemanagers/topology.go
@@ -32,7 +32,10 @@ import (
 	"k8s.io/klog/v2"
 )
 
-const instanceTopologyManagerCacheTimeout = 24 * time.Hour
+const (
+	instanceTopologyManagerCacheTimeout     = 24 * time.Hour
+	instanceTopologyAPIResponseCacheTimeout = 24 * time.Hour
+)
 
 /*
 We need to ensure that instance types that we expect a response will not successfully complete syncing unless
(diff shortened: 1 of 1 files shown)


pkg/resourcemanagers/topology.go
```diff
@@ -85,12 +102,20 @@ func NewInstanceTopologyManager(ec2 services.Ec2SdkV2, cfg *config.CloudConfig)
 		supportedTopologyInstanceTypePattern: supportedTopologyInstanceTypePattern,
 		// These should change very infrequently, if ever, so checking once a day sounds fair.
 		unsupportedKeyStore: cache.NewTTLStore(topStringKeyFunc, instanceTopologyManagerCacheTimeout),
+		// In this cache we store the response made to DescribeInstanceTopology API for an instanceID
+		instanceTopologyAPIResponseCache: cache.NewTTLStore(topologyCacheKeyFunc, instanceTopologyAPIResponseCacheTimeout),
```
kmala, line 106: Won't it cause an issue with the memory usage if there is some churn of the nodes in a day. Since we are storing topology struct, can we estimate how much memory will be used for an instance which should help justify if its okay or not? Also, is it not possible to remove from cache on node deletion?

shvbsle: Note that we will only cache for instance-types that support topology labels. Now, the instance-topology struct would contain `NetworkNodes` which is a list of strings. Most supported nodes would have three entries in this list and for 100k unique instances, the memory foot-print of caching would come out to ~15mb

> Also, is it not possible to remove from cache on node deletion?

correct me if I'm wrong but I dont think that we are listening for node-deletion events. We'd have to loop through the cache and check if the nodes are deleted which would require additional API calls. The TTL for the cache entry is 24 hours in the worst-case, a deleted node cache entry would be flushed 24 hours later.

cartermckinnon:
/hold

I don’t think this is the right solution to this issue, let’s chat about it

PR #1119, opened 2025-03-20
add ImageId to the set of additional cloud provider labels

### LINKED ISSUE #1118: Export AWS image id as a node label
 This introduces a new label applied to each node in a cluster: `k8s.aws/image-id=`.

**Why is this needed**:
The ImageId is per-node metadata that allows cluster owners to:
  1. Identify the set of image ids running a cluster
  2. Track rollouts of images through a cluster
  3. Quickly identify bad image builds being rolled out to cluster

/kind feature

justinmir:
**What type of PR is this?**
/kind feature

Introduces  new label applied to each node in a cluster: `k8s.aws/image-id=`.

This gives cluster administrators better visibility into images being used by nodes
in their cluster. 

It does not require any additional API calls as the describe instance call already has the ImageId.

Fixes #1118

N/A

```release-note
Export image id as an node label: `node.k8s.aws/image-id=`
```


pkg/providers/v1/well_known_labels.go
```diff
@@ -24,4 +24,6 @@ const (
 	// but will be initially applied to nodes. The suffix should be an incremented
 	// integer starting at 1.
 	LabelNetworkNodePrefix = "topology.k8s.aws/network-node-layer-"
+	// LabelImageID is a machine image label that can be applied to node resources.
+	LabelImageID = "k8s.aws/image-id"
```
kmala: i think `node.k8s.aws/image-id` would be better wdyt?

PR #1120, opened 2025-03-21
Batch tagging controller create/delete tags calls

shiv-amz:
/kind feature

* Tagging controller runs in single threaded loop which can create backlog in clusters with large number of nodes. Added worker count of 10, which can make progress in parallel. 
* Added createTag/deleteTag batcher which batches multiple api calls into a single call with multiple instances.
* Batching is behind option ```tagging-controller-enable-batching``` flag.

We have a similar batching for other EC2 calls in [Karpenter](https://github.com/aws/karpenter-provider-aws/tree/0a8251d6eba027c7d34acaf9d0fee2fc40832b67/pkg/batcher)

NONE


pkg/controllers/options/tagging_controller.go
```diff
@@ -21,10 +21,11 @@ import (
...
 type TaggingControllerOptions struct {
-	Tags       map[string]string
-	Resources  []string
-	RateLimit  float64
-	BurstLimit int
+	Tags           map[string]string
+	Resources      []string
+	RateLimit      float64
+	BurstLimit     int
+	EnableBatching bool
```
cartermckinnon, line 28: What's the rationale for making this configurable? If it's useful, a `BatchSize` config option seems more flexible for tuning.

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -64,6 +65,7 @@ const (
 	taggingControllerLabelKey = "k8s.io/cloud-provider-aws"
 
 	maxRequeuingCount = 9
+	workerQueueCount  = 10
```
cartermckinnon, line 68: This seems like a useful config option, similar to `--concurrent-node-syncs`, `--concurrent-service-syncs`, etc.

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -194,7 +198,13 @@ func (tc *Controller) Run(stopCh <-chan struct{}) {
 	}
 
 	klog.Infof("Starting the tagging controller")
-	go wait.Until(tc.work, tc.nodeMonitorPeriod, stopCh)
+	if tc.enableBatching {
+		for i := int32(0); i < workerQueueCount; i++ {
+			go wait.Until(tc.work, tc.nodeMonitorPeriod, stopCh)
+		}
```
cartermckinnon, line 204: Why is parallelism of API calls gated by the batching-per-call option? these seem like 2 different concerns

pkg/providers/v1/batcher/batcher.go
cartermckinnon, line 1: There's a lot going on here :)

Is there a simpler way we could implement this? Could a goroutine in the tagging controller poll multiple items off the queue (if they're available) and fire off the request? In Karpenter, it makes sense why we'd need a decoupled "batcher" because `controller-runtime` doesn't expose the underlying work queue to `Reconcile`, but we have some more flexibility in this controller

pkg/controllers/options/tagging_controller.go
shiv-amz: Yes, we don't need the enable batching flag. I added this for my load testing, to easily compare results between batching versus single calls. I'll remove this in next rev.

pkg/controllers/tagging/tagging_controller.go
shiv-amz: They are related since, with single concurrent batching does not work. If all calls are made by single thread, they won't ever get batched. So to enable batching, parallel threads consuming from workqueue is required.

As pointed above, will make workerQueueCount as configurable, and remove the batching flag.

cartermckinnon:
Chatted offline -- we'll split this up into 2 PR's, one that adds a simple parallelism mechanism (with a worker ppl), and we'll keep iterating on the batching approach.

pkg/providers/v1/batcher/createTags.go
cartermckinnon, line 1: I think these batcher implementations should go somewhere else, to keep the `batcher` package generic. Maybe just in `providers/v1/tags.go`?

cartermckinnon: Do we really need 2 separate Batcher impls for these API calls? Seems like a lot of shared code. Also don't think most of this needs to be public, could be package-private.

pkg/providers/v1/batcher/batch_test.go
```diff
@@ -0,0 +1,135 @@
...
+	})
+	Context("Concurrency", func() {
+		It("should limit the number of threads that run concurrently from the batcher", func() {
+			// This batcher will get canceled at the end of the test run
+			fakeBatcher = NewFakeBatcher(cancelCtx, time.Minute, 100)
+
+			// Generate 300 items that add to the batcher
+			for i := 0; i < 300; i++ {
+				go func() {
+					fakeBatcher.batcher.Add(cancelCtx, lo.ToPtr(test.RandomName()))
```
cartermckinnon, line 67: if `test.RandomName()` is the only usage of `operatorpkg`, can we drop it and avoid the dependency?

pkg/providers/v1/batcher/batch_test.go
```diff
@@ -0,0 +1,135 @@
...
+	ctx = context.TODO()
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Batcher")
+}
+
+var _ = BeforeSuite(func() {
+	fakeEC2API = &aws.FakeEC2Impl{}
+})
+
+var _ = Describe("Batcher", func() {
```
cartermckinnon, line 47: We don't use ginkgo anywhere else in this project, can you keep this in line with the existing tests?

pkg/providers/v1/batcher/createTags.go
shiv-amz: > Do we really need 2 separate Batcher impls for these API calls? Seems like a lot of shared code.

The input objects are different like ```CreateTagsInput```/```DeleteTagsInput``` which lead to separate implementations. We could come up with a more generic implementation which works with any api call. However settings like ```maxItems``` /```timeouts``` would be different based on the api getting batched.

pkg/providers/v1/batcher/batch_test.go
shiv-amz: I noticed ginko being used in cloud-provider-aws repo for E2E tests. From what I can tell, there aren’t much downsides to using Ginkgo for batching tests.
* [Node](https://github.com/kubernetes/cloud-provider-aws/blob/a1565f3b163367c3a71169fe2fdf9dc58e89e0eb/tests/e2e/nodes.go#L23)
* [suite test](https://github.com/kubernetes/cloud-provider-aws/blob/a1565f3b163367c3a71169fe2fdf9dc58e89e0eb/tests/e2e/suite_test.go#L12)
* [load balancer](https://github.com/kubernetes/cloud-provider-aws/blob/a1565f3b163367c3a71169fe2fdf9dc58e89e0eb/tests/e2e/loadbalancer.go#L17)

pkg/providers/v1/batcher/batch_test.go
shiv-amz: Nice catch. Removed the ```operatorpkg``` dependency.

pkg/providers/v1/batcher/createTags.go
shiv-amz: > Also don't think most of this needs to be public, could be package-private.

Moving internal ones to package-private.

go.mod
```diff
@@ -3,33 +3,47 @@ module k8s.io/cloud-provider-aws
...
 	k8s.io/api v0.33.0-rc.0
 	k8s.io/apimachinery v0.33.0-rc.0
 	k8s.io/client-go v0.33.0-rc.0
-	k8s.io/cloud-provider v0.33.0-rc.0
-	k8s.io/code-generator v0.33.0-rc.0
+	k8s.io/cloud-provider v0.32.2
+	k8s.io/code-generator v0.32.1
 	k8s.io/component-base v0.33.0-rc.0
-	k8s.io/controller-manager v0.33.0-rc.0
+	k8s.io/controller-manager v0.32.2
```
cartermckinnon, line 27: We need to revert this, to stay on 1.33

pkg/providers/v1/createTagsBatch.go
cartermckinnon, line 1: can you use snake-case on these files to match the convention?

pkg/providers/v1/create_tags_batch.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+)
+
+// CreateTagsBatcher contains the batcher details
+type CreateTagsBatcher struct {
+	batcher *batcher2.Batcher[ec2.CreateTagsInput, ec2.CreateTagsOutput]
+}
+
+// NewCreateTagsBatcher creates a NewCreateTagsBatcher object
+func NewCreateTagsBatcher(ctx context.Context, ec2api iface.EC2) *CreateTagsBatcher {
```
cartermckinnon, line 39:
```suggestion
type createTagsBatcher struct {
	batcher *batcher2.Batcher[ec2.CreateTagsInput, ec2.CreateTagsOutput]
}

// newCreateTagsBatcher creates a NewCreateTagsBatcher object
func NewCreateTagsBatcher(ctx context.Context, ec2api iface.EC2) *CreateTagsBatcher {
```

pkg/providers/v1/delete_tags_batch.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+)
+
+// DeleteTagsBatcher contains the batcher details
+type DeleteTagsBatcher struct {
+	batcher *batcher2.Batcher[ec2.DeleteTagsInput, ec2.DeleteTagsOutput]
+}
+
+// NewDeleteTagsBatcher creates a NewDeleteTagsBatcher object
+func NewDeleteTagsBatcher(ctx context.Context, ec2api iface.EC2) *DeleteTagsBatcher {
```
cartermckinnon, line 39: These should be package-private
```suggestion
type deleteTagsBatcher struct {
	batcher *batcher2.Batcher[ec2.DeleteTagsInput, ec2.DeleteTagsOutput]
}

func newDeleteTagsBatcher(ctx context.Context, ec2api iface.EC2) *DeleteTagsBatcher {
```

pkg/providers/v1/create_tags_batch.go
```diff
@@ -0,0 +1,105 @@
...
+limitations under the License.
+*/
+
+package aws
+
+import (
+	"context"
+	"fmt"
+	"github.com/mitchellh/hashstructure/v2"
+	batcher2 "k8s.io/cloud-provider-aws/pkg/providers/v1/batcher"
```
cartermckinnon, line 23: what's this for?

pkg/providers/v1/delete_tags_batch.go
```diff
@@ -0,0 +1,105 @@
...
+limitations under the License.
+*/
+
+package aws
+
+import (
+	"context"
+	"fmt"
+	"github.com/mitchellh/hashstructure/v2"
+	batcher2 "k8s.io/cloud-provider-aws/pkg/providers/v1/batcher"
```
cartermckinnon, line 23: +1

pkg/providers/v1/create_tags_batch.go
shiv-amz: We don't really need the alias, will push out a fix.

pkg/providers/v1/delete_tags_batch.go
shiv-amz: Same as above.

cartermckinnon:
/lgtm
/approve

@shiv-amz can you remove the usage of controller-runtime's logging framework in a followup? We should use the same logger used in the rest of the project

PR #1126, opened 2025-03-28
Enable concurrent workers for tagging controller

shiv-amz:
/kind feature

* Tagging controller currently operates as a single-threaded process, leading to a backlog when processing nodes in larger clusters. PR introduces ```tagging-controller-concurrent-node-syncs``` option, allowing multiple parallel workers to pull from the work queue.
* Testing with 1,000 nodes at a concurrency level of 10, over 50% of the nodes are tagged in under 1 second. In contrast, with single concurrency, 90% of the nodes take between 1 to 10 seconds to be tagged.

```release-note
A new command-line flag, `tagging-controller-concurrent-node-syncs`, allows more than one node to be processed at once.
```

### DIFF
--- a/pkg/controllers/options/tagging_controller.go
+++ b/pkg/controllers/options/tagging_controller.go
@@ -58,6 +58,10 @@ func (o *TaggingControllerOptions) Validate() error {
 		return fmt.Errorf("--tagging-controller-burst-limit should not be less than zero")
 	}
 
+	if o.WorkerCount <= 0 {
+		return fmt.Errorf("--tagging-controller-concurrent-node-syncs must be a positive number")
+	}
+
 	for _, r := range o.Resources {
 		for _, resource := range SupportedResources {
 			if r != resource {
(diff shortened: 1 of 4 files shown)


cartermckinnon:
CI account has hit a service quota:
```
   I0328 06:38:39.671578 87733 dump.go:53] At 2025-03-28 06:38:16 +0000 UTC - event for svc-proxy-terminating: {service-controller } SyncLoadBalancerFailed: Error syncing load balancer: failed to ensure load balancer: error authorizing security group ingress: "RulesPerSecurityGroupLimitExceeded: The maximum number of rules per security group has been reached.\n\tstatus code: 400, request id: 10e8ee61-8c6d-4f66-8845-0872eced910c" 
```

Seeing if we can get it resolved, if not we can override...

cartermckinnon:
Test flakes are not related to this change, going to continue looking into those out of band.

/lgtm
/approve

PR #1146, opened 2025-05-07
Upgrade AWS EC2 usage to Go SDK V2

gargipanatula:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

The AWS SDK Go V1 is being deprecated on July 31, 2025. This PR migrates usage of the EC2 SDK to use V2. 
Context: https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

Changes:
* CredentialsChainVerboseErrors has been removed from the EC2 Client (see https://github.com/aws/aws-sdk-go/issues/2914)
* In v1, `BeforeSign` caused some requests to be marked as non-retryable in some cases. Now, due to the V2's migration to middleware, the Finalize step (through `delayPreSign()`) now throws errors with a custom `nonRetryableError` error code when the same conditions are met. The EC2 Client is configured with a custom retrier, `customRetryer`, which will catch this error and cause the request to not be retried.
* Logging: AWS API Send has been split up into two logs due to changes in where the logged information is available throughout the middleware. Request parameters are only available in the Initialize and Serialize step.
Old logs:
```
newer

I0514 05:05:54.252687      11 log_handler.go:27] AWS request: ec2 CreateTags
  
I0514 05:05:54.452106      11 log_handler.go:32] AWS API Send: ec2 CreateTags &{CreateTags POST / <nil> <nil>} {
  Resources: ["i-09271f1185db6f280"],
  Tags: [{
      Key: "aws:eks:cluster-name",
      Value: "panatula513116"
    }]
}

I0514 05:05:54.452139      11 log_handler.go:37] AWS API ValidateResponse: ec2 CreateTags &{CreateTags POST / <nil> <nil>} {
  Resources: ["i-09271f1185db6f280"],
  Tags: [{
      Key: "aws:eks:cluster-name",
      Value: "panatula513116"
    }]
} 200 OK

older
```
New logs:
```
newer

AWS API Send: EC2 CreateTags &{[i-0adbe1d0ed4bb2329] [{0xc000a20e60 0xc000a20e70 {}}] <nil> {}}

AWS API Send: EC2 CreateTags POST /

AWS request: EC2 CreateTags

AWS API ValidateResponse: EC2 CreateTags 200

older
```

Fixes #

```release-note
NONE
```

Testing details:
Added unit tests to verify custom endpoint resolver and retrier in the EC2 Client, included in the `aws_sdk_test.go` file. 
Ran `make && make test`. Created a personal cluster and manually added nodes to verify that tags were still properly added. Ran e2e tests on the cluster using the following commands: 
```
export KUBECONFIG=<kubeconfig path>
export AWS_REGION=us-west-2
export TEST_PATH=./tests/e2e/...
export GINKGO_NODES=4
export GINKGO_FOCUS=[cloud-provider-aws-e2e]
export GINGKO_SKIP=[Disruptive]
go install github.com/onsi/ginkgo/ginkgo@v1.14.0
export PATH=$PATH:$HOME/go/bin
pushd ./tests/e2e
ginkgo . -v -p --nodes=$GINKGO_NODES --focus=$GINKGO_FOCUS --skip=$GINKGO_SKIP

<...test output...>
Ran 3 of 3 Specs in 72.299 seconds
SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestE2E (72.30s)
PASS

```


pkg/providers/v1/aws.go
```diff
@@ -3255,29 +3255,29 @@ func mapNodeNameToPrivateDNSName(nodeName types.NodeName) string {
...
 	return types.NodeName(aws.StringValue(i.PrivateDnsName))
 }
 
 var aliveFilter = []string{
-	ec2.InstanceStateNamePending,
-	ec2.InstanceStateNameRunning,
-	ec2.InstanceStateNameShuttingDown,
-	ec2.InstanceStateNameStopping,
-	ec2.InstanceStateNameStopped,
+	string(ec2types.InstanceStateNamePending),
```
yue9944882, line 3262: nit: change the array type `[]string{..}` to the new ec2 type and cast it to string when needed

pkg/providers/v1/aws_ec2.go
```diff
@@ -17,34 +17,57 @@ limitations under the License.
...
-	"github.com/aws/aws-sdk-go/service/ec2"
-	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 )
 
+// An interface to satisfy the ec2.Client API.
+// More details about this pattern: https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/unit-testing.html
+type EC2API interface {
```
yue9944882, line 31: why do we need this interface type? is it for testing purpose?

pkg/providers/v1/aws_sdk.go
```diff
@@ -223,3 +241,30 @@ func (p *awsSDKProvider) KeyManagement(regionName string) (KMS, error) {
 
 	return kmsClient, nil
 }
+
+func (p *awsSDKProvider) AddHandlersV2(ctx context.Context, regionName string, cfg awsv2.Config) {
```
yue9944882, line 245: can you add a code comment here to note that original implementation in SDK v1?

pkg/providers/v1/iface/types.go
```diff
@@ -1,38 +1,41 @@
...
+	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 )
 
 // EC2 is an abstraction over AWS', to allow mocking/other implementations
 // Note that the DescribeX functions return a list, so callers don't need to deal with paging
 // TODO: Should we rename this to AWS (EBS & ELB are not technically part of EC2)
 type EC2 interface {
 	// Query EC2 for instances matching the filter
-	DescribeInstances(request *ec2.DescribeInstancesInput) ([]*ec2.Instance, error)
+	DescribeInstances(ctx context.Context, request *ec2.DescribeInstancesInput, optFns ...func(*ec2.Options)) ([]ec2types.Instance, error)
```
yue9944882, line 15: why do we need `EC2API` interface above, if we already have this one.

pkg/providers/v1/log_handler.go
```diff
@@ -46,3 +52,45 @@ func awsServiceAndName(req *request.Request) (string, string) {
...
+	response, ok := out.RawResponse.(*http.Response)
+	if !ok {
+		return out, metadata, &smithy.DeserializationError{Err: fmt.Errorf("unknown transport type %T", out.RawResponse)}
+	}
+	service, name := awsServiceAndNameV2(ctx)
+	klog.V(4).Infof("AWS API ValidateResponse: %s %s %d", service, name, response.StatusCode)
+	return out, metadata, err
+}
+
+func awsServiceAndNameV2(ctx context.Context) (string, string) {
```
yue9944882, line 88: could you share the original implementation for this in v1?

pkg/providers/v1/aws_ec2.go
gargipanatula: Yes. Previously, `awsSdkEC2` and `MockedEC2API` wrapped `ec2iface.EC2API` to mock the ec2 client. In v2, the API doesn't exist, so we have to create our own interface and wrap that instead. 
More details: https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/unit-testing.html

pkg/providers/v1/iface/types.go
gargipanatula: `EC2API` is used to mock the EC2 client in general, whereas this interface is used to provide custom implementations over existing EC2 functionality (e.g. returning `[]*ec2.SecurityGroup` instead of the default `ec2.DescribeSecurityGroupsOutput` in `DescribeSecurityGroups()`). I assume this was originally created for mocking, because some operations return paginated results which are hard to mock. 

One example of how this is in [findSecurityGroup](https://github.com/gargipanatula/cloud-provider-aws/blob/aws-sdk-ec2-v2/pkg/providers/v1/aws.go#L1218). `Cloud.ec2` is `iface.EC2`, and the result of  `cloud.ec2.DescribeSecurityGroups` is `[]*ec2.SecurityGroup`.

Compare this to DescribeSecurityGroups, where `awsSdkEC2.ec2` is `EC2API`, and the result of `awsSdkEC2.ec2.DescribeSecurityGroups` is `*ec2.DescribeSecurityGroupsOutput`.

pkg/providers/v1/aws.go
gargipanatula: I agree that this would be better, but it's only used as a parameter to newEc2Filter, which expects a []string{}. Converting `aliveFilter` to strings just for that function might be a little messy, so maybe we can keep it as a `[]string{}`? Let me know if that sounds ok, open to either way.

pkg/providers/v1/aws_sdk.go
```diff
@@ -223,3 +241,30 @@ func (p *awsSDKProvider) KeyManagement(regionName string) (KMS, error) {
...
+	cfg.APIOptions = append(cfg.APIOptions,
```
kmala, line 248: is there a way to test these in unit test? if not can you share the log details just to make sure we the behavior is same?

pkg/providers/v1/aws_sdk.go
```diff
@@ -223,3 +241,30 @@ func (p *awsSDKProvider) KeyManagement(regionName string) (KMS, error) {
...
+		middleware.AddUserAgentKeyValue("kubernetes", version.Get().String()),
+		func(stack *smithymiddleware.Stack) error {
+			return stack.Finalize.Add(&awsHandlerLoggerV2{}, smithymiddleware.Before)
+		},
+	)
+
+	delayer := p.getCrossRequestRetryDelay(regionName)
+	if delayer != nil {
+		cfg.APIOptions = append(cfg.APIOptions,
```
kmala, line 257: can we test this using an unit test?

pkg/providers/v1/retry_handler.go
```diff
@@ -173,3 +181,102 @@ func (b *Backoff) ReportError() {
...
+
+// Standard retry implementation, except that it doesn't retry RequestLimitExceeded errors.
+// This works in tandem with CrossRequestRetryDelay, which reports these errors before
+// the middleware exits upon seeing this error
+type customRetryer struct {
+	awsv2.Retryer
+}
+func (r customRetryer) IsErrorRetryable(err error) bool {
+	var ae smithy.APIError
+	if errors.As(err, &ae) && strings.Contains(ae.Error(), NON_RETRYABLE_ERROR) {
```
kmala, line 193: i see you are creating this error as a normal one using errors.new,  then will errors.As succeed here ?

pkg/providers/v1/iface/types.go
haoranleo: I don't get it, the detailed implementation for mock can be different but shouldn't the interface be the same? We can only keep this interface and reuse it for mocking

pkg/providers/v1/retry_handler.go
```diff
@@ -173,3 +181,102 @@ func (b *Backoff) ReportError() {
...
+		return false
+	}
+	return r.Retryer.IsErrorRetryable(err)
+}
```
haoranleo, line 197: Why do we need this? I didn't see it used anywhere. 

v2 retryer itself has an implementation to decide which errors are retryable https://github.com/aws/aws-sdk-go-v2/blob/main/aws/retry/retryable_error.go#L15. But I didn't see the purpose of custom retryer here

pkg/providers/v1/iface/types.go
gargipanatula: I think it's because the original authors wanted two levels of mocking. The first one is that they wanted to mock the EC2 Client functions, hence their usage of ec2iface.EC2API (now EC2API). The second one is that they had their own set of functions, and to mock these, they made iface.EC2. These functions are not really related to the EC2 Client, they're essentially functions that happen to call the EC2 API and do their own thing. 

I believe that if we take out iface.EC2, then we don't have a way to mock things like [awsSdkEC2.DescribeSecurityGroups](https://github.com/kubernetes/cloud-provider-aws/blob/9aab0da979074ac2f65123f7c30f208a8774aec3/pkg/providers/v1/aws_ec2.go#L78)

pkg/controllers/tagging/tagging_controller.go
```diff
@@ -201,22 +201,22 @@ func (tc *Controller) Run(stopCh <-chan struct{}) {
 
 	klog.Infof("Starting the tagging controller")
 	for i := 0; i < tc.workerCount; i++ {
-		go wait.Until(tc.work, tc.nodeMonitorPeriod, stopCh)
+		go wait.Until(func() { tc.work(ctx) }, tc.nodeMonitorPeriod, stopCh)
```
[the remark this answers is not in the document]
gargipanatula: If we don't wrap, it interprets `tc.work(ctx)` as a function call, so I wrapped it to satisfy the fact that .Until needs a `func()`

pkg/providers/v1/aws.go
```diff
@@ -898,14 +900,14 @@ func (c *Cloud) InstanceExistsByProviderID(ctx context.Context, providerID strin
 	}
 
 	if v := variant.GetVariant(string(instanceID)); v != nil {
-		return v.InstanceExists(string(instanceID), c.vpcID)
+		return v.InstanceExists(ctx, string(instanceID), c.vpcID)
 	}
 
 	request := &ec2.DescribeInstancesInput{
-		InstanceIds: []*string{instanceID.awsString()},
+		InstanceIds: []string{},
```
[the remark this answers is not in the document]
gargipanatula: My bad - fixed.

pkg/providers/v1/retry_handler.go
gargipanatula: Ah you're right - errors.As would be false for NON_RETRYABLE_ERRORs. Will do a simple error message check instead.

pkg/providers/v1/aws_sdk.go
```diff
@@ -112,27 +131,26 @@ func (p *awsSDKProvider) getCrossRequestRetryDelay(regionName string) *CrossRequ
...
+	cfg.Region = regionName
+	cfg.Credentials = p.credsV2
+
+	p.AddHandlersV2(ctx, regionName, cfg)
 
-	p.AddHandlers(regionName, &service.Handlers)
+	ec2Client := ec2.NewFromConfig(cfg, func(o *ec2.Options) {
+		o.Retryer = &customRetryer{
+			retry.NewStandard(),
+		}
```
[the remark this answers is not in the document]
gargipanatula: The reason we need to have a custom retryer is because we need to mark requests as non retryable under specific conditions. In v2, we do this via a combination of middleware and custom retry logic (see here: https://github.com/kubernetes/cloud-provider-aws/pull/1146/files#r2082328615). 

The custom retryer you shared would definitely limit the number of retries, but it would do it for all requests. In this case, we need to be able to catch a specific case and never retry that.

pkg/providers/v1/aws_ec2.go
gargipanatula: Yeah that makes sense. All these guys are mocked though, mostly by `FakeEC2Impl`, so maybe we can keep it as is for now to minimize changes in this PR and address it in the future.

pkg/providers/v1/aws_sdk.go
```diff
@@ -112,27 +131,26 @@ func (p *awsSDKProvider) getCrossRequestRetryDelay(regionName string) *CrossRequ
 	return delayer
 }
 
-func (p *awsSDKProvider) Compute(regionName string) (iface.EC2, error) {
-	awsConfig := &aws.Config{
-		Region:      &regionName,
-		Credentials: p.creds,
-	}
-	awsConfig = awsConfig.WithCredentialsChainVerboseErrors(true).
-		WithEndpointResolver(p.cfg.GetResolver())
```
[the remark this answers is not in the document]
gargipanatula: Yep, but interestingly the custom resolver written there is only really used for testing, which is when the ServiceOverride field gets set. In normal cases, the function just relies on the default endpoint resolution logic of a client (which is already handled for us in V2).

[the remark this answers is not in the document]
gargipanatula: Thank you - added client level overrides!

pkg/providers/v1/aws_sdk.go
gargipanatula: Yep will paste the logs in the PR description, thank you.

pkg/providers/v1/aws_sdk.go
gargipanatula: Yep will paste the logs in the PR description, thank you.

pkg/providers/v1/aws_sdk.go
gargipanatula: For clarification, do we want to add a test for `getCrossRequestRetryDelay` or for the middleware functions themselves? The functions called in middleware, like `ComputeDelayForRequest` are tested in separate unit tests.

pkg/providers/v1/log_handler.go
```diff
@@ -46,3 +52,65 @@ func awsServiceAndName(req *request.Request) (string, string) {
...
+		func(ctx context.Context, in middleware.DeserializeInput, next middleware.DeserializeHandler) (
+			out middleware.DeserializeOutput, metadata middleware.Metadata, err error,
+		) {
+			out, metadata, err = next.HandleDeserialize(ctx, in)
+			response, ok := out.RawResponse.(*http.Response)
+			if !ok {
+				return out, metadata, &smithy.DeserializationError{Err: fmt.Errorf("unknown transport type %T", out.RawResponse)}
+			}
+			service, name := awsServiceAndNameV2(ctx)
+			klog.V(4).Infof("AWS API ValidateResponse: %s %s %d", service, name, response.StatusCode)
```
kmala, line 85: can you also print the in.Parameters so that we can correlate for which request this response is which will be useful if there are a lot of requests happening in parallel?

pkg/providers/v1/aws_sdk.go
```diff
@@ -56,12 +64,12 @@ func (p *awsSDKProvider) AddHandlers(regionName string, h *request.Handlers) {
 	h.Build.PushFrontNamed(request.NamedHandler{
 		Name: "k8s/user-agent",
 		Fn:   request.MakeAddToUserAgentHandler("kubernetes", version.Get().String()),
-	})
+	}) // done
```
kmala, line 67: nit: is this for your testing/verfification?

pkg/providers/v1/aws_sdk.go
```diff
@@ -88,7 +96,42 @@ func (p *awsSDKProvider) addAPILoggingHandlers(h *request.Handlers) {
...
+}
+
+// Adds handlers to AWS SDK Go V2 clients. For AWS SDK Go V1 clients,
+// func (p *awsSDKProvider) AddHandlers is used.
+func (p *awsSDKProvider) AddHandlersV2(ctx context.Context, regionName string, cfg *awsv2.Config) {
+	cfg.APIOptions = append(cfg.APIOptions,
+		middleware.AddUserAgentKeyValue("kubernetes", version.Get().String()),
+		func(stack *smithymiddleware.Stack) error {
+			return stack.Finalize.Add(awsHandlerLoggerMiddleware(), smithymiddleware.Before)
```
kmala, line 108: i think this should be initialize or serialize instead of finalize such that it at the start of the request ?

pkg/providers/v1/log_handler.go
gargipanatula: Unfortunately the parameters are only available in the Initialize and Serialize phase ([pkg](https://pkg.go.dev/github.com/aws/smithy-go/middleware#InitializeInput)), and ValidateResponse maps to the Deserialize phase, so we can't really keep v1's "AWS API ValidateResponse" log output. In v2, we are now logging the parameters at the Serialize phase through `awsSendHandlerLoggerMiddleware`.

pkg/providers/v1/aws_sdk.go
gargipanatula: This logger was initially added to the Sign handler, which maps to the Finalize phase in the middleware, so that's why Finalize is used here. However, if it's better in another phase, I can move it around.

pkg/providers/v1/config/config.go
```diff
@@ -205,7 +208,24 @@ func (cfg *CloudConfig) GetResolver() endpoints.ResolverFunc {
...
 }
 
+// GetEC2Endpoint returns client configuration options that override
+// the signing name and region, if appropriate. Replicates logic
+// from GetResolver() for AWS SDK Go V2 clients.
+func (cfg *CloudConfig) GetEC2Endpoint(region string) []func(*ec2.Options) {
+	opts := []func(*ec2.Options){}
+	for _, override := range cfg.ServiceOverride {
+		if override.Service == ec2.ServiceID && override.Region == region {
+			opts = append(opts,
```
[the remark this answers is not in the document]
gargipanatula: Ah yes thanks for the catch, added a separate endpoint resolver that's added in Compute() that does so.

pkg/providers/v1/config/config.go
```diff
@@ -205,7 +211,58 @@ func (cfg *CloudConfig) GetResolver() endpoints.ResolverFunc {
...
+func (r *EC2Resolver) ResolveEndpoint(
+	ctx context.Context, params ec2.EndpointParameters,
+) (
+	endpoint smithyendpoints.Endpoint, err error,
+) {
+	for _, override := range r.Cfg.ServiceOverride {
+		if override.Service == ec2.ServiceID && override.Region == aws.ToString(params.Region) {
+			customURL, err := url.Parse(override.URL)
+			if err != nil {
+				return r.Resolver.ResolveEndpoint(ctx, params)
```
kmala, line 254: this should be an error as its a config issue.

pkg/providers/v1/aws_sdk_test.go
```diff
@@ -0,0 +1,163 @@
...
+	"k8s.io/cloud-provider-aws/pkg/providers/v1/config"
+	"k8s.io/cloud-provider-aws/pkg/providers/v1/iface"
+)
+
+type MockAWSSDKProvider struct {
+	p *awsSDKProvider
+}
+
+// Mocks endpoint resolution and custom retries in the EC2 Client.
+func (m *MockAWSSDKProvider) Compute(ctx context.Context, regionName string, mockRT *MockRoundTripper) (iface.EC2, error) {
```
kmala, line 24: from what i see this function is duplicating the code of the actual sdk provider compute for testing purposes but this means that if we are modifying the actual function we also need to update this and its easy to miss it code reviews. Can we not use the actual function with cloud config overrides?

pkg/providers/v1/config/config.go
```diff
@@ -205,7 +211,58 @@ func (cfg *CloudConfig) GetResolver() endpoints.ResolverFunc {
...
+
+// EC2Resolver overrides the endpoint for an AWS SDK Go V2 EC2 Client,
+// using the provided CloudConfig to determine if an override
+// is appropriate.
+type EC2Resolver struct {
+	Resolver ec2.EndpointResolverV2
+	Cfg      *CloudConfig
+}
+
+func (r *EC2Resolver) ResolveEndpoint(
```
kmala, line 245: can this be a generic resolver as use for other services also by having the service name in struct?

gargipanatula: This would definitely be ideal, but I don't believe that there's a type that could satisfy resolvers across multiple services, since each service defines its own `EndpointResolverV2` and `EndpointParameters`.

gargipanatula:
`/retest`

pkg/providers/v1/aws_ec2.go
yue9944882: nit: rename it to EC2APIForTesting?

yue9944882 (approved):

PR #1154, opened 2025-06-02
fix: reduce log verbosity of tagging controller

cartermckinnon:
**What type of PR is this?**

/kind feature

The tagging controller log's are quite chatty, and this can cause issues in large clusters.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/tagging/tagging_controller.go
+++ b/pkg/controllers/tagging/tagging_controller.go
@@ -165,7 +165,7 @@ func NewTaggingController(
 			// and when it gets tagged, there might be another event which put the same item in the work queue
 			// (since the node won't have the labels yet) and hence prevents us from making an unnecessary EC2 call.
 			if !tc.isTaggingRequired(node) {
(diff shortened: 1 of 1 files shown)


pkg/controllers/tagging/tagging_controller.go
```diff
@@ -165,7 +165,7 @@ func NewTaggingController(
 			// and when it gets tagged, there might be another event which put the same item in the work queue
 			// (since the node won't have the labels yet) and hence prevents us from making an unnecessary EC2 call.
 			if !tc.isTaggingRequired(node) {
-				klog.Infof("Skip putting node %s in work queue since it was already tagged earlier.", node.GetName())
+				klog.V(3).Infof("Skip putting node %s in work queue since it was already tagged earlier.", node.GetName())
```
cartermckinnon, line 168: This will be logged for every Node, every time its status is updated by kubelet (every 5 minutes by default). With many nodes, this creates a ton of noise in the CCM logs.

PR #1155, opened 2025-06-02
[WIP] Support ServiceAccountToken in ecr-credential-provider

### LINKED ISSUE #1147: Add support for service account token in ece-credential-provider

Kubernetes’ alpha feature `KubeletServiceAccountTokenForCredentialProviders` allows the kubelet to inject a projected ServiceAccount token into the credential-provider plugin’s request. By leveraging this token, on-premise or self-hosted clusters can avoid baking static cloud credentials into each node’s file system and instead perform OIDC-based authentication at runtime.

**Why is this needed**:

Using projected ServiceAccount tokens lets on-prem and air-gapped clusters avoid baking long-lived cloud secrets into every node. Instead, nodes request short-lived, scoped credentials at pull time via OIDC, improving security and automating rotation without impacting existing workflows.

/kind feature

fletcherw:
Extend ecr-credential-provider to support fine-grained access via kubernetes ServiceAccount tokens and STS's AssumeRoleWithWebIdentity.

This allows users to avoid long-lived secrets in their pods/nodes and instead use a short-lived credential generated by kubernetes in order to access private ECR images.

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This PR

Fixes #1147 

```release-note

This PR adds support for using ServiceAccount tokens to authenticate with AWS when pulling images from private ECR registries. (See https://github.com/kubernetes/enhancements/issues/4412)
Users who wish to use this new feature will need to opt in to the feature flag and configure their kubernetes cluster and AWS resources appropriately.

```

### DIFF
--- a/cmd/ecr-credential-provider/main.go
+++ b/cmd/ecr-credential-provider/main.go
@@ -31,6 +31,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/ecr"
 	"github.com/aws/aws-sdk-go-v2/service/ecrpublic"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/spf13/cobra"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -54,9 +55,15 @@ type ECRPublic interface {
 	GetAuthorizationToken(ctx context.Context, params *ecrpublic.GetAuthorizationTokenInput, optFns ...func(*ecrpublic.Options)) (*ecrpublic.GetAuthorizationTokenOutput, error)
 }
 
+// For abstracting in tests.
+type STS interface {
+	AssumeRoleWithWebIdentity(context.Context, *sts.AssumeRoleWithWebIdentityInput, ...func(*sts.Options)) (*sts.AssumeRoleWithWebIdentityOutput, error)
+}
+
 type ecrPlugin struct {
 	ecr       ECR
 	ecrPublic ECRPublic
+	sts       STS
 }
 
 func defaultECRProvider(ctx context.Context, region string) (ECR, error) {
@@ -89,6 +96,24 @@ func publicECRProvider(ctx context.Context) (ECRPublic, error) {
 	}
 
 	return ecrpublic.NewFromConfig(cfg), nil
+}
+
+func stsProvider(ctx context.Context, region string) (STS, error) {
+	var cfg aws.Config
+	var err error
+	if region != "" {
+		cfg, err = config.LoadDefaultConfig(ctx,
+			config.WithRegion(region),
+		)
+	} else {
+		klog.Warningf("No region found in the image reference, the default region will be used. Please refer to AWS SDK documentation for configuration purpose.")
+		cfg, err = config.LoadDefaultConfig(ctx)
+	}
+
+	if err != nil {
+		return nil, err
+	}
+	return sts.NewFromConfig(cfg), nil
 }
 
 type credsData struct {
@@ -126,7 +151,7 @@ func (e *ecrPlugin) getPublicCredsData(ctx context.Context) (*credsData, error) 
 	}, nil
 }
 
-func (e *ecrPlugin) getPrivateCredsData(ctx context.Context, imageHost string, image string) (*credsData, error) {
+func (e *ecrPlugin) getPrivateCredsData(ctx context.Context, imageHost string, image string, optFn func(*ecr.Options)) (*credsData, error) {
 	klog.Infof("Getting creds for private image %s", image)
 	var err error
 
@@ -137,7 +162,7 @@ func (e *ecrPlugin) getPrivateCredsData(ctx context.Context, imageHost string, i
 			return nil, err
 		}
 	}
-	output, err := e.ecr.GetAuthorizationToken(ctx, &ecr.GetAuthorizationTokenInput{})
+	output, err := e.ecr.GetAuthorizationToken(ctx, &ecr.GetAuthorizationTokenInput{}, optFn)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


cmd/ecr-credential-provider/main.go
```diff
@@ -153,19 +178,66 @@ func (e *ecrPlugin) getPrivateCredsData(ctx context.Context, imageHost string, i
...
+	}
+
+	imageHost, err := parseHostFromImageReference(request.Image)
 	if err != nil {
 		return nil, err
 	}
 
 	if imageHost == ecrPublicHost {
+		if request.ServiceAccountToken != "" {
+			klog.Warningf("Ignoring provided service account token for public ECR")
```
cartermckinnon: why's that?

cmd/ecr-credential-provider/main.go
```diff
@@ -153,19 +178,66 @@ func (e *ecrPlugin) getPrivateCredsData(ctx context.Context, imageHost string, i
...
+				}, nil
+			})
+
+		}
+
+		creds, err = e.getPrivateCredsData(ctx, imageHost, request.Image, func(options *ecr.Options) {
+			if credentialProvider != nil {
+				options.Credentials = credentialProvider
+			}
+		})
```
cartermckinnon: maybe cleaner if we declared a `var opts []func(*ecr.Options)` at the top and appended the credentialProvider opt within the `if request.ServiceAccountToken != ""` block?

cmd/ecr-credential-provider/main.go
```diff
@@ -153,19 +178,66 @@ func (e *ecrPlugin) getPrivateCredsData(ctx context.Context, imageHost string, i
...
+
+			arn, ok := request.ServiceAccountAnnotations["eks.amazonaws.com/ecr-role-arn"]
+			if !ok {
+				arn = os.Getenv("AWS_ECR_ROLE_ARN")
+			}
+			if arn == "" {
+				return nil, errors.New("no arn provided, cannot assume role using ServiceAccountToken")
+			}
+
+			credentialProvider = aws.CredentialsProviderFunc(func(ctx context.Context) (aws.Credentials, error) {
```
cartermckinnon: I don't think this cast is needed?

cmd/ecr-credential-provider/main.go
```diff
@@ -91,6 +98,24 @@ func publicECRProvider(ctx context.Context) (ECRPublic, error) {
...
+		)
+	} else {
+		klog.Warningf("No region found in the image reference, the default region will be used. Please refer to AWS SDK documentation for configuration purpose.")
+		cfg, err = config.LoadDefaultConfig(ctx)
+	}
+
+	if err != nil {
+		return nil, err
+	}
+	return sts.NewFromConfig(cfg), nil
```
cartermckinnon, line 116: can you check if this will use STS regional endpoints by default? I think the go sdk v2 does

tests/e2e/ecr_creds.go
```diff
@@ -0,0 +1,61 @@
...
+							},
+							RunAsNonRoot: &[]bool{true}[0],
+							RunAsUser:    &[]int64{1000}[0],
+							RunAsGroup:   &[]int64{1000}[0],
+							SeccompProfile: &v1.SeccompProfile{
+								Type: v1.SeccompProfileTypeRuntimeDefault,
+							},
+						},
+						Name:  "test",
+						Image: "public.ecr.aws/nginx/nginx-unprivileged:latest",
```
cartermckinnon: maybe could we use something like the k8s pause image for this? We have a regional ECR repo (not public) that are doc'd here: https://docs.aws.amazon.com/eks/latest/userguide/add-ons-images.html

which should give us a deeper validation, since we need the cred provider to do its thing for that to work.

IIRC the k8s CI cluster is in us-east-1, so we could try:
```
602401143452.dkr.ecr.us-east-1.amazonaws.com/eks/pause:3.5
```

cmd/ecr-credential-provider/main.go
fletcherw: I wasn't aware of reasons to do this before. I've updated it to support using the token for public ecr as well.

cmd/ecr-credential-provider/main.go
fletcherw: The CredentialProvider interface is
```
type CredentialsProvider interface {
	// Retrieve returns nil if it successfully retrieved the value.
	// Error is returned if the value were not obtainable, or empty.
	Retrieve(ctx context.Context) (Credentials, error)
}
```

so the anonymous function can't be substituted

cmd/ecr-credential-provider/main.go
fletcherw: good idea, done!

cmd/ecr-credential-provider/main.go
fletcherw: It looks like it does by default: https://github.com/aws/aws-sdk-go-v2/blob/main/service/sts/endpoints.go#L266-L273

tests/e2e/ecr_creds.go
fletcherw: good idea, will give this a try and see what happens.

cmd/ecr-credential-provider/main.go
```diff
@@ -153,19 +179,83 @@ func (e *ecrPlugin) getPrivateCredsData(ctx context.Context, imageHost string, i
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	if imageHost == ecrPublicHost {
-		creds, err = e.getPublicCredsData(ctx)
+		var optFns = []func(*ecrpublic.Options){}
+		if credentialsProvider != nil {
+			optFns = append(optFns, func(o *ecrpublic.Options) {
```
cartermckinnon, line 246: argh, I didn't notice earlier that these opt funcs take different Options types. That's annoying :( I thought we could pass the same []opts to both ecr and ecrpublic

tests/e2e/ecr_creds.go
```diff
@@ -0,0 +1,61 @@
...
+							SeccompProfile: &v1.SeccompProfile{
+								Type: v1.SeccompProfileTypeRuntimeDefault,
+							},
+						},
+						Name:  "test",
+						Image: "602401143452.dkr.ecr.us-east-1.amazonaws.com/eks/pause:3.5",
+					},
+				},
+			},
+		})
```
cartermckinnon, line 59: I think we need to wait for the pod to become ready to verify the image was pulled, looks like this will do the trick: https://pkg.go.dev/k8s.io/kubernetes/test/e2e/framework/pod#WaitTimeoutForPodReadyInNamespace

cmd/ecr-credential-provider/main.go
cartermckinnon: ah gotcha, thanks!

cartermckinnon:
This is great, thanks @fletcherw!

/lgtm
/approve

cmd/ecr-credential-provider/main.go
fletcherw: yeah I was bummed as well, that would have been a lot cleaner.

PR #1157, opened 2025-06-05
[DRAFT] AWS SDK Go V2 Upgrade of ELB and ELBv2

gargipanatula:
Upgrading the ELB and ELBv2 packages of the AWS SDK Go V2.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -31,6 +31,8 @@ import (
 	stscredsv2 "github.com/aws/aws-sdk-go-v2/credentials/stscreds"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	elb "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancing"
+	elbtypes "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancing/types"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
@@ -38,7 +40,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/elb"
 	"github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/aws/aws-sdk-go/service/sts"
@@ -291,7 +292,7 @@ const MaxReadThenCreateRetries = 30
 // Services is an abstraction over AWS, to allow mocking/other implementations
 type Services interface {
 	Compute(ctx context.Context, region string, assumeRoleProvider *stscredsv2.AssumeRoleProvider) (iface.EC2, error)
-	LoadBalancing(region string) (ELB, error)
+	LoadBalancing(ctx context.Context, regionName string, assumeRoleProvider *stscredsv2.AssumeRoleProvider) (ELB, error)
 	LoadBalancingV2(region string) (ELBV2, error)
 	Metadata() (config.EC2Metadata, error)
 	KeyManagement(region string) (KMS, error)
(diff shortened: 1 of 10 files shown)


pkg/providers/v1/aws.go
```diff
@@ -291,69 +293,69 @@ const MaxReadThenCreateRetries = 30
...
-	DescribeListeners(*elbv2.DescribeListenersInput) (*elbv2.DescribeListenersOutput, error)
-	DeleteListener(*elbv2.DeleteListenerInput) (*elbv2.DeleteListenerOutput, error)
-	ModifyListener(*elbv2.ModifyListenerInput) (*elbv2.ModifyListenerOutput, error)
+	CreateListener(ctx context.Context, input *elbv2.CreateListenerInput, optFns ...func(*elbv2.Options)) (*elbv2.CreateListenerOutput, error)
+	DescribeListeners(ctx context.Context, input *elbv2.DescribeListenersInput, optFns ...func(*elbv2.Options)) (*elbv2.DescribeListenersOutput, error)
+	DeleteListener(ctx context.Context, input *elbv2.DeleteListenerInput, optFns ...func(*elbv2.Options)) (*elbv2.DeleteListenerOutput, error)
+	ModifyListener(ctx context.Context, input *elbv2.ModifyListenerInput, optFns ...func(*elbv2.Options)) (*elbv2.ModifyListenerOutput, error)
 
-	WaitUntilLoadBalancersDeleted(*elbv2.DescribeLoadBalancersInput) error
+	// WaitUntilLoadBalancersDeleted(ctx context.Context, input *elbv2.DescribeLoadBalancersInput, optFns ...func(*elbv2.WaiterOptions)) error
```
kmala: any reason for commenting it?

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -653,20 +656,20 @@ func (c *Cloud) ensureTargetGroup(targetGroup *elbv2.TargetGroup, serviceName ty
...
 		}
-		if mapping.HealthCheckConfig.HealthyThreshold != aws.Int64Value(targetGroup.HealthyThresholdCount) {
+		if mapping.HealthCheckConfig.HealthyThreshold != aws.Int32Value(targetGroup.HealthyThresholdCount) {
 			dirtyHealthCheck = true
-			input.HealthyThresholdCount = aws.Int64(mapping.HealthCheckConfig.HealthyThreshold)
-			input.UnhealthyThresholdCount = aws.Int64(mapping.HealthCheckConfig.UnhealthyThreshold)
+			input.HealthyThresholdCount = aws.Int32(mapping.HealthCheckConfig.HealthyThreshold)
+			input.UnhealthyThresholdCount = aws.Int32(mapping.HealthCheckConfig.UnhealthyThreshold)
 		}
-		if !strings.EqualFold(mapping.HealthCheckConfig.Protocol, elbv2.ProtocolEnumTcp) {
```
yue9944882, line 661: looks like this string compare change is non-trivial. do we know why we need `EqualFold` in v1?

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -1664,3 +1659,47 @@ func filterTargetNodes(nodes []*v1.Node, annotations map[string]string) []*v1.No
 
 	return targetNodes
 }
+
+// ValidateHealthCheck replaces ELB.HealthCheck.Validate() from AWS SDK Go V1, which has been deprecated in V2
+func ValidateHealthCheck(s *elbtypes.HealthCheck) error {
```
yue9944882, line 1665: can we link the original v1 implementation either in the code comment or here?

pkg/providers/v1/aws_sdk.go
```diff
@@ -186,45 +187,56 @@ func (p *awsSDKProvider) Compute(ctx context.Context, regionName string, assumeR
...
-	})
 	if err != nil {
-		return nil, fmt.Errorf("unable to initialize AWS session: %v", err)
+		return nil, fmt.Errorf("unable to initialize AWS config: %v", err)
 	}
-	elbClient := elb.New(sess)
-	p.AddHandlers(regionName, &elbClient.Handlers)
+
+	p.AddHandlersV2(ctx, regionName, &cfg)
+	var opts []func(*elb.Options) = p.cfg.GetELBEndpointOpts(regionName)
```
yue9944882, line 201: nit: opts := []func(*elb.Options) {... <put elements here>...} instead of appending one-by-one. each append is a malloc op under the hood

pkg/providers/v1/aws.go
```diff
@@ -1124,71 +1130,73 @@ func (c *Cloud) buildSelfAWSInstance(ctx context.Context) (*awsInstance, error)
...
 		tags = append(tags, tag)
 	}
 
 	request := &elb.AddTagsInput{}
-	request.LoadBalancerNames = []*string{&loadBalancerName}
+	request.LoadBalancerNames = []string{loadBalancerName}
 	request.Tags = tags
 
-	_, err := c.elb.AddTags(request)
+	_, err := c.elb.AddTags(context.Background(), request)
```
kmala: we should be passing the context from the parent function since there controller can cancel a request based on some conditions. The k8s CCM interface has ctx for all the functions so we should be albe to use them ex : https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws.go#L2166

kmala: can we do this for other places also where `context.Background()` is used and is a downstream function of the main controller function.

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -277,12 +280,12 @@ func (c *Cloud) ensureLoadBalancerv2(namespacedName types.NamespacedName, loadBa
 					targetGroupRecreated := false
 					targetGroup, ok := nodePortTargetGroup[nodePort]
 
-					if targetGroup != nil && (!strings.EqualFold(mapping.HealthCheckConfig.Protocol, aws.StringValue(targetGroup.HealthCheckProtocol)) ||
-						mapping.HealthCheckConfig.Interval != aws.Int64Value(targetGroup.HealthCheckIntervalSeconds)) {
+					if targetGroup != nil && ((mapping.HealthCheckConfig.Protocol != targetGroup.HealthCheckProtocol) ||
```
kmala: EqualFold is not case sensitive, should we still keep it?

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -709,36 +712,36 @@ func (c *Cloud) ensureTargetGroupTargets(tgARN string, expectedTargets []*elbv2.
...
-	resp, err := c.elbv2.DescribeTargetHealth(req)
+	resp, err := c.elbv2.DescribeTargetHealth(context.Background(), req)
 	if err != nil {
 		return nil, fmt.Errorf("error describing target group health: %q", err)
 	}
-	actualTargets := make([]*elbv2.TargetDescription, 0, len(resp.TargetHealthDescriptions))
+	actualTargets := make([]*elbv2types.TargetDescription, 0, len(resp.TargetHealthDescriptions))
 	for _, targetDesc := range resp.TargetHealthDescriptions {
-		if targetDesc.TargetHealth.Reason != nil && aws.StringValue(targetDesc.TargetHealth.Reason) == elbv2.TargetHealthReasonEnumTargetDeregistrationInProgress {
+		if string(targetDesc.TargetHealth.Reason) != "" && targetDesc.TargetHealth.Reason == elbv2types.TargetHealthReasonEnumDeregistrationInProgress {
```
kmala: nit: its not a pointer right, can we just skip the `!= ""` condition

pkg/providers/v1/aws_loadbalancer.go
gargipanatula: The values we're comparing here are now both enums, whereas before they were strings. AFAIK this removes the need to worry about casing, hence why `EqualFold` was removed.

More enums are supported by V2, so that's why I changed structs to use enums wherever possible.

pkg/providers/v1/aws.go
gargipanatula: Nope - forgot to remove it, thank you. Removing as its mock is not used anywhere in the code, so it doesn't need to be a part of the interface.

pkg/providers/v1/aws_loadbalancer.go
gargipanatula: I took it out because the values we're comparing are now both enums of type `elb.ProtocolEnum`, whereas before they were strings. AFAIK we don't need to worry about casing anymore but if there is an edge case I should be aware of then I can put it back in.

pkg/providers/v1/aws_sdk.go
gargipanatula: Oh I see, do you mean something like this:
```
var opts []func(*ec2.Options) = p.cfg.GetEC2EndpointOpts(regionName)
	opts = append(opts, func(o *ec2.Options) {
		o.Retryer = &customRetryer{
			retry.NewStandard(),
		}
		o.EndpointResolverV2 = p.cfg.GetCustomEC2Resolver()
	})
```
Couldn't figure out how spread p.cfg.GetELBEndpointOpts(regionName) into a composite literal ([]func(*elb.Options){ ... }), let me know if i'm missing something

pkg/providers/v1/aws_loadbalancer.go
kmala: ProtocolEnum is just a [string type](https://github.com/aws/aws-sdk-go-v2/blob/main/service/elasticloadbalancingv2/types/enums.go#L300) and doesn't mandate specific values, since we are converting the string to ProtocolEnum for `mapping.HealthCheckConfig.Protocol` i think its possible for it value to be just `http` and hence would fail the check

gargipanatula: I see, ok in that case I'll add the `EqualFold` back in. Thanks for the catch!

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -747,16 +750,16 @@ func (c *Cloud) obtainTargetGroupActualTargets(tgARN string) ([]*elbv2.TargetDes
 }
 
 // diffTargetGroupTargets computes the targets to register and targets to deregister based on existingTargets and desired instances.
-func (c *Cloud) diffTargetGroupTargets(expectedTargets []*elbv2.TargetDescription, actualTargets []*elbv2.TargetDescription) (targetsToRegister []*elbv2.TargetDescription, targetsToDeregister []*elbv2.TargetDescription) {
-	expectedTargetsByUID := make(map[string]*elbv2.TargetDescription, len(expectedTargets))
+func (c *Cloud) diffTargetGroupTargets(expectedTargets []*elbv2types.TargetDescription, actualTargets []*elbv2types.TargetDescription) (targetsToRegister []elbv2types.TargetDescription, targetsToDeregister []elbv2types.TargetDescription) {
+	expectedTargetsByUID := make(map[string]elbv2types.TargetDescription, len(expectedTargets))
```
haoranleo, line 754: Why we keep the function parameter as pointer type but change the return value to value type?

pkg/providers/v1/aws_test.go
```diff
@@ -2719,46 +2721,46 @@ func (m *MockedFakeELBV2) DescribeTargetGroups(request *elbv2.DescribeTargetGrou
 	}, nil
 }
 
-func (m *MockedFakeELBV2) ModifyTargetGroup(request *elbv2.ModifyTargetGroupInput) (*elbv2.ModifyTargetGroupOutput, error) {
-	var matchingTargetGroup *elbv2.TargetGroup
-	dirtyGroups := []*elbv2.TargetGroup{}
+func (m *MockedFakeELBV2) ModifyTargetGroup(ctx context.Context, input *elbv2.ModifyTargetGroupInput, optFns ...func(*elbv2.Options)) (*elbv2.ModifyTargetGroupOutput, error) {
+	var matchingTargetGroup *elbv2types.TargetGroup
+	dirtyGroups := []elbv2types.TargetGroup{}
```
haoranleo, line 2726: nit: maybe we don't need pointer value for matchingTargetGroup as well

pkg/providers/v1/aws_test.go
```diff
@@ -3615,7 +3615,12 @@ func TestCloud_buildNLBHealthCheckConfiguration(t *testing.T) {
 
 			hc, err := c.buildNLBHealthCheckConfiguration(tt.service)
 			if !tt.wantError {
-				assert.Equal(t, tt.want, hc)
+				same := assert.Equal(t, tt.want, hc)
+				if !same {
+					fmt.Println("wanted: ", tt.want)
+					fmt.Println("got: ", hc)
+
+				}
```
haoranleo: these should already be done in `assert.Equal`

pkg/providers/v1/config/config.go
```diff
@@ -263,6 +265,110 @@ func (r *EC2Resolver) ResolveEndpoint(
...
+	}
+	return opts
+}
+
+// GetCustomELBResolver returns an endpoint resolver for ELB Clients
+func (cfg *CloudConfig) GetCustomELBResolver() elb.EndpointResolverV2 {
+	return &ELBResolver{
+		Resolver: elb.NewDefaultEndpointResolverV2(),
+		Cfg:      cfg,
+	}
```
haoranleo, line 289: Is it intentionally using `NewDefaultEndpointResolverV2` for v1 resolver?

haoranleo: The `ELBResolver` seems identical to `ELBV2Resolver`

gargipanatula: Sorry couldn't quite grok the first comment - we're using `NewDefaultEndpointResolverV2` as the fallback resolver for when there's no endpoint overrides.

As for the custom resolvers looking identical - yes, they pretty much are, but I couldn't find a clean way to abstract them since they each have service specific parameters and return types (e.g. `elb.EndpointResolverV2` and `elbv2.EndpointResolverV2`)

pkg/providers/v1/aws_loadbalancer.go
gargipanatula: In general, I tried to reduce the number of changes from ptr->value to avoid latent bugs. We need to make the return type an array of values, but it's ok to keep the input as an array of pointers.

pkg/providers/v1/config/config.go
haoranleo: Oh I see! my bad missed the elb/elbv2 package name difference

haoranleo (approved):

pkg/providers/v1/aws_sdk.go
yue9944882:
```
var opts []func(*ec2.Options) = append(
         p.cfg.GetEC2EndpointOpts(regionName),
	 func(o *ec2.Options) {
		o.Retryer = &customRetryer{
			retry.NewStandard(),
		}
		o.EndpointResolverV2 = p.cfg.GetCustomEC2Resolver()
	})
```
i was thinking about an oneliner like this

Issue #1160, opened 2025-06-16
Hairpin connection failed on Service type-LoadBalancer NLB with internal scheme


mtulio:
**What happened**:

Service-type LoadBalancer NLB with internal scheme are failing to expose services in hairpin scenarios: when the client is hosted in the same node as server/backend/target.

```mermaid
graph TD
    subgraph Instance
       OVN
       NIC["Network Card (ENI)"]
    end
    subgraph OVN["OVN-Kubernetes network plugin"]
        direction TB
        A[Client Pod]
        C[Backend Pod]
    end
    B["Service NLB-internal (target instance)"]
    A -.->NIC  -.-> B -.-> NIC  -.-> C
```

This happens in the default target type (instance) on the Service-type NLB on internal scheme (default scheme). CCM by default enables the target attribute `Preserve client IP addresses`, which would be the root cause of this problem. The proxy is also off by default, and CCM does not expose API to customize both attributes.

Service object:

```yaml
$ oc get service/lbconfig-test-hp-nlb-int -o yaml -n cloud-provider-aws-9871
apiVersion: v1
kind: Service
metadata:
  annotations:
    aws-load-balancer-backend-protocol: http
    aws-load-balancer-ssl-ports: https
    service.beta.kubernetes.io/aws-load-balancer-internal: "true"
    service.beta.kubernetes.io/aws-load-balancer-target-node-labels: kubernetes.io/hostname=ip-10-0-4-233.ec2.internal
    service.beta.kubernetes.io/aws-load-balancer-type: nlb
  creationTimestamp: "2025-06-16T01:43:54Z"
  name: lbconfig-test-hp-nlb-int
  namespace: cloud-provider-aws-9871
spec:
  allocateLoadBalancerNodePorts: true
  clusterIP: 172.30.143.233
  clusterIPs:
  - 172.30.143.233
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: http
    nodePort: 31419
    port: 80
    protocol: TCP
    targetPort: 8080
  type: LoadBalancer
```

**What you expected to happen**:

Hairpin connection works on service type NLB with `internal` scheme when the targe type is `instance` by allowing customization of target group attributes:
- Preserve client IP addresses
- Proxy protocol v2

[...]

PR #1161, opened 2025-06-16
e2e/loadbalancer: added hairpin connection cases

mtulio:
**What type of PR is this?**

/kind bug
/kind failing-test

Implementing the hairpin connection test cases, and exposing an issue on NLB with internal scheme which fails when the client is trying to access a service loadbalancer which is hosted in the same node.

The hairpin connection is caused by the client IP preservation attribute is set to true (default), and the service does not provide an interface to prevent the issue.

The e2e is expecting to pass to prevent permanent failures in CI, but it is tracked by an issue https://github.com/kubernetes/cloud-provider-aws/issues/1160.

Fixes #

Those tests are important to increase coverage of scenarios that CCM declares as supported.

I also believe we can remove the hairpin with scheme internet-facing (public) LBs because the source IPs would be traversing a VPC gateway (IGW/NGW) and masquerade the real source, not reproducing the problem we are trying to expose in https://github.com/kubernetes/cloud-provider-aws/issues/1160. Thoughts?

```release-note
NONE
```

### DIFF
--- a/tests/e2e/loadbalancer.go
+++ b/tests/e2e/loadbalancer.go
@@ -16,12 +16,15 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2eservice "k8s.io/kubernetes/test/e2e/framework/service"
@@ -39,10 +42,6 @@ const (
 )
 
 var (
-	// clusterNodeSelector is the discovered node(compute/worker) selector used in the cluster.
-	clusterNodesSelector string
-	clusterNodesCount    int = 0
-
 	// lookupNodeSelectors are valid compute/node/worker selectors commonly used in different kubernetes
 	// distributions.
 	lookupNodeSelectors = []string{
@@ -50,6 +49,13 @@ var (
 		"node-role.kubernetes.io/node",   // used in ccm-aws CI
 	}
 )
+
+// ClusterNodeDiscovery holds information about discovered worker nodes.
+type ClusterNodeDiscovery struct {
+	Selector     string
+	Count        int
+	SingleWorker string
+}
 
 // loadbalancer tests
 var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
@@ -71,11 +77,14 @@ var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
 	})
 
 	type loadBalancerTestCases struct {
-		Name              string
-		ResourceSuffix    string
-		Annotations       map[string]string
-		PostConfigService func(cfg *configServiceLB, svc *v1.Service)
-		PostRunValidation func(cfg *configServiceLB, svc *v1.Service)
+		Name                           string
+		ResourceSuffix                 string
+		Annotations                    map[string]string
+		PostConfigService              func(cfg *configServiceLB, svc *v1.Service, nodeDiscovery ClusterNodeDiscovery)
+		PostRunValidation              func(cfg *configServiceLB, svc *v1.Service, nodeDiscovery ClusterNodeDiscovery)
+		HookInClusterTestReachableHTTP bool
+		RequireAffinity                bool
+		SkipTestFailure                bool
 	}
 	cases := []loadBalancerTestCases{
 		{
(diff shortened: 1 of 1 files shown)


mtulio:
Hi @kmala , would you mind stamping ok to test those new jobs, please?

I also have a question in the description related to the public cases, I used in the begging of exploration, but looks like we don't need it, LMK WDYT. Thanks

cc @elmiko

mtulio:
> @mtulio: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> pull-cloud-provider-aws-e2e	[24a0041](https://github.com/kubernetes/cloud-provider-aws/commit/24a0041499b3c43989b9c897cb918769c1968317)	[link](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/cloud-provider-aws/1161/pull-cloud-provider-aws-e2e/1934500455337955328)	true	`/test pull-cloud-provider-aws-e2e`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes&repo=cloud-provider-aws&pr=1161). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3Amtulio). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes/cloud-provider-aws/issues?q=is:issue+is:open) when you hit one in your PR.

CI infra issue.

mtulio:
I am observing a permanent failure on CI when launching the cluster trying to use an image that is no longer available:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/cloud-provider-aws/1161/pull-cloud-provider-aws-e2e/1934643980113285120#1:build-log.txt%3A751
```
s invalid: could not find Image for "099720109477/ubuntu/images/hvm-ssd-gp3/ubuntu-noble-24.04-amd64-server-20250502.1"
```

Hi @kmala @elmiko, do you know if is this comes from the test framework or is it possible to use a valid image in CCM repo?

mtulio:
An issue has been opened to track the CI problem: https://github.com/kubernetes/cloud-provider-aws/issues/1167

mtulio:
looks like [pull-cloud-provider-aws-e2e](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/cloud-provider-aws/1161/pull-cloud-provider-aws-e2e/1934695886340755456) is running (and stuck) in the last 48h. Just stopped it and trying to run again:

mtulio:
Looks like https://github.com/kubernetes/cloud-provider-aws/issues/1167 has been resolved. I manually stopped the running job (42h+); Triggering it again:

mtulio:
Hi @kmala @dims , PTAL? Thanks

elmiko (commented):
i think this mostly makes sense to me, curious to see the test output.

mtulio:
> i think this mostly makes sense to me, 

Thanks for reviewing.

> curious to see the test output.

Me too! =]  I just rebase and triggering the test again to try to get a good run. the tests so far are running til timeout.

mtulio:
Looks like those tests aren't healthy yet to any k8s distribution, converting to draft to increase verbosity on distro used on CI.

mtulio:
> Looks like those tests aren't healthy yet to any k8s distribution, converting to draft to increase verbosity on distro used on CI.

mtulio:
rebased and enhanced the SCC to prevent failing to schedule the test pod.

mtulio:
tuning the timeout:

mtulio:
updating improving the expected test failure to force to skip instead of passing:

mtulio:
> updating improving the expected test failure to force to skip instead of passing:
> 
> /test pull-cloud-provider-aws-e2e

I also removed the internet-facing tests that does not make sense to be tested in hairpin scenarios.

mtulio:
This PR is ready for review. In the latest version I reviewed:
- SCC to fit CI constraints running with non-root
- improve the doc strings of new functions 
- removed hairpin tests for internet-facing LBs as src IP is NATed and never hit frontend with node private IP to reproduce hairpin scenario
- Forcing to skip failed e2e on NLB-internal due known issue - preventing introducing permanent failure to CI


/assign dmis @kmala @elmiko

tests/e2e/loadbalancer.go
```diff
@@ -322,6 +367,25 @@ func (s *configServiceLB) buildReplicationController() func(rc *v1.ReplicationCo
...
+			rc.Spec.Template.Spec.Affinity = &v1.Affinity{
+				NodeAffinity: &v1.NodeAffinity{
+					RequiredDuringSchedulingIgnoredDuringExecution: &v1.NodeSelector{
+						NodeSelectorTerms: []v1.NodeSelectorTerm{
+							{
+								MatchExpressions: []v1.NodeSelectorRequirement{
+									{
+										Key:      "kubernetes.io/hostname",
+										Operator: v1.NodeSelectorOpIn,
+										Values:   []string{clusterNodeSingleWorker},
```
elmiko: does `clusterNodeSingleWorker` need to be a global variable?

is there any risk of parallel execution or similar in the test environment, this is my concern.

tests/e2e/loadbalancer.go
```diff
@@ -400,3 +463,244 @@ func getLBTargetCount(ctx context.Context, lbDNSName string, expectedTargets int
...
+			clusterNodesCount = len(nodeList.Items)
+			clusterNodesSelector = selector
+			for _, node := range nodeList.Items {
+				workerNodeList = append(workerNodeList, node.Name)
+			}
+			break
+		}
+	}
+
+	// Fail when no worker node is found - pourpuse of the function.
```
elmiko: spelling s/pourpuse/purpose/

elmiko (commented):
looking mostly good, i just have a question about the global variables.

tests/e2e/loadbalancer.go
mtulio: Thanks for review! Initially was intentionally global with a lock, but I just enclosed it to a struct. We are are now safer to run in parallel. :)

mtulio:
> looking mostly good, i just have a question about the global variables.

Thanks, @elmiko, good suggestions. Fixed.

elmiko (commented):
thanks Marco, i think that makes it a little less fragile.

/lgtm

mtulio:
Hey @cartermckinnon  , would you mind taking a look at this e2e test improvements to help us troubleshooting known issues? This won't fix https://github.com/kubernetes/cloud-provider-aws/issues/1160, but help us test it. Thanks!

kmala:
@oliviassss can you take a look?

oliviassss:
@mtulio Hi, thanks for the contribution. Just for my understanding, this PR mostly adds the test coverage for internal NLB hairpin issue, but in the test case itself we expect the test to fail and skip the failed test? What's the main purpose for adding the test cases?

From the AWS ELB doc I think the internal NLB will have hairpin issue with UDP or TCP_UDP protocol since the preserve client IP attribute cannot be disabled. (and I don't think CCM provide an annotation to disable this TG attributes anyway)

>By default, client IP preservation is enabled (and can't be disabled) for instance and IP type target groups with UDP and TCP_UDP protocols. However, you can enable or disable client IP preservation for TCP and TLS target groups using the preserve_client_ip.enabled target group attribute.

ref: https://docs.aws.amazon.com/elasticloadbalancing/latest/network/edit-target-group-attributes.html#client-ip-preservation

mtulio:
> @mtulio Hi, thanks for the contribution. Just for my understanding, this PR mostly adds the test coverage for internal NLB hairpin issue, but in the test case itself we expect the test to fail and skip the failed test? What's the main purpose for adding the test cases?
> 
> From the AWS ELB doc I think the internal NLB will have hairpin issue with UDP or TCP_UDP protocol since the preserve client IP attribute cannot be disabled. (and I don't think CCM provide an annotation to disable this TG attributes anyway)
> 
> > By default, client IP preservation is enabled (and can't be disabled) for instance and IP type target groups with UDP and TCP_UDP protocols. However, you can enable or disable client IP preservation for TCP and TLS target groups using the preserve_client_ip.enabled target group attribute.
> 
> ref: https://docs.aws.amazon.com/elasticloadbalancing/latest/network/edit-target-group-attributes.html#client-ip-preservation

Hi @oliviassss , answering your questions:

> his PR mostly adds the test coverage for internal NLB hairpin issue, but in the test case itself we expect the test to fail and skip the failed test?

Correct, we are skipping only the NLB test case as it works on CLB. https://github.com/kubernetes/cloud-provider-aws/pull/1161/files#diff-05b1c14f2de829d8a0c5f65b1b492a9ed9ab9d100ce6daa89d2d2347c8a14c77R122-R160

> What's the main purpose for adding the test cases?

The main purpose is to expose the test scenario (hairpin connection) for both supported Load Balancer by CCM: CLB and NLB, skipping the NLB case to prevent know failures on CCM CI.

This problem was unknown until now, we are exposing this scenario as "known issue" to CCM in https://github.com/kubernetes/cloud-provider-aws/issues/1160, and using this e2e as a helper to reproduce and **fix in follow up PRs**.

mtulio:
Hi @oliviassss @kmala , would you mind also triage the related issue https://github.com/kubernetes/cloud-provider-aws/issues/1160, please?

oliviassss (approved):

tests/e2e/loadbalancer.go
```diff
@@ -190,7 +219,16 @@ var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
 			ingressIP := e2eservice.GetIngressPoint(&lbService.Status.LoadBalancer.Ingress[0])
 			framework.Logf("Load balancer's ingress IP: %s", ingressIP)
 
-			e2eservice.TestReachableHTTP(ingressIP, svcPort, e2eservice.LoadBalancerLagTimeoutAWS)
+			// Hook: RemoteTestReachableHTTP changes the default test function to run the client in the cluster.
+			if tc.RemoteTestReachableHTTP {
```
kmala: nit: should we change `RemoteTestReachableHTTP` to `InClusterTestReachableHTTP` as remote suggests out of cluster ?

mtulio: Good suggestion, @kmala ! I will adjust it in the next commit. Thanks

PR #1166, opened 2025-06-17
chore: update go version to 1.24 in 1.31 release

shvbsle:
**What type of PR is this?**
>
> /kind feature

This updates the go version of 1.31 releases to 1.24. After bumping the version up during the release, we should be able to promote the image and artifacts for 1.31

Fixes # 
To test this I ran:
```
make
make test
```
And ensured that build succeeds and tests pass.

```release-note
NONE
```


fletcherw:
I don't think this fixes the mentioned issue until you run the image promo.

shvbsle:
> I don't think this fixes the mentioned issue until you run the image promo.

We are unable to run image-promo on 1.31.6 since the last-release was over 60 days old and objects in staging bucket expire after 60 days. The only way is to bump up the version to 1.31.7 but there were no changes added to 1.31 release since last bump of 1.31.6. This PR has some maintenance that can be used to cut a release and then we can run image-promo for 1.31.7

.ko.yaml
```diff
@@ -1 +1 @@
-defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.3.1-go1.22.7-bookworm.0
+defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.4.0-go1.24.0-bookworm.0
```
cartermckinnon: latest should be `1.24.4` AFAICT?

shvbsle: updated in the latest commit

Dockerfile
```diff
@@ -14,15 +14,15 @@
 ##                               BUILD ARGS                                   ##
 ################################################################################
 # This build arg allows the specification of a custom Golang image.
-ARG GOLANG_IMAGE=golang:1.22.7
+ARG GOLANG_IMAGE=golang:1.24.4
```
kmala, line 17: why directly 1.31? can we update the go version on master and then cherry pick for other versions?

shvbsle:
Closing this. There is a PR for master: https://github.com/kubernetes/cloud-provider-aws/pull/1170

Will raise another PR to cherry pick into 1.31

PR #1168, opened 2025-06-18
Upgrade KMS and Metadata clients to AWS SDK Go V2

gargipanatula:
**What type of PR is this?**
/kind cleanup

1. Upgrades usage of the AWS SDK for Go KMS and IMDS packages to V2.
2. Combines Ec2V2 client with the EC2 client returned by `Compute()`
3. Removed `assumeRoleProvider` since this was only used by SDK Go V1 clients
4. Added to e2e SDK client testing in `aws_sdk_test.go`. Now checks that the signing region & name, not only the URL were properly overridden.
5. Removed lots of unused code that was used for SDK V1 implementations. 

Fixes #

```release-note
NONE
```


pkg/providers/v1/aws_test.go
```diff
@@ -399,71 +393,6 @@ func TestOverridesActiveConfig(t *testing.T) {
 			if err != nil {
 				t.Errorf("Should succeed for case: %s, got %v", test.name, err)
 			}
```
gargipanatula, line 395: Note: this test used to test `ValidateOverrides`, and check for successful overrides of the URL, signing region, and signing name. Now, it only tests `ValidateOverrides`. Checking whether the override occurred successfully is now in `aws_sdk_test.go`. This is because endpoint resolution now happens at a client specific level, the AWS SDK Go V2 does not use a generic endpoint resolver.

PR #1169, opened 2025-06-18
Bump KMS and Metadata Clients to AWS SDK Go V2

gargipanatula:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


pkg/providers/v1/aws_test.go
```diff
@@ -192,16 +200,15 @@ type ServiceDescriptor struct {
 	signingName                  string
 }
```
gargipanatula, line 202: Note: This test used to test both `ValidateOverride` and the endpoint/signingname/signingregion overriding functionality. Now, it only tests `ValidateOverride`, and checking the overrides themselves has been moved to `aws_sdk_test.go` since override logic is configured directly with the clients.

pkg/providers/v1/variant/fargate/fargate.go
```diff
@@ -5,10 +5,10 @@ import (
 	"fmt"
 	"strings"
 
+	awsv2 "github.com/aws/aws-sdk-go-v2/aws"
```
kmala, line 8: nit: i see that the aws is only use for converting pointer to values and vice versa, can we remove that usage and such that we need not use awsv2

pkg/providers/v1/aws_ec2.go
```diff
@@ -87,6 +88,21 @@ func (s *awsSdkEC2) DescribeInstances(ctx context.Context, request *ec2.Describe
 	return results, nil
 }
 
+func (s *awsSdkEC2) DescribeInstanceTopology(ctx context.Context, input *ec2.DescribeInstanceTopologyInput, optFns ...func(*ec2.Options)) ([]ec2types.InstanceTopology, error) {
```
kmala, line 91: I see we are moving the functionality of the https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/services/aws_ec2.go to here, can we deprecate that file or is it used somewhere else?

pkg/providers/v1/aws.go
```diff
@@ -1122,9 +1070,15 @@ func (c *Cloud) buildSelfAWSInstance(ctx context.Context) (*awsInstance, error)
 	// information from the instance returned by the EC2 API - it is a
 	// single API call to get all the information, and it means we don't
 	// have two code paths.
-	instance, err := c.getInstanceByID(ctx, instanceID)
+	instanceIDBytes, err := io.ReadAll(instanceIDMetadata.Content)
+	if err != nil {
+		panic("unable to parse instance id")
```
kmala: why panic here?

pkg/providers/v1/aws_sdk.go
```diff
@@ -237,16 +190,15 @@ func (p *awsSDKProvider) LoadBalancingV2(ctx context.Context, regionName string,
...
 	if err != nil {
-		return nil, fmt.Errorf("unable to initialize AWS session: %v", err)
+		return nil, fmt.Errorf("unable to initialize AWS config: %v", err)
 	}
-	client := ec2metadata.New(sess)
-	p.addAPILoggingHandlers(&client.Handlers)
+	p.addAPILoggingMiddleware(&cfg)
+	imdsClient := imds.New(imds.Options{ClientEnableState: imds.ClientEnabled})
+	getInstanceIdentityDocumentOutput, err := imdsClient.GetInstanceIdentityDocument(context.Background(), &imds.GetInstanceIdentityDocumentInput{})
+	identity := getInstanceIdentityDocumentOutput.InstanceIdentityDocument
```
kmala: will getInstanceIdentityDocumentOutput have any value if there is error ?

pkg/providers/v1/aws_sdk.go
```diff
@@ -237,16 +190,15 @@ func (p *awsSDKProvider) LoadBalancingV2(ctx context.Context, regionName string,
 }
 
 func (p *awsSDKProvider) Metadata() (config.EC2Metadata, error) {
-	sess, err := session.NewSession(&aws.Config{
-		EndpointResolver: p.cfg.GetResolver(),
```
kmala, line 241: This should be migrated right?

pkg/providers/v1/variant/fargate/fargate.go
gargipanatula: Totally agree, but I think I'll punt this to my final PR in this migration, which will be solely focused on upgrading the remaining packages (just `github.com/aws/aws-sdk-go/aws` and `github.com/aws/aws-sdk-go/aws/awserr`). Wanted to keep this PR as simple as possible for ease of review.

pkg/providers/v1/aws_ec2.go
gargipanatula: Thanks for the catch - these have indeed been moved over so we can deprecate these files.

pkg/providers/v1/aws.go
gargipanatula: Ah yeah, this should probably be an error to match the rest of the code. Will change this in the next commit.

pkg/providers/v1/aws_sdk.go
gargipanatula: Thanks for the catch - we should check if `err != nil` here and return an error, because we get an error if the request fails/is unable to be parsed.

pkg/providers/v1/aws_sdk.go
```diff
@@ -237,16 +190,15 @@ func (p *awsSDKProvider) LoadBalancingV2(ctx context.Context, regionName string,
...
-	})
+	cfg, err := awsConfig.LoadDefaultConfig(context.TODO())
```
haoranleo: nit: use the default mode whenever possible

pkg/providers/v1/aws_sdk.go
haoranleo: Yeah this line `identity := getInstanceIdentityDocumentOutput.InstanceIdentityDocument` should be moved under `err==nil`

pkg/providers/v1/aws_sdk.go
gargipanatula: Thanks for the catch - migrated endpoint resolution. IMDS doesn't support signing, so its resolver will only override the URL. Standard SDK clients use [SigV4](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html) while IMDS uses a [different protocol](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html).

pkg/providers/v1/config/config.go
```diff
@@ -369,8 +339,71 @@ func (r *ELBV2Resolver) ResolveEndpoint(
...
+		}
+	}
+	return r.Resolver.ResolveEndpoint(ctx, params)
+}
+
+// GetIMDSEndpointOpts overrides the endpoint URL for IMDS clients
+func (cfg *CloudConfig) GetIMDSEndpointOpts() []func(*imds.Options) {
+	opts := []func(*imds.Options){}
+	for _, override := range cfg.ServiceOverride {
+		if override.Service == imds.ServiceID {
```
gargipanatula, line 397: Note: Instead of checking the ServiceID and region here, like we do for other clients, we only check ServiceID. This does not result in a behavioral change - the resolver for the SDK V1 `ec2metadata` package always had an empty value for the region at the time of the request, so only the serviceID field was used for determining whether to override.

Also, we are not overriding the signing name and region here since IMDS does not support signing.

pkg/providers/v1/aws.go
```diff
@@ -1204,23 +1159,34 @@ func (c *Cloud) describeLoadBalancerv2(ctx context.Context, name string) (*elbv2
...
-func (c *Cloud) findVPCID() (string, error) {
-	macs, err := c.metadata.GetMetadata("network/interfaces/macs/")
+func (c *Cloud) findVPCID(ctx context.Context) (string, error) {
+	macsMetadata, err := c.metadata.GetMetadata(ctx, &imds.GetMetadataInput{Path: "network/interfaces/macs/"})
 	if err != nil {
 		return "", fmt.Errorf("could not list interfaces of the instance: %q", err)
 	}
+	macsBytes, err := io.ReadAll(macsMetadata.Content)
+	if err != nil {
+		panic("unable to parse macs")
```
kmala: i don't think we should panic here?

pkg/providers/v1/aws_sdk.go
```diff
@@ -236,18 +189,27 @@ func (p *awsSDKProvider) LoadBalancingV2(ctx context.Context, regionName string,
...
-	identity, err := client.GetInstanceIdentityDocument()
+	p.addAPILoggingMiddleware(&cfg)
+
+	// Unlike other SDK clients, the IMDS client does not support signing, so any overrides of the signing region and name
+	// from awsSDKProvider.cfg will not be recognized.
+	// Standard SDK clients use SigV4: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html
+	// But IMDS uses a different request pattern: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
+	var opts []func(*imds.Options) = p.cfg.GetIMDSEndpointOpts()
+	opts = append(opts, func(o *imds.Options) {
+		o.ClientEnableState = imds.ClientEnabled // enable requests, otherwise the AWS_EC2_METADATA_DISABLED env var will need to be set
```
kmala: for my understanding what is the difference between default enabled and enabled https://github.com/aws/aws-sdk-go-v2/blob/feature/ec2/imds/v1.16.32/feature/ec2/imds/api_client.go#L37 as that is the default value ?

pkg/providers/v1/aws.go
kmala: there shouldn't be panic here right?

pkg/providers/v1/aws_test.go
```diff
@@ -399,71 +393,6 @@ func TestOverridesActiveConfig(t *testing.T) {
 			if err != nil {
 				t.Errorf("Should succeed for case: %s, got %v", test.name, err)
 			}
-
-			if len(cfg.ServiceOverride) != len(test.servicesOverridden) {
```
kmala, line 403: why is this removed? can't we test some if not all?

pkg/providers/v1/aws_sdk.go
gargipanatula: If we do not configure this variable, it is set to `defaultEnable` and falls back to whatever the `AWS_EC2_METADATA_DISABLED` env var is set to.

If we do configure it, then this overrides the value of `AWS_EC2_METADATA_DISABLED` and acts as if `AWS_EC2_METADATA_DISABLED = false`.

However, I dug into the old `ec2metadata` code and it essentially behaves the same way (it needs the same env var to be set to make successful requests). So, I'm going to remove this line so we don't override `AWS_EC2_METADATA_DISABLED` with `o.ClientEnableState = imds.ClientEnabled` and we can maintain parity with the old code. Thanks for the callout!

Info about `AWS_EC2_METADATA_DISABLED` from the [docs](https://docs.aws.amazon.com/sdkref/latest/guide/feature-imds-credentials.html#:~:text=Configure%20this%20functionality%20by%20using,Default%20value:%20false%20.):
```
AWS_EC2_METADATA_DISABLED - environment variable
Whether or not to attempt to use Amazon EC2 Instance Metadata Service (IMDS) to obtain credentials.

Default value: false.

Valid values:

true – Do not use IMDS to obtain credentials.

false – Use IMDS to obtain credentials.
```

pkg/providers/v1/aws_test.go
gargipanatula: This part of the test was checking that the signing fields and URL were properly overridden using the generic resolver from V1. Since we now have resolvers on a per client basis, we are now testing the resolvers on a more e2e scale, creating a client and making mock requests (all in `aws_sdk_test.go`). We're still testing the override functionality, just in a different place.

This test used to test both `ValidateOverride()` and overriding, now it just tests `ValidateOverride()`.

Let me know if I'm missing something here, though.

kmala: i was thinking if we can keep the check `len(cfg.ServiceOverride) != len(test.servicesOverridden)` but i think it should be okay as this should be tested indirectly in the other function.

PR #1187, opened 2025-07-02
Fix backend protocol annotation incorrectly documented

### LINKED ISSUE #1186: Backend protocol annotation incorrectly documented as supported for NLB
**What happened**:

Inside the documentation the following statements are incorrect/unclear:
- `service.beta.kubernetes.io/aws-load-balancer-backend-protocol` this annoation is only supported when the underlying balancer is Classic ELB, therefore the description doesn't apply to NLB. The default protocol when this annotation is not specified is `tcp` not `http`. The description could also a bit improved a little bit to be clear to the reader

Therefore it could be usefult to further clarify the usage of the following annotation as well:

- `service.beta.kubernetes.io/aws-load-balancer-type`: if the annotation above is not supported for the NLB, it could not be clear how the NLB understands which protocol to use (it is automatically derived from the Kubernetes service itself). So it's better to explicitly define what is the behavior in this case

/kind bug

samuel-esp:
**What type of PR is this?**
 /kind bug
 
Fix backend protocol and type annotations in documentation: the backend protocol annotation only applies to Classic ELB, therefore there are incorrect statements inside the description (`tcp` is the default protocol, not `http`). Therefore, in the type annotation, the user should be able to understand the behavior behind the protocol assignment for NLB since it is different from the one described inside the backend protocol annotation (in the past, the user might have been misled into thinking the process was the same since both ELB type were shown to be compatible with the backend protocol annotation)
 
Fixes #1186
 
None
 
```release-note
Fix backend protocol annotation scope in documentation: which only applies to classic ELB. The description for the backend is rewritten to be more clear and to clean incorrect statements. The description for the type annotation is enhanced to explain the behavior behind the protocol assignment for NLB
```


docs/service_controller.md
```diff
@@ -3,33 +3,33 @@
...
+| service.beta.kubernetes.io/aws-load-balancer-type                              | [nlb]                               | -                         | ELB,NLB | Indicates the type of Load Balancer. The only valid value is `nlb`, this means that leaving this field blank or omitting the annotation is equivalent to selecting ELB. When selecting `nlb`, the backend protocol is automatically derived from the protocol defined in the Kubernetes Service, as long as that protocol is supported by NLB (supported protocols include TCP, the default for a Kubernetes Service, and UDP). If the protocol is TCP and the `aws-load-balancer-ssl-cert` annotation is also present, the NLB protocol will be set to TLS                                                                         |
```
mtulio: nit
```suggestion
| service.beta.kubernetes.io/aws-load-balancer-type                              | [nlb]                               | -                         | ELB,NLB | Indicates the type of Load Balancer. The only valid value is `nlb`, this means that leaving this field blank or omitting the annotation is equivalent to selecting ELB. When selecting `nlb`, the backend protocol is automatically derived from the protocol defined in the Kubernetes Service, as long as that protocol is supported by NLB (supported protocols include TCP, the default for a Kubernetes Service, and UDP). If the protocol is TCP and the `aws-load-balancer-ssl-cert` annotation is also present, the NLB protocol will be set to TLS.                                                                        |
```

docs/service_controller.md
```diff
@@ -3,33 +3,33 @@
...
+| service.beta.kubernetes.io/aws-load-balancer-backend-protocol                  | [http\|https\|ssl\|tcp]             | tcp                       | ELB | Specifies the protocol spoken by the backend (pod) behind a listener. If set to `http` and the `aws-load-balancer-ssl-cert` annotation is missing or its value is empty (`""`), a HTTP listener is created; if set to `http` or `https` and the `aws-load-balancer-ssl-cert`annotation is also present, an HTTPS listener is created to terminate SSL and parse headers. If set to `ssl` or `tcp`, a "raw" SSL/TCP listener is used. if the annotation is missing, its value is empty (`""`), or set to `https` without the `aws-load-balancer-ssl-cert`annotation also being present, the default protocol is assumed to be `tcp`. |
```
mtulio: Looks like NLB also supports it: https://github.com/kubernetes/cloud-provider-aws/blob/af0e606f271a47809c50e0e84a7249f1d188f229/pkg/providers/v1/aws.go#L2176-L2196
```suggestion
| service.beta.kubernetes.io/aws-load-balancer-backend-protocol                  | [http\|https\|ssl\|tcp]             | tcp                       | ELB,NLB | Specifies the protocol spoken by the backend (pod) behind a listener. If set to `http` and the `aws-load-balancer-ssl-cert` annotation is missing or its value is empty (`""`), a HTTP listener is created; if set to `http` or `https` and the `aws-load-balancer-ssl-cert`annotation is also present, an HTTPS listener is created to terminate SSL and parse headers. If set to `ssl` or `tcp`, a "raw" SSL/TCP listener is used. if the annotation is missing, its value is empty (`""`), or set to `https` without the `aws-load-balancer-ssl-cert`annotation also being present, the default protocol is assumed to be `tcp`. |
```

samuel-esp: Hey @mtulio, could you try to replicate that? Because when i wrote the docs I did several tests and the protocol showed in the console never matched the one I used in the annotation; looking at the code this made sense to me. A double check would be awesome

samuel-esp: Hey @mtulio did you have time to check this?

mtulio: Hello @samuel-esp , apologies by the delay - you are correct, I followed incorrectly and the snippeted I shared above only changes the `TrafficProtocol` when `ssl`, skipping when `http`. I believe you can skip this suggestion.

mtulio:
Hey @samuel-esp , do you still want to proceed with this proposal? If so, would you mind rebsae it? thanks!

samuel-esp:
Hey @mtulio, yes of course I'm going to rebase it as soon as i can

samuel-esp:
@mtulio should be ok by now

mtulio:
> @mtulio should be ok by now

Thanks, one more comment: would you mind applying only the changes proposed in the PR, looks like the formatting isn't part of the documentation/fix, am I reading correctly? ( I am not sure if I followed the benefit of inserting the additional tabulation as the markdown will rendered the table properly).

samuel-esp:
@mtulio reverted the formatting, i think it was an artifact of rebasing. Note that this annotation:

`service.beta.kubernetes.io/aws-load-balancer-internal: true`

should be modified to

`service.beta.kubernetes.io/aws-load-balancer-internal: "true"`

but this is outside the scope of this PR

docs/service_controller.md
elmiko: looks like the period got dropped from the latest commit.

docs/service_controller.md
```diff
@@ -10,7 +10,7 @@ The service controller is responsible for watch for service and node object chan
...
```
elmiko: minor nit

```suggestion
| service.beta.kubernetes.io/aws-load-balancer-backend-protocol                  | [http\|https\|ssl\|tcp]             | tcp                       | ELB | Specifies the protocol spoken by the backend (pod) behind a listener. If set to `http` and the `aws-load-balancer-ssl-cert` annotation is missing or its value is empty (`""`), an HTTP listener is created; if set to `http` or `https` and the `aws-load-balancer-ssl-cert`annotation is also present, an HTTPS listener is created to terminate SSL and parse headers. If set to `ssl` or `tcp`, a "raw" SSL/TCP listener is used. if the annotation is missing, its value is empty (`""`), or set to `https` without the `aws-load-balancer-ssl-cert`annotation also being present, the default protocol is assumed to be `tcp`. |
```

elmiko (commented):
this is great, thank you for the update @samuel-esp . just a couple minor nits, then i think we can merge this.

docs/service_controller.md
samuel-esp: Thanks for pointing that out!

mtulio:
@samuel-esp would you mind squash the commits?

samuel-esp:
> @samuel-esp would you mind squash the commits?

@mtulio do you want me to force push as well to clean the rebase commits?

elmiko:
@samuel-esp i would say just squash your commits, then rebase it.

are we carrying commits for another PR as well? (i wouldn't expect so many commits in there)

mtulio:
> > @samuel-esp would you mind squash the commits?
> 
> @mtulio do you want me to force push as well to clean the rebase commits?

+1 to what @elmiko said in the last comment. We expect only your commits and documentation updates in this PR.

samuel-esp:
@mtulio @elmiko should be ok now

PR #1207, opened 2025-07-10
fix/lb SG rule update failures when service port is changed

### LINKED ISSUE #1206: Unable to update Service ports due wrong diff calculation of security group rules
**What happened**:

When you create a service[1], and then update the ports, the controller is not calculating correctly the difference between existing security group rules and the new one, leading in duplicated list to add failing when trying to insert it as `\" already exists\""`.

Service updated and Security Group rules for NLB are correctly updated.

Steps to reproduce:
- 1) Create a Service
```sh
SVC_NAME=$APP_NAME_BASE-svc-clb1
cat << EOF | kubectl create -f -
apiVersion: v1
kind: Service
metadata:
  name: $SVC_NAME
  namespace: ${APP_NAMESPACE}
spec:
  selector:
    app: $APP_NAME_BASE
  ports:
    - name: http80
      port: 80
      targetPort: 8080
      protocol: TCP
  type: LoadBalancer
EOF
```
- 2) Update the Service - example appending a new port:
```sh
kubectl patch service $SVC_NAME -n ${APP_NAMESPACE} --type=json \
  --patch '[{"op": "add", "path": "/spec/ports/-", 
    "value": {"name":"http81","port":81,"protocol":"TCP","targetPort":8080}}]'
```
- 3. Check the controller logs:
```sh
I0710 13:38:38.052182 2677571 aws.go:1344] Adding security group ingress: sg-0dbd438198aded0a7 [{0xc000aef930 0xc000d5e8c0 [{0xc000d5e890 <nil> {}}] [] [] 0xc000aef934 [] {}} {0xc000aef908 0xc000d5e8a0 [{0xc000d5e890 <nil> {}}] [] [] 0xc000aef90c [] {}} {0xc000aef928 0xc000d5e8b0 [{0xc000d5e890 <nil> {}}] [] [] 0xc000aef92c [] {}}]

E0710 13:38:38.314462 2677571 controller.go:301] "Unhandled Error" err="error processing service app-sample/app-sample-svc-clb1 (retrying with exponential backoff): failed to ensure load balancer: error authorizing security group ingress: \"operation error EC2: AuthorizeSecurityGroupIngress, https response error StatusCode: 400, RequestID: 2b738da5-4c37-46dc-b6eb-6ff33868d89f, api error InvalidPermission.Duplicate: the specified rule \\\"peer: 0.0.0.0/0, TCP, from port: 80, to port: 80, ALLOW\\\" already exists\"" logger="UnhandledError"

I0710 13:38:38.314507 2677571 event.go:389] "Event occurred" object="app-sample/app-sample-svc-clb1" fieldPath="" kind="Service" apiVersion="v1" type="Warning" reason="SyncLoadBalancerFailed" message="Error syncing load balancer: failed to ensure load balancer: error authorizing security group ingress: \"operation error EC2: AuthorizeSecurityGroupIngress, https response error StatusCode: 400, RequestID: 2b738da5-4c37-46dc-b6eb-6ff33868d89f, api error InvalidPermission.Duplicate: the specified rule \\\"peer: 0.0.0.0/0, TCP, from port: 80, to port: 80, ALLOW\\\" already exists\""
```

Looks like the problem is in the `IPPermissionSet.Difference()` in the AWS Go SDK v2 comparing pointers and not identifying correctly the diff, always adding existing rules to the SG, leading to the failure in the step 3.

https://github.com/kubernetes/cloud-provider-aws/blob/af0e606f271a47809c50e0e84a7249f1d188f229/pkg/providers/v1/aws.go#L1329-L1330

More details and debug log available here: https://github.com/kubernetes/cloud-provider-aws/pull/1158#issuecomment-3055026029 

**Environment**:
- Kubernetes version (use `kubectl version`): k8s 1.33, CCM built from main
- Cloud provider or hardware configuration: Default configuration. Version in main
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

/kind bug

mtulio:
**What type of PR is this?**

/kind bug

The Difference method calculation are returning wrong values on service update calculating the security group rules.

The core issue is caused by it is always comparing the pointer address, not the values, always generating difference between lists.

Example:
given s1:{a1, a2} and s2:{a2,a3,a4}
s2.Difference(s1) was returning {a2, a3, a4}, instead {a3, a4}

Fixes #1206

```release-note
Fixed security group rule difference calculation comparing pointers instead of values, 
eliminating failures on rule updates.
```

### DIFF
--- a/pkg/providers/v1/sets_ippermissions.go
+++ b/pkg/providers/v1/sets_ippermissions.go
@@ -141,7 +141,7 @@ func (s IPPermissionSet) Equal(s2 IPPermissionSet) bool {
 	return len(s) == len(s2) && s.IsSuperset(s2)
 }
 
-// Difference returns a set of objects that are not in s2
+// Difference returns a set of objects that are not in s2.
 // For example:
 // s1 = {a1, a2, a3}
 // s2 = {a1, a2, a4, a5}
@@ -149,10 +149,16 @@ func (s IPPermissionSet) Equal(s2 IPPermissionSet) bool {
 // s2.Difference(s1) = {a4, a5}
 func (s IPPermissionSet) Difference(s2 IPPermissionSet) IPPermissionSet {
 	result := NewIPPermissionSet()
-	for k, v := range s {
-		_, found := s2[k]
+	for _, desired := range s.List() {
+		found := false
+		for _, existing := range s2.List() {
+			if ipPermissionExists(&desired, &existing, false) {
+				found = true
+				break
+			}
+		}
 		if !found {
-			result[k] = v
+			result.Insert(desired)
 		}
 	}
 	return result
--- a/pkg/providers/v1/sets_ippermissions_test.go
+++ b/pkg/providers/v1/sets_ippermissions_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestUngroup(t *testing.T) {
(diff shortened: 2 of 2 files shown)


elmiko:
not sure this will work

/ok-to-test

elmiko (commented):
this makes sense to me, thanks for adding all the test too!

defer to @kmala for approval

/lgtm

PR #1230, opened 2025-08-22
fix: update instance types requiring topology labels

ndbaker1:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Adds `p6e-gb200` and `p6-b200` to matching regexp for instances that support DescribeInstanceTopology API, and should block node initialization.

updates values overall based on latest in https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-topology-prerequisites.html

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/providers/v1/topology.go
+++ b/pkg/providers/v1/topology.go
@@ -38,14 +38,9 @@ const instanceTopologyManagerCacheTimeout = 24 * time.Hour
 We need to ensure that instance types that we expect a response will not successfully complete syncing unless
 we get a response, so we can track known instance types that we expect to get a response for.
 
-Supported instance types for DescribeInstanceTopology as of 2/6/25 from API documentation:
-https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceTopology.html
-
-hpc6a.48xlarge | hpc6id.32xlarge | hpc7a.12xlarge | hpc7a.24xlarge | hpc7a.48xlarge | hpc7a.96xlarge | hpc7g.4xlarge | hpc7g.8xlarge | hpc7g.16xlarge
-p3dn.24xlarge | p4d.24xlarge | p4de.24xlarge | p5.48xlarge | p5e.48xlarge | p5en.48xlarge
-trn1.2xlarge | trn1.32xlarge | trn1n.32xlarge | trn2.48xlarge | trn2u.48xlarge
+https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-topology-prerequisites.html
 */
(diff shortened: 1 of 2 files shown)


pkg/providers/v1/topology.go
```diff
@@ -38,14 +38,9 @@ const instanceTopologyManagerCacheTimeout = 24 * time.Hour
...
-Supported instance types for DescribeInstanceTopology as of 2/6/25 from API documentation:
-https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceTopology.html
-
-hpc6a.48xlarge | hpc6id.32xlarge | hpc7a.12xlarge | hpc7a.24xlarge | hpc7a.48xlarge | hpc7a.96xlarge | hpc7g.4xlarge | hpc7g.8xlarge | hpc7g.16xlarge
-p3dn.24xlarge | p4d.24xlarge | p4de.24xlarge | p5.48xlarge | p5e.48xlarge | p5en.48xlarge
-trn1.2xlarge | trn1.32xlarge | trn1n.32xlarge | trn2.48xlarge | trn2u.48xlarge
+https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-topology-prerequisites.html
 */
-var defaultSupportedTopologyInstanceTypePattern = regexp.MustCompile(`^(hpc|trn|p|inf)[0-9]+[a-z]*(\.[0-9a-z]*)$`)
+var defaultSupportedTopologyInstanceTypePattern = regexp.MustCompile(`^(hpc|trn|p|inf)[0-9]+[a-z]*(-[a-z0-9]+)?(\.[0-9a-z]*)$`)
```
mselim00: do you think we could get away with just enforcing the instance series?
```suggestion
var defaultSupportedTopologyInstanceTypePattern = regexp.MustCompile(`^(hpc|trn|p|inf)[0-9]+`)
```

mselim00: not directly related, but is `inf` actually supposed to be on this list? it's not in the doc

ndbaker1: yea, seeing that too. will update and add to the test cases for regressions

ndbaker1: a similar topic from the original PR https://github.com/kubernetes/cloud-provider-aws/pull/1100#discussion_r1951312764. i think it would be fine, but keeping this expectation for now

mselim00: I think we'd match a few unsupported instance types like `p3.2xlarge` in either case, I'm just not really sure what benefit we get from enforcing the rest of the scheme. seems like we might just end up constantly updating the rest of the format to satisfy any new format changes in a supported series

ndbaker1: yea, what we really need is either (1) an api that lists the instance types or (2) the DIT API to report support OOTB instead of an empty list

PR #1232, opened 2025-08-23
Add support private ipv4 addresses annotation for NLB

### LINKED ISSUE #1231: Add NLB loadbalancer private ipv4 addresses annotation similar to the elastic allocation ids annotation
For an NLB in AWS, for subnet mapping I should be able to add a private ips from the subnets that I'm mapping to
https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/service/annotations/#private-ipv4-addresses

**Why is this needed**:
When using in production like environments, whenever we are deploying we want use static ips for the nlb so that it's reliable incase even if the loadbalancer is recreated. This feature is already present in the aws-loadbalancer-controller.

/kind feature

luckyganesh:
**What type of PR is this?**

/kind feature

For an NLB in AWS, for subnet mapping I should be able to add a private ips from the subnets that I'm mapping to
https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/service/annotations/#private-ipv4-addresses

Fixes #1231

```release-note
NONE
```


pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -1466,14 +1474,17 @@ func elbListenersAreEqual(actual, expected elbtypes.Listener) bool {
...
+func createSubnetMappings(subnetIDs []string, allocationIDs []string, privateIPv4Addresses []string) []elbv2types.SubnetMapping {
 	response := []elbv2types.SubnetMapping{}
 
 	for index, id := range subnetIDs {
 		sm := elbv2types.SubnetMapping{SubnetId: aws.String(id)}
 		if len(allocationIDs) > 0 {
 			sm.AllocationId = aws.String(allocationIDs[index])
 		}
+		if len(privateIPv4Addresses) > 0 {
+			sm.PrivateIPv4Address = aws.String(privateIPv4Addresses[index])
```
mtulio, line 1486: is there any guarantee that the order of private IP belongs to the subnet in the same index?

luckyganesh: There is no guarantee, but the user must provide it in the same order.

Even in the aws-load-balancer-controller it is the same behaviour.
https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/service/annotations/#private-ipv4-addresses

luckyganesh: if we don't provide it in the same order, then the loadbalancer creation will fail from aws sdk itself.

luckyganesh:
@olemarkus please check it, all the tests are passing. Let me know incase if any changes are required

olemarkus (approved):

PR #1239, opened 2025-08-26
label nodes with the name of the autoscaling group they belong to

### LINKED ISSUE #884: label nodes with the name of the autoscaling group they belong to (if they belong to one)
It would be extremely helpful for operators if nodes were labelled with the autoscaling group they belong to (if there is an autoscaling group associated with a given node). I think `node.kubernetes.io/auto-scaling-group-name` would be appropriate for this.

**Why is this needed**:
Alleviate the need to cross-reference autoscaling group instances in AWS with node names in kubectl.

I would be happy to submit a PR if this would be considered for acceptance. Let me know!

/kind feature

luckyganesh:
**What type of PR is this?**
> /kind feature

Fixes #884 

```release-note
NONE
```


pkg/providers/v1/well_known_labels.go
```diff
@@ -24,4 +24,6 @@ const (
 	// but will be initially applied to nodes. The suffix should be an incremented
 	// integer starting at 1.
 	LabelNetworkNodePrefix = "topology.k8s.aws/network-node-layer-"
+
+	LabelAutoScalingGroupName = "node.k8s.aws/auto-scaling-group-name"
```
lmktfy, line 28: Did AWS say that this this is OK?

luckyganesh: I have taken it from the issue itself, but incase if we want to change I can do that

PR #1152, opened 2025-05-27
Adding batching to describe instance API calls

sarthakkothari:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

Node-controller calls DescribeInstances API every time a node joins the cluster. When multiple nodes join the cluster simultaneously, separate EC2 DescribeInstances API calls are made for each node, even though the API supports querying multiple instances in a single call. We are batching the call to EC2 API which reduces the overall call volumes made to EC2 and improve performance.

We have a similar batching for other EC2 calls in [Karpenter](https://github.com/aws/karpenter-provider-aws/tree/0a8251d6eba027c7d34acaf9d0fee2fc40832b67/pkg/batcher)

Sample Logs
```
I0520 00:04:26.135372   16837 descibe_instance_batch.go:83] Batched describe instances {
  InstanceIds: [
    "i-038c319c1b9c0c75e",
    "i-02b5e9c459a6cc573",
    "i-04908d55bbfe8be74",
    "i-072a0a9491bacf270"
  ]
}
```

Reduction in EC2 call volumes with batch -
Screenshot 2025-05-27 at 11 40 39 AM

Without batch -

image

```release-note
Improve tagging controller performance by batching EC2 describe instance API calls.
```

PR #1260, opened 2025-09-23
Fix the batched describe insance method for instance not found case

kmala:
**What type of PR is this?**

 /kind bug

Fix the describe instance batcher when the instance is not found or the batch call return error. If the instance is not found we shouldn't be returning an array of empty object but should be returning a nil. This is causing CCM to panic when the nodes deletion in apiserver is blocked because of finalizers but the instance in the aws is deleted. This bug is introduced in the https://github.com/kubernetes/cloud-provider-aws/pull/1152

Fixes #

```release-note

```

### DIFF
--- a/pkg/providers/v1/describe_instance_batch.go
+++ b/pkg/providers/v1/describe_instance_batch.go
@@ -57,6 +57,9 @@ func (b *describeInstanceBatcher) DescribeInstances(ctx context.Context, input *
 		return nil, fmt.Errorf("expected to receive a single instance only, found %d", len(input.InstanceIds))
 	}
 	result := b.batcher.Add(ctx, input)
+	if result.Output == nil {
+		return nil, result.Err
+	}
 	return []*ec2types.Instance{result.Output}, result.Err
 }
 
@@ -74,7 +77,8 @@ func describeInstanceHasher(ctx context.Context, input *ec2.DescribeInstancesInp
 func execDescribeInstanceBatch(ec2api iface.EC2) batcher.BatchExecutor[ec2.DescribeInstancesInput, ec2types.Instance] {
 	return func(ctx context.Context, inputs []*ec2.DescribeInstancesInput) []batcher.Result[ec2types.Instance] {
 		results := make([]batcher.Result[ec2types.Instance], len(inputs))
-		firstInput := inputs[0]
+
+		firstInput := *inputs[0]
 		// aggregate instanceIDs into 1 input
 		for _, input := range inputs[1:] {
 			firstInput.InstanceIds = append(firstInput.InstanceIds, input.InstanceIds...)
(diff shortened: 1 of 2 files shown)


pkg/providers/v1/describe_instance_batch.go
```diff
@@ -74,7 +77,8 @@ func describeInstanceHasher(ctx context.Context, input *ec2.DescribeInstancesInp
 func execDescribeInstanceBatch(ec2api iface.EC2) batcher.BatchExecutor[ec2.DescribeInstancesInput, ec2types.Instance] {
 	return func(ctx context.Context, inputs []*ec2.DescribeInstancesInput) []batcher.Result[ec2types.Instance] {
 		results := make([]batcher.Result[ec2types.Instance], len(inputs))
-		firstInput := inputs[0]
+
+		firstInput := *inputs[0]
```
sarthakkothari, line 81: why did we translate this from a pointer? can potentially cause another panic?

kmala: if we don't translate we are overriding it in the below for loop as firstInput would just be a pointer to the first item item in slice. inputs should never have nil items for the panic to happen and there should always be atleast 1 item in the inputs.

gargipanatula: Should we still have a `len` check on `inputs` before this line to catch the case where `inputs` is empty? Unless we have a hard guarantee somewhere that `inputs` is not empty?

kmala: yes, we are guaranteed to not have empty list as this gets called only when we make a describe instance call and there it verifies the instance count.

pkg/providers/v1/describe_instance_batch.go
```diff
@@ -57,6 +57,9 @@ func (b *describeInstanceBatcher) DescribeInstances(ctx context.Context, input *
 		return nil, fmt.Errorf("expected to receive a single instance only, found %d", len(input.InstanceIds))
 	}
 	result := b.batcher.Add(ctx, input)
+	if result.Output == nil {
```
yue9944882, line 60: is `if result.Err != nil` a better check here? i wonder whether we're actually getting an error e.g. 404 not found wen describing a non-existent instance, it's trickier when e.g. we batch two describe instance IDs where one exists and the other is non-existent, are we getting an error in this case?

yue9944882:
I think it's more complex than a nil check. Bundling the describe calls via introducing a batcher did lower down the API counts but it also adds some complexity regarding error handling. i just did a local experiment:

1. Calling describe API with two valid instance ID -> returns an array of two elements
2. Calling describe API with 1 valid and 1 non-existent instance ID -> returns 404 error

@sarthakkothari @shiv-amz can you confirm the batcher behavior and add related test cases to that?

yue9944882:
/lgtm

Discussed with @kmala offline, merge this in as this PR will fix the gap for error handling.

Issue #1264, opened 2025-09-28
provide an option to use `eks.amazonaws.com/role-arn` annotation instead of `eks.amazonaws.com/ecr-role-arn`


bear-san:
**What would you like to be added**:
- when finding IAM Role using for assume role also support `eks.amazonaws.com/role-arn` annotation instead of `eks.amazonaws.com/ecr-role-arn`
- If option not provided both `eks.amazonaws.com/ecr-role-arn` and environment variable of `AWS_ECR_ROLE_ARN` , provider will try to find a role by `eks.amazonaws.com/role-arn` annotation

**Why is this needed**:
- in the default, Pod identity webhook injects project token to annotated `eks.amazonaws.com/role-arn` ref: https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/pkg/annotations.go#L21
- If user wants to project token for only pull image, still must specify both  of `eks.amazonaws.com/role-arn` and `eks.amazonaws.com/ecr-role-arn`

/kind feature

PR #1281, opened 2025-10-17
fix: allow providing unused ServiceAccountToken

fletcherw:
/kind bug

When ecr-credential-provider is configured to use ServiceAccountTokens for fetching ECR credentials, it should allow ignoring those credentials if the service account doesn't have the corresponding annotations.

Update the provider to try to use a ServiceAccountToken and fall back to standard local credentials otherwise.

---

Kind of related to #1264.

```
Changed ecr-credential-provider to fall back to using default credentials when a ServiceAccountToken is provided but there is no corresponding role ARN provided to assume. This should not cause changes for users using the default credential provider config.
```

### DIFF
--- a/cmd/ecr-credential-provider/main.go
+++ b/cmd/ecr-credential-provider/main.go
@@ -207,20 +207,20 @@ func (e *ecrPlugin) buildCredentialsProvider(ctx context.Context, request *v1.Cr
 	}
 
 	return aws.CredentialsProviderFunc(func(ctx context.Context) (aws.Credentials, error) {
-			assumeOutput, err := e.sts.AssumeRoleWithWebIdentity(ctx, &sts.AssumeRoleWithWebIdentityInput{
-				RoleArn:          aws.String(arn),
-				RoleSessionName:  aws.String("ecr-credential-provider"),
-				WebIdentityToken: aws.String(request.ServiceAccountToken),
-			})
-			if err != nil {
-				return aws.Credentials{}, fmt.Errorf("failed to assume role: %w", err)
-			}
-			return aws.Credentials{
-				AccessKeyID:     *assumeOutput.Credentials.AccessKeyId,
-				SecretAccessKey: *assumeOutput.Credentials.SecretAccessKey,
-				SessionToken:    *assumeOutput.Credentials.SessionToken,
-			}, nil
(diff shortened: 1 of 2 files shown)


cmd/ecr-credential-provider/main.go
jicowan, line 1: So far as I know, you can only assign 1 IAM role to a pod through IRSA. However, it's unclear whether SA token authentication requires IRSA. If the `ecr-role-arn` and `role-arn` have to be the same, then we should update this PR to only use `role-arn` as that is key that IRSA uses.

fletcherw: I'm a little new to this but this is my understanding

Each pod has exactly one service account associated with it. By default, if you annotate that service account with `eks.amazonaws.com/role-arn`, the pod identity agent webhook will get a service account token for that ~node~ account, assume that role and pass those credentials into the pod's containers.

Separately, before a pod is created, if there is a CredentialProvider that has `matchImages` matching the image URL and that configures `tokenAttributes`, the kubelet will also create a Service Account token and pass that to the credential provider.

I don't see why the IAM role that the credential provider assumes would have to match the role that the pod identity webhook assumes, though I'm sure that for many people they will be the same.

jicowan:
Using only the ecr-role-arn did not work when I tested it. The service account needed both annotations: ecr-role-arn and role-arn. Having a separate annotation for image pulls could be useful for separating concerns: 1 role for image pulls, and another for accessing AWS APIs.

Sent from Outlook<https://aka.ms/qtex0l> for iOS
________________________________

fletcherw:
What credential provider config were you using in that situation? If you had specified `"requiredServiceAccountAnnotationKeys":
  ["eks.amazonaws.com/ecr-role-arn", "eks.amazonaws.com/role-arn"]` it makes sense it would require both, but with `optionalServiceAccountAnnotationKeys` instead I think just one would be fine.

jicowan:
I tried it with both. Required and Optional. I can re-test next week.
________________________________

cmd/ecr-credential-provider/main.go
kmala: i agree that we shouldn't force to use the same arn for application and the image.

fletcherw:
/release-note-edit
```release-note
Changed ecr-credential-provider to fall back to using default credentials when a ServiceAccountToken is provided but there is no corresponding role ARN provided to assume. This should not cause changes for users using the default credential provider config from EKS AMIs.
```

PR #1294, opened 2025-11-06
Retornam/golang

retornam:
**What type of PR is this?**
>
> /kind api-change
 /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR updates Go version from 1.24.4 to 1.24.9 to pickup updates to the stdlib and also these fixes 
- [CVE-2025-58186](https://www.wiz.io/vulnerability-database/cve/cve-2025-58186) which affects `net/http`
- [CVE-2025-58187](https://www.wiz.io/vulnerability-database/cve/cve-2025-47912) which afftects `net/url`

N/A

```release-note
Update Go version to pick up stdblib updates and fixes to  CVE-2025-58186 and CVE-2025-58187
```
cc @cartermckinnon

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module k8s.io/cloud-provider-aws
 
-go 1.24.7
+go 1.24
+
+toolchain go1.24.9
 
 require (
 	github.com/aws/aws-sdk-go-v2 v1.36.5
--- a/.github/workflows/deps.yml
+++ b/.github/workflows/deps.yml
@@ -42,10 +42,10 @@ jobs:
       - id: govulncheck
         uses: golang/govulncheck-action@v1
         with:
-          go-version-input: 1.24.4
+          go-version-input: 1.24.9
           go-version-file: go.mod
       - id: govulncheck-tests-e2e
         uses: golang/govulncheck-action@v1
         with:
-          go-version-input: 1.24.4
+          go-version-input: 1.24.9
           go-version-file: tests/e2e/go.mod
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,1 +1,1 @@
-defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.4.0-go1.24.7-bookworm.0
+defaultBaseImage: registry.k8s.io/build-image/go-runner:v2.4.0-go1.24.9-bookworm.0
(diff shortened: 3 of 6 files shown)


version.txt
```diff
@@ -1 +1 @@
-1.30.9
+1.30.10
```
cartermckinnon: This should go in a separate PR

retornam: ok

retornam:
> /ok-to-test

@cartermckinnon fixed and pushed, can you please comment ok-to-test again, thanks.

go.mod
```diff
@@ -1,6 +1,6 @@
 module k8s.io/cloud-provider-aws
 
-go 1.24.7
+go 1.24.9
```
kmala: we stopped updating the patch version in the master branch and want to follow the same in release branches https://github.com/kubernetes/cloud-provider-aws/pull/1282
Can we change this to 1.24.0 ?

retornam: @kmala The patch/fix for the CVEs will not be in version 1.24.0. An alternative would be to upgrade to version 1.25.0, which would introduce further changes. I would appreciate an explanation for the decision to cease updating patch versions.

kmala: the directive here doesn't determine which version of go to build. when we don't mention the patch version then we can build with any patch version. For fixing CVE's dockerfile changes are good as that is updated to use 1.24.9 and is used for official binary.

retornam: > the directive here doesn't determine which version of go to build. when we don't mention the patch version then we can build with any patch version. For fixing CVE's dockerfile changes are good as that is updated to use 1.24.9 and is used for official binary.


Oh, that makes sense. I thought you wanted `1.24.0` everywhere, not just in `go.mod`.

To make this crystal clear for future contributors,  how about we keep the module’s minimum semantics at Go  `1.xy` (`1.24` in this instance)  , and pin the exact compiler we build with to go `1.xy.z` (`1.24.9` in this instance) for CVE fixes and reproducibility:

```
go 1.24
toolchain go1.24.9
```

This means any `1.24.z` toolchain satisfies the module’s minimum, but our CI/Docker builds use Go `1.24.9` specifically.

retornam: @kmala added the toolchain and then modified the go directive to 1.24. Let me know if this satisfies the requirement so I can resolve this conversation. Thanks

PR #1313, opened 2025-12-02
WIP: Add support for dual stack load balancers

### LINKED ISSUE #1219: Dual-stack support for Service type-loadBalancer NLB

Support for dual-stack Network Load Balancers (NLB) within the Kubernetes Cloud Controller Manager (CCM) for AWS. This enhancement would enable the provisioning of NLBs with both IPv4 and IPv6 addresses (dual-stack) for Kubernetes Services of type LoadBalancer.

The AWS Load Balancer Controller (LBC) provides an annotation for IP Address Type [`alb.ingress.kubernetes.io/ip-address-type`](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/annotations/#ip-address-type) with the valid values: `ipv4` or `dualstack`. It would be nice to have similar experience.

**Why is this needed**:

AWS officially announced IPv6 support for Network Load Balancers (NLB) in 2020, as highlighted in the provided announcement. This capability is increasingly vital for modern cloud-native deployments that require dual-stack networking, allowing services to be accessible via both IPv4 and IPv6 addresses. https://www.amazonaws.cn/en/new/2020/network-load-balancer-now-supports-ipv6/

There has been a consistent demand and prior efforts from the community to integrate this functionality, as evidenced by existing issues (#477, #243) and a past pull request (#497). Implementing dual-stack NLB support in cloud-provider-aws would align Kubernetes' load balancer provisioning with AWS's native features, offering a comprehensive solution for users aiming to deploy IPv6-enabled services. This enhancement would significantly improve network flexibility, and facilitate adherence to IPv6 mandates, enabling Kubernetes services to fully leverage AWS's dual-stack networking capabilities.

/kind feature

nrb:
**What type of PR is this?**
/kind feature

Adds basic support for configuring the IP address family on Services.

This functionality overlaps with the ALBC somewhat, but the goal here is to provide much more basic functionality, not the full scope of ALBC.

Fixes #1219 

```release-note
Support for the Kubernetes Service dual-stack networking API.
```

### DIFF
--- a/docs/service_controller.md
+++ b/docs/service_controller.md
@@ -86,3 +86,65 @@ metadata:
     service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: preserve_client_ip.enabled=false,proxy_protocol_v2.enabled=true
 [...]
 ```
+
+## Dual stack services with IPv6
+
+Services can be created using solely IPv4 networking (the default), or with dual stack support per the [Kubernetes Service specification](https://kubernetes.io/docs/concepts/services-networking/dual-stack/).
+The service must be created with a Network Load Balancer, and the Kubernetes control plane must be configured to support IPv6 CIDRs.
+
+Note: When using the [AWS Load Balancer Controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/), Services will default to having the `spec.loadBalancerClass` field populated via a MutatingWebhookConfiguration.
+This webhook must be disabled to allow the cloud controller manager to handle services.
+
+Some limitations to be aware of when using dual stack load balancers:
+
+- The `spec.ipFamilies` field can have a second family added or removed, but the first entry is immutable after Service creation.
+- Load balanced targets are registered based on the instances, not their IP addresses.
+- A Service cannot be IPv6 only; it must either be IPv4 or dual stack, even if IPv6 is the only IP family specified.
+
+### Usage Example 1 - creating a dual stack load balancer, requiring both stacks
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: $SVC_NAME
+  namespace: ${APP_NAMESPACE}
+  annotations:
+    service.beta.kubernetes.io/aws-load-balancer-type: nlb
+spec:
+    type: LoadBalancer
+    ipFamilies:
+        - IPv6
+        - IPv4
+    ipFamilyPolicy: RequireDualStack # Require both stacks are present on the service.
+    selector:
+        app: myapp
+    ports:
+        - port: 80
+          targetPort: 8080
+          protocol: TCP
+```
+
+### Usage Example 2 - creating a dual stack load balancer, falling back to IPv4
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: $SVC_NAME
+  namespace: ${APP_NAMESPACE}
+  annotations:
+    service.beta.kubernetes.io/aws-load-balancer-type: nlb
+spec:
+    type: LoadBalancer
+    ipFamilies:
+        - IPv4
+        - IPv6
+    ipFamilyPolicy: PreferDualStack # If dual stack is not configured or present, fall back to IPv4.
+    selector:
+        app: myapp
+    ports:
+        - port: 80
+          targetPort: 8080
+          protocol: TCP
+```
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -2009,15 +2009,24 @@ func (c *Cloud) createSecurityGroupRules(ctx context.Context, sgID string, rules
 	if len(sgID) == 0 {
 		return fmt.Errorf("security group ID cannot be empty")
 	}
-	// Allow ICMP fragmentation packets, important for MTU discovery
-	permission := ec2types.IpPermission{
-		IpProtocol: aws.String("icmp"),
-		FromPort:   aws.Int32(3),
-		ToPort:     aws.Int32(4),
-		IpRanges:   ec2SourceRanges,
-		Ipv6Ranges: ec2Ipv6SourceRanges,
+	// Allow ICMP fragmentation packets, important for IPv4 MTU discovery
+	if len(ec2SourceRanges) > 0 {
+		rules.Insert(ec2types.IpPermission{
+			IpProtocol: aws.String("icmp"),
+			FromPort:   aws.Int32(3),
+			ToPort:     aws.Int32(4),
+			IpRanges:   ec2SourceRanges,
+		})
 	}
-	rules.Insert(permission)
+	// Allow ICMPv6 "Packet Too Big" messages, important for IPv6 MTU discovery
+	if len(ec2Ipv6SourceRanges) > 0 {
+		rules.Insert(ec2types.IpPermission{
+			IpProtocol: aws.String("icmpv6"),
+			FromPort:   aws.Int32(2),
+			ToPort:     aws.Int32(-1),
+			Ipv6Ranges: ec2Ipv6SourceRanges,
+		})
+	}
 
 	// Setup ingress rules
 	if _, err := c.setSecurityGroupIngress(ctx, sgID, rules); err != nil {
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -2001,10 +2001,11 @@ func (c *Cloud) buildELBSecurityGroupList(ctx context.Context, serviceName types
 //   - sgID: The ID of the security group to configure.
 //   - rules: An existing permission set of rules to be added to the security group.
 //   - ec2SourceRanges: A slice of *ec2.IpRange objects specifying the source IP ranges for the rules.
+//   - ec2Ipv6SourceRanges: A slice of *ec2.Ipv6Range objects specifying the source IPv6 ranges for the rules.
 //
 // Returns:
 //   - error: An error if any issue occurs while creating or applying the security group rules.
-func (c *Cloud) createSecurityGroupRules(ctx context.Context, sgID string, rules IPPermissionSet, ec2SourceRanges []ec2types.IpRange) error {
+func (c *Cloud) createSecurityGroupRules(ctx context.Context, sgID string, rules IPPermissionSet, ec2SourceRanges []ec2types.IpRange, ec2Ipv6SourceRanges []ec2types.Ipv6Range) error {
 	if len(sgID) == 0 {
 		return fmt.Errorf("security group ID cannot be empty")
 	}
@@ -2014,6 +2015,7 @@ func (c *Cloud) createSecurityGroupRules(ctx context.Context, sgID string, rules
 		FromPort:   aws.Int32(3),
 		ToPort:     aws.Int32(4),
 		IpRanges:   ec2SourceRanges,
+		Ipv6Ranges: ec2Ipv6SourceRanges,
 	}
 	rules.Insert(permission)
 
@@ -2116,7 +2118,16 @@ func (c *Cloud) getSubnetCidrs(ctx context.Context, subnetIDs []string) ([]strin
 
 	cidrs := make([]string, 0, len(subnets))
 	for _, subnet := range subnets {
+		// Add IPv4 CIDR
 		cidrs = append(cidrs, aws.ToString(subnet.CidrBlock))
+
+		// Add IPv6 CIDRs if present
+		for _, ipv6Association := range subnet.Ipv6CidrBlockAssociationSet {
+			if ipv6Association.Ipv6CidrBlockState != nil &&
+				ipv6Association.Ipv6CidrBlockState.State == ec2types.SubnetCidrBlockStateCodeAssociated {
+				cidrs = append(cidrs, aws.ToString(ipv6Association.Ipv6CidrBlock))
+			}
+		}
 	}
 	return cidrs, nil
 }
@@ -2278,6 +2289,32 @@ func (c *Cloud) ensureNLBSecurityGroup(ctx context.Context, loadBalancerName, cl
 	return []string{securityGroupID}, nil
 }
 
+// separateIPv4AndIPv6CIDRs separates a list of CIDR strings into IPv4 and IPv6 ranges
+// Returns EC2 IpRange and Ipv6Range slices for use in security group rules
+func separateIPv4AndIPv6CIDRs(cidrs []string) ([]ec2types.IpRange, []ec2types.Ipv6Range) {
+	var ipv4Ranges []ec2types.IpRange
+	var ipv6Ranges []ec2types.Ipv6Range
+
+	for _, cidr := range cidrs {
+		_, ipNet, err := net.ParseCIDR(cidr)
+		if err != nil {
+			klog.Warningf("Failed to parse CIDR %q: %v", cidr, err)
+			continue
+		}
+
+		// Check if this is an IPv4 or IPv6 CIDR
+		if ipNet.IP.To4() != nil {
+			// IPv4
+			ipv4Ranges = append(ipv4Ranges, ec2types.IpRange{CidrIp: aws.String(cidr)})
+		} else {
+			// IPv6
+			ipv6Ranges = append(ipv6Ranges, ec2types.Ipv6Range{CidrIpv6: aws.String(cidr)})
+		}
+	}
+
+	return ipv4Ranges, ipv6Ranges
+}
+
 // ensureNLBSecurityGroupRules ensures the NLB frontend security group rules are created and configured
 // for the specified security groups based on the load balancer port mappings (Load Balancer listeners),
 // allowing traffic from the specified source ranges.
@@ -2285,27 +2322,41 @@ func (c *Cloud) ensureNLBSecurityGroup(ctx context.Context, loadBalancerName, cl
 // Parameters:
 //   - ctx: The context for the request.
 //   - securityGroups: The security group IDs to configure rules for (only first SG is used).
-//   - ec2SourceRanges: The CIDR ranges allowed to access the load balancer.
+//   - sourceCIDRs: The CIDR ranges (IPv4 and/or IPv6) allowed to access the load balancer.
 //   - v2Mappings: The NLB port mappings defining frontend ports and protocols.
 //
 // Returns:
 //   - error: An error if any issue occurs while ensuring the NLB security group rules.
-func (c *Cloud) ensureNLBSecurityGroupRules(ctx context.Context, securityGroups []string, ec2SourceRanges []ec2types.IpRange, v2Mappings []nlbPortMapping) error {
+func (c *Cloud) ensureNLBSecurityGroupRules(ctx context.Context, securityGroups []string, sourceCIDRs []string, v2Mappings []nlbPortMapping) error {
 	if len(securityGroups) == 0 {
 		return nil
 	}
 	securityGroupID := securityGroups[0]
 
+	// Separate source CIDRs into IPv4 and IPv6 ranges
+	ec2SourceRanges, ec2Ipv6SourceRanges := separateIPv4AndIPv6CIDRs(sourceCIDRs)
+
 	ingressRules := NewIPPermissionSet()
 	for _, mapping := range v2Mappings {
-		ingressRules.Insert(ec2types.IpPermission{
+		permission := ec2types.IpPermission{
 			FromPort:   aws.Int32(int32(mapping.FrontendPort)),
 			ToPort:     aws.Int32(int32(mapping.FrontendPort)),
 			IpProtocol: aws.String(strings.ToLower(string((mapping.FrontendProtocol)))),
-			IpRanges:   ec2SourceRanges,
-		})
+		}
+
+		// Add IPv4 ranges if present
+		if len(ec2SourceRanges) > 0 {
+			permission.IpRanges = ec2SourceRanges
+		}
+
+		// Add IPv6 ranges if present
+		if len(ec2Ipv6SourceRanges) > 0 {
+			permission.Ipv6Ranges = ec2Ipv6SourceRanges
+		}
+
+		ingressRules.Insert(permission)
 	}
-	if err := c.createSecurityGroupRules(ctx, securityGroupID, ingressRules, ec2SourceRanges); err != nil {
+	if err := c.createSecurityGroupRules(ctx, securityGroupID, ingressRules, ec2SourceRanges, ec2Ipv6SourceRanges); err != nil {
 		return fmt.Errorf("error while updating rules to security group %q: %w", securityGroupID, err)
 	}
 	return nil
@@ -2328,6 +2379,10 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 		return nil, err
 	}
 
+	if !isNLB(annotations) && serviceRequestsIPv6(apiService) {
+		return nil, fmt.Errorf("classic load balancer for service %s does not support IPv6", apiService.Name)
+	}
+
 	if apiService.Spec.SessionAffinity != v1.ServiceAffinityNone {
 		// ELB supports sticky sessions, but only when configured for HTTP/HTTPS
 		return nil, fmt.Errorf("unsupported load balancer affinity: %v", apiService.Spec.SessionAffinity)
@@ -2348,9 +2403,18 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 	if err != nil {
 		return nil, err
 	}
-	ec2SourceRanges := []ec2types.IpRange{}
-	for _, srcRange := range sourceRanges.StringSlice() {
-		ec2SourceRanges = append(ec2SourceRanges, ec2types.IpRange{CidrIp: aws.String(srcRange)})
+
+	sourceCIDRs := sourceRanges.StringSlice()
+
+	// If no source ranges specified, add defaults based on service IP families
+	// This should be populated by GetLoadBalancerSourceRanges most of the time.
+	if len(sourceCIDRs) == 0 {
+		sourceCIDRs = append(sourceCIDRs, "0.0.0.0/0")
+	}
+
+	// Add IPv6 default range if service supports IPv6.
+	if serviceRequestsIPv6(apiService) && !contains(sourceCIDRs, "::/0") {
+		sourceCIDRs = append(sourceCIDRs, "::/0")
 	}
 
 	sslPorts := getPortSets(annotations[ServiceAnnotationLoadBalancerSSLPorts])
@@ -2452,13 +2516,14 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 			internalELB,
 			annotations,
 			securityGroups,
+			apiService,
 		)
 		if err != nil {
 			return nil, err
 		}
 
 		// Ensure SG rules only if the LB reconciliator finished successfully.
-		if err := c.ensureNLBSecurityGroupRules(ctx, securityGroups, ec2SourceRanges, v2Mappings); err != nil {
+		if err := c.ensureNLBSecurityGroupRules(ctx, securityGroups, sourceCIDRs, v2Mappings); err != nil {
 			return nil, fmt.Errorf("error ensuring NLB security group rules: %w", err)
 		}
 
@@ -2483,6 +2548,10 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 		}
 		if len(sourceRangeCidrs) == 0 {
 			sourceRangeCidrs = append(sourceRangeCidrs, "0.0.0.0/0")
+			// Add IPv6 default range if service supports IPv6
+		}
+		if serviceRequestsIPv6(apiService) && !contains(sourceRangeCidrs, "::/0") {
+			sourceRangeCidrs = append(sourceRangeCidrs, "::/0")
 		}
 
 		err = c.updateInstanceSecurityGroupsForNLB(ctx, loadBalancerName, instances, subnetCidrs, sourceRangeCidrs, v2Mappings)
@@ -2629,6 +2698,9 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 	}
 
 	if setupSg {
+		// Separate source CIDRs into IPv4 and IPv6 ranges for classic ELB
+		ec2SourceRanges, ec2Ipv6SourceRanges := separateIPv4AndIPv6CIDRs(sourceCIDRs)
+
 		permissions := NewIPPermissionSet()
 		for _, port := range apiService.Spec.Ports {
 			protocol := strings.ToLower(string(port.Protocol))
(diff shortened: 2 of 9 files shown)


damdo:
Hey @kmala do we know what's happening with the Netlify checks? It looks like they are broken across many open PRs. TY

nrb:

One more run to see if this is a rate limiting issue or some sort of genuine issue with my code.

pkg/providers/v1/aws_validations_test.go
```diff
@@ -414,6 +414,16 @@ func TestValidateServiceAnnotations(t *testing.T) {
 			expectedError: "target group attributes annotation is only supported for NLB",
 		},
 
+		// IPFamily annotation set to dualstack (should succeed)
+		{
+			name: "NLB with dualstack annotation - success",
```
tthvo: Should we have a case "CLB with dualstack annotation" that expects a validation error?

pkg/providers/v1/aws.go
```diff
@@ -234,6 +234,11 @@ const ServiceAnnotationLoadBalancerEIPAllocations = "service.beta.kubernetes.io/
 // static IP addresses for the NLB. Only supported on elbv2 (NLB)
 const ServiceAnnotationLoadBalancerPrivateIPv4Addresses = "service.beta.kubernetes.io/aws-load-balancer-private-ipv4-addresses"
 
+// ServiceAnnotationLoadBalancerIPAddressType is the annotation used on the service
+// to specify the IP address type for the load balancer. Supported values are "ipv4" and "dualstack".
+// Defaults to "ipv4". Only supported on NLB.
+const ServiceAnnotationLoadBalancerIPAddressType = "service.beta.kubernetes.io/aws-load-balancer-ip-address-type"
```
tthvo: Wondering if we should add support for configuring Target Group IP type (i.e. `ipv4` and `ipv6`)? Currently, the target groups are always created as `ipv4` only regardless of the IP type of Load Balancer.

My thought is that we also want to support dual-stack with IPv6 as primary. In this case, the target group should preferably be `ipv6`. WDYT?

tthvo (commented):
I tested with a simple service below:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: wordpress-mysql
  labels:
    app: wordpress
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-ip-address-type: dualstack
spec:
  ipFamilyPolicy: PreferDualStack
  ports:
    - port: 3306
  selector:
    app: wordpress
  type: LoadBalancer
```

The CCM does create the dualstack NLB as expected 🚀  

```bash
$ kubectl -n myns get svc/wordpress-mysql -o yaml | yq .status.loadBalancer.ingress
- hostname: <id>.elb.us-east-1.amazonaws.com

$ nslookup <id>.elb.us-east-1.amazonaws.com
# nslookup shows both public IPv4 EIP and IPv6 GUA addresses

$ aws elbv2 describe-load-balancers --names=<nlb-id> --query='LoadBalancers[0].[Type,IpAddressType,Scheme]'
[
    "network",
    "dualstack",
    "internet-facing"
]
```

pkg/providers/v1/aws.go
nrb: Yes, we should definitely do this. I was thinking of doing this in a separate PR in order to keep the changes more manageable. Do you think it would be better as a commit within this one?

tthvo: Oh sure, sorry I didn't realize you planned to make a follow-up PR :sweat_smile:  Let's keep this PR like this.

nrb:
@kmala I'm going to do some manual testing with this PR starting today before I mark it as ready for review, but would you be able to give it a look by any chance?

nrb:
Tagging you all because we've talked at various points.

@tthvo (also working on dual stack enablement in Cluster API Provider AWS)
@mtulio (also helping with dual stack tasks)
@kmala (AWS CCM maintainer)
@alebedev87 (ingress controller SME)

Summing up some of the various discussions. Please correct me if any of these things are wrong.

- https://kubernetes.io/docs/concepts/services-networking/dual-stack/ talks about an extension to the `Service` resource to specify IP family information. For our purposes, it allows us to specify...
  - Precedence of IPv4 or IPv6 addresses depending on their order in the `spec.ipFamilies` field
  - How to handle failure on dual stack LBs - fall back to IPv4/single stack, or fail creation call
- What this API _does not_ give us, compared to using annotations
  - No way to specify attachment type, whether `ip` or `instance`; that is, what backend the LB will route to - the instance, even with a new IP, or the IP, even if it's a new instance.
  - Difficult or unclear on how to specify the IP family on a target group; a dual stack LB could use an IPv4 target group _or_ an IPv6 target group, but not both. As-is, we'd need to infer it based on the order of `spec.ipFamilies`.
  - Similar issues with security groups; no way to specify them, but we could infer based on `spec.ipFamilies`

I can see us proceeding in a few ways here.

1) Keep the API in this repo fairly simple; infer information based on `spec.ipFamilies` and `spec.ipFamilyPolicy` fields, and do not use any annotations
   - Security/target groups will be created based on whichever IP family is first in `spec.ipFamilies`
   - Use `instance` attachment since that's how the CCM has behaved so far.
   - This aligns us with the upstream API and keeps the implementation simple, but at the cost of reduced flexibility
2) Use only annotations to specify this information, the same way AWS Load Balancer Controller does. Use the `Service.spec.LoadBalancerClass` to determine if the CCM should process the Service or the ALBC should.
   - This is beneficial because the user or higher level component can explicitly and fully communicate settings.
   - This might be challenging with the Gateway API, which can only [propagate 8 annotations](https://gateway-api.sigs.k8s.io/reference/spec/#gatewayinfrastructure)
   - Maximal flexibility, but annotations as an API is difficult to fully validate and may not be fully supported by higher-level controllers.
3) Use a combination of the Service spec and annotations; `service.spec.ipFamilies` and `service.spec.ipFamilyPolicy` would be used for their respective purposes, and further settings, like security/target groups or attachment type, are specified via annotations.
   - Combining both approaches allows us to present a small required API, but still offer customization for users that need it.
   - This includes the downsides of both approaches, plus it adds mental overhead in understanding what information goes where.
  
In any of the 3 approaches here, I'd like to keep the number of total fields to the absolute minimum for supporting dual stack on load balancers. If users need something beyond that, they should opt for the ALBC.

That minimum would be:
 - Fallback behavior (`Service.spec.ipFamilyPolicy` fulfills this)
 - Supported IP families and priority ('Service.spec.ipFamilies' fulfills this)
 - Target group IP family (could be inferred/defaulted)
 - Security group IP family (could be inferred/defaulted)
 - Instance attachment type (could be inferred/defaulted)
 
I'd appreciate your thoughts here - which of these options seems more workable to you?

pkg/providers/v1/aws.go
```diff
@@ -234,6 +234,16 @@ const ServiceAnnotationLoadBalancerEIPAllocations = "service.beta.kubernetes.io/
...
+
+// ServiceAnnotationLoadBalancerTargetGroupIPAddressType is the annotation used on the service
+// to specify the IP address type for the target groups. Supported values are "ipv4" and "ipv6".
+// Defaults to "ipv4". Only supported on NLB.
+const ServiceAnnotationLoadBalancerTargetGroupIPAddressType = "service.beta.kubernetes.io/aws-load-balancer-target-group-ip-address-type"
+
```
mtulio: Please also update the service controller documentation: https://github.com/kubernetes/cloud-provider-aws/blob/master/docs/service_controller.md

pkg/providers/v1/aws.go
```diff
@@ -2483,6 +2512,12 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 		}
 		if len(sourceRangeCidrs) == 0 {
 			sourceRangeCidrs = append(sourceRangeCidrs, "0.0.0.0/0")
+
+			// For dual-stack or IPv6 load balancers, also add IPv6 default route
```
mtulio: route or rule?
```suggestion
			// For dual-stack or IPv6 load balancers, also add IPv6 default rule
```

pkg/providers/v1/aws_validations.go
```diff
@@ -98,6 +103,64 @@ func validateServiceAnnotations(v *awsValidationInput) error {
...
+	}
+	return nil
+}
+
+// validateIPFamilyPolicy validates the ipFamilyPolicy and ipFamilies fields in the Service spec.
+func validateIPFamilyPolicy(v *awsValidationInput) error {
+	isNLB := isNLB(v.annotations)
+
+	// Validate ipFamilyPolicy and ipFamilies
+	if v.apiService.Spec.IPFamilyPolicy != nil {
```
mtulio: nit: what about return early when `IPFamilyPolicy==nil`?

pkg/providers/v1/aws_validations.go
```diff
@@ -98,6 +103,64 @@ func validateServiceAnnotations(v *awsValidationInput) error {
...
+		switch *v.apiService.Spec.IPFamilyPolicy {
+		case v1.IPFamilyPolicyRequireDualStack:
+			// RequireDualStack is only supported for NLB
+			if !isNLB {
```
mtulio: this was checked in the outer function, or was that added to increase unit coverage? if so, maybe returning early too?

pkg/providers/v1/aws_validations.go
```diff
@@ -98,6 +103,64 @@ func validateServiceAnnotations(v *awsValidationInput) error {
...
+		if !isNLB {
+			return fmt.Errorf("ip address type annotation is only supported for NLB")
+		}
+	}
+
+	if _, present := v.annotations[ServiceAnnotationLoadBalancerTargetGroupIPAddressType]; present {
+		if !isNLB {
+			return fmt.Errorf("target group ip address type annotation is only supported for NLB")
+		}
+	}
```
mtulio: do we need to validate valid values were set here?

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -94,6 +95,71 @@ func isLBExternal(annotations map[string]string) bool {
...
+// Priority order: annotations > spec fields > default (ipv4)
+func (c *Cloud) getLBIPAddressType(service *v1.Service) elbv2types.IpAddressType {
+	// Check annotation first for backward compatibility
+	if ipType, ok := service.Annotations[ServiceAnnotationLoadBalancerIPAddressType]; ok {
+		klog.V(4).Infof("Using load balancer IP address type from annotation: %s", ipType)
+		return elbv2types.IpAddressType(ipType)
+	}
+
+	// Fall back to spec.ipFamilies and spec.ipFamilyPolicy
+	if service.Spec.IPFamilyPolicy != nil {
```
mtulio: what about return early here with `elbv2types.IpAddressTypeIpv4`?

pkg/providers/v1/aws_validations.go
```diff
@@ -98,6 +103,64 @@ func validateServiceAnnotations(v *awsValidationInput) error {
 			return err
 		}
 	}
+
+	if _, present := v.annotations[ServiceAnnotationLoadBalancerIPAddressType]; present {
+		if !isNLB {
+			return fmt.Errorf("ip address type annotation is only supported for NLB")
+		}
```
mtulio: Considering this statement from [AWS docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-register-targets.html#target-security-groups):

> Also, with dualstack Network Load Balancers, client IP address preservation does not work when translating IPv4 addresses to IPv6, or IPv6 to IPv4 addresses. Client IP address preservation only works when client and target IP addresses are both IPv4 or both IPv6.

Do we need to add some validation when the preserve ip is set?

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -886,13 +968,48 @@ func (c *Cloud) ensureTargetGroupTargets(ctx context.Context, tgARN string, expe
...
+	}
+
+	// Check network interfaces for IPv6 addresses
+	for _, networkInterface := range instance.NetworkInterfaces {
+		if networkInterface.Status != ec2types.NetworkInterfaceStatusInUse {
+			continue
+		}
+		if len(networkInterface.Ipv6Addresses) > 0 {
+			// Return the first IPv6 address
+			return aws.ToString(networkInterface.Ipv6Addresses[0].Ipv6Address)
```
mtulio: In order to [satisfy the requirement](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-register-targets.html#register-targets) of the target ip family ipv6, I believe you need to verify if the Ipv6Address is primary (`IsPrimaryIpv6: true`).

Thoughts @tthvo @nrb ?

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_NetworkInterfaceIpv6Address.html

> When registering targets by instance ID for a IPv6 target group, the targets must have an assigned primary IPv6 address. To learn more, see [IPv6 addresses](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html#ipv6-addressing) in the Amazon EC2 User Guide

> When registering targets by IP address for an IPv6 target group, the IP addresses that you register must be within the VPC IPv6 CIDR block or within the IPv6 CIDR block of a peered VPC.

mtulio (commented):
Great work! A few nits, and questions related to the restrictions on target ipv6.

pkg/providers/v1/aws_loadbalancer.go
tthvo: Hey @mtulio That's right, but the primary IPv6 requirement actually depends on the [target type](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#target-type) of the Target Group. In our case, the following type is applicable:

- `instance`: The instance must be in the same VPC as NLB and is assigned a "stable" IPv6 address. Unlike IPv4, IPv6 address, by default, can be re-assigned elsewhere. Thus, we need to set `IsPrimaryIpv6: true` to make sure the IPv6 address is tied to the ENI of the instance for its entire lifecycle. I guess "primary" on AWS means "stable/owned".
  > When registering targets by instance ID for a IPv6 target group, the targets must have an assigned primary IPv6 address. To learn more, see [IPv6 addresses](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html#ipv6-addressing) in the Amazon EC2 User Guide

- `IP`: There is no requirement for a "stable" IPv6 address (or setting flag `IsPrimaryIpv6`). But the IPv6 address must be within the VPC IPv6 CIDR of the cluster.
  > When registering targets by IP address for an IPv6 target group, the IP addresses that you register must be within the VPC IPv6 CIDR block or within the IPv6 CIDR block of a peered VPC.

My understanding is that we should prefer `instanceID` target type in order to ensure traffic will always reach the intended target (i.e. cluster nodes). Registering with IP address will pin the traffic to the IP, which may not be the cluster node at all as it can be re-assigned to other services, for example, another non-cluster EC2 instance.

@nrb IIUC, we should make the CCM prefer registering by `instance` when, let's say, primary IPv6 is available. Otherwise, fall back to register by the IP address?

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -784,6 +787,16 @@ func (c *Cloud) ensureTargetGroup(ctx context.Context, targetGroup *elbv2types.T
 			input.HealthCheckPath = aws.String(mapping.HealthCheckConfig.Path)
 		}
 
+		// Set IP address type based on annotation
+		if tgIPAddressType, ok := annotations[ServiceAnnotationLoadBalancerTargetGroupIPAddressType]; ok {
+			if tgIPAddressType == string(elbv2types.TargetGroupIpAddressTypeEnumIpv6) || tgIPAddressType == string(elbv2types.TargetGroupIpAddressTypeEnumIpv4) {
```
sadasu: Can we document why we can have only "ipv4" or "ipv6" target groups and not "[ipv4,ipv6]" for this annotation esp when annotationLBIPAddressType is "dualstack"?

pkg/providers/v1/aws.go
sadasu: Route.  "0.0.0.0/0" and "::/0" act as a gateway of last resort, directing unmatched traffic to a specific next-hop or upstream router.

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -1111,23 +1112,49 @@ func (c *Cloud) updateInstanceSecurityGroupsForNLB(ctx context.Context, lbName s
...
+
 // updateInstanceSecurityGroupForNLBTraffic will manage permissions set(identified by ruleDesc) on securityGroup to match desired set(allow protocol traffic from ports/cidr).
 // Note: sgPerms will be updated to reflect the current permission set on SG after update.
 func (c *Cloud) updateInstanceSecurityGroupForNLBTraffic(ctx context.Context, sgID string, sgPerms IPPermissionSet, ruleDesc string, protocol string, ports sets.Set[int32], cidrs []string) error {
 	desiredPerms := NewIPPermissionSet()
 	for port := range ports {
 		for _, cidr := range cidrs {
-			desiredPerms.Insert(ec2types.IpPermission{
+			perm := ec2types.IpPermission{
 				IpProtocol: aws.String(protocol),
```
sadasu, line 1154: Should IpProtocol be set below based on result of isIPv6CIDR(cidr)?

pkg/providers/v1/aws_loadbalancer.go
sadasu: Also, do we have to make sure that when annotation and spec fields are both present, they match?

sadasu (commented):
@nrb Thanks for making these updates. A few comments inline.

nrb:
I've updated the change to only use the Kubernetes API fields in `service.Spec` rather than annotations. I think this keeps the implementation straightforward, and there should be nothing in this PR that blocks us from adding support for more granular settings via annotations in the future.

nrb:
govulncheck notice will be resolved by pulling https://github.com/kubernetes/kubernetes/pull/136820 in during a kube bumpl.

pkg/providers/v1/aws.go
tthvo, line 2371: I think we should also pass in the `ec2IPv6SourceRanges` here to `createSecurityGroupRules` so that we can create the `icmpv6` protocol rule, right? 

https://github.com/kubernetes/cloud-provider-aws/blob/b6c9d76f366ecc10a83de1771630e917d9b25660/pkg/providers/v1/aws.go#L2011-L2018

pkg/providers/v1/aws_loadbalancer.go
tthvo, line 214: I guess we need to configure the `IpAddressType` according to service `ipFamilies`, right? We can do it similarly to target group IP type.

```go
func getLoadBalancerIPAddressTypeFromService(service *v1.Service) elbv2types.IpAddressType {
    if serviceRequestsIPv6(service) {
        return elbv2types.IpAddressTypeDualstack
    }
    return elbv2types.IpAddressTypeIpv4
}
```

And a unit test to cover the ip type would be great too!

pkg/providers/v1/aws.go
```diff
@@ -2452,13 +2505,14 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
 			internalELB,
 			annotations,
 			securityGroups,
+			apiService,
```
kmala, line 2529: nit: if we are passing the entire service then we can skip passing other params like service name, annotation right?

pkg/providers/v1/aws.go
```diff
@@ -3221,6 +3275,11 @@ func (c *Cloud) buildSecurityGroupsToDelete(ctx context.Context, service *v1.Ser
 
 // EnsureLoadBalancerDeleted implements LoadBalancer.EnsureLoadBalancerDeleted.
 func (c *Cloud) EnsureLoadBalancerDeleted(ctx context.Context, clusterName string, service *v1.Service) error {
+	// If spec.loadBalancerClass is set, another controller is responsible for this Service
+	if service.Spec.LoadBalancerClass != nil {
```
kmala: the condition is fine but any reason this is needed for dual stack support or are we just adding normally?

pkg/providers/v1/aws.go
tthvo, line 2571: We should consider/allow IPv6 CIDRs when updating the instance's security groups (to allow traffic from NLB), right?  IIUC, there are 2 types of CIDRs here:
- `sourceRangeCidrs`: Already defined on line 2396-2405
- `subnetCidrs`: IPv6 CIDRs can be extracted from subnet info returned by AWS.

tthvo: I guess that means we need to update the following functions to set the `Ipv6Range` field correctly?

https://github.com/kubernetes/cloud-provider-aws/blob/b6c9d76f366ecc10a83de1771630e917d9b25660/pkg/providers/v1/aws_loadbalancer.go#L957

https://github.com/kubernetes/cloud-provider-aws/blob/b6c9d76f366ecc10a83de1771630e917d9b25660/pkg/providers/v1/aws_loadbalancer.go#L1049

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -159,6 +183,16 @@ func (c *Cloud) ensureLoadBalancerv2(ctx context.Context, namespacedName types.N
...
 
+	// Determine target group IP address type based on Service spec.ipFamilies
+	targetGroupIPAddressType := getTargetGroupIPAddressTypeFromService(service)
+
+	// Validate that single stack IPv6 is not being used (not supported on NLB)
+	if service.Spec.IPFamilyPolicy != nil && *service.Spec.IPFamilyPolicy == v1.IPFamilyPolicySingleStack {
+		if serviceRequestsIPv6(service) {
+			return nil, fmt.Errorf("single stack IPv6 is not supported for network load balancers")
+		}
+	}
```
tthvo: I wonder if we can:
- Place this validation in pkg/providers/v1/aws_validations.go
- Additionally, we should add a validation that only allows dualstack settings (i.e. `ipFamilies: [IPv4, IPv6]`, `ipFamilyPolicy: RequireDualStack`) when LB type is NLB.

WDYT?

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -94,6 +95,29 @@ func isLBExternal(annotations map[string]string) bool {
 	return false
 }
 
+// getTargetGroupIPAddressTypeFromService determines the IP address type for the target group
+// based on the Service's spec.ipFamilies field. According to Kubernetes dual-stack documentation:
+// - If ipFamilies[0] is IPv6, the target group should use IPv6
+// - If ipFamilies is not set or ipFamilies[0] is IPv4, the target group should use IPv4 (default)
```
tthvo: From [k8s docs](https://kubernetes.io/docs/concepts/services-networking/dual-stack/#services), I guess the `ipFamilies` will never be empty as said 👇 

> The address family of a Service defaults to the address family of the first service cluster IP range (configured via the --service-cluster-ip-range flag to the kube-apiserver).

So, it's always going to be the primary IP family by default. I think the code is good, but maybe we can change the comment a bit?

docs/service_controller.md
```diff
@@ -85,4 +85,66 @@ metadata:
...
+
+Services can be created using solely IPv4 networking (the default), or with dual stack support per the [Kubernetes Service specification](https://kubernetes.io/docs/concepts/services-networking/dual-stack/).
+The service must be created with a Network Load Balancer, and the Kubernetes control plane must be configured to support IPv6 CIDRs.
+
+Note: When using the [AWS Load Balancer Controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/), Services will default to having the `spec.loadBalancerClass` field populated via a MutatingWebhookConfiguration.
+This webhook must be disabled to allow the cloud controller manager to handle services.
+
+Some limitations to be aware of when using dual stack load balancers:
+
+- The `spec.ipFamilies` field can have a second family added or removed, but the first entry is immutable after Service creation.
```
tthvo, line 100: Maybe, we should mention what would happen if the secondary family is removed? Like what will be the state of:
- The Load Balancer
- The Security Groups (those for LB and those for instances)
- Target Group
- (Optional) Listeners

docs/service_controller.md
```diff
@@ -85,4 +85,66 @@ metadata:
...
+metadata:
+  name: $SVC_NAME
+  namespace: ${APP_NAMESPACE}
+  annotations:
+    service.beta.kubernetes.io/aws-load-balancer-type: nlb
+spec:
+    type: LoadBalancer
+    ipFamilies:
+        - IPv6
+        - IPv4
```
tthvo, line 118: nit: The list indentation is a lot more than usual. Let's format it a bit 😁

tthvo (commented):
I just have a few questions 🙏  But overall, I think we are having great progress!

pkg/providers/v1/aws.go
tthvo: From the [docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/service/nlb/#configuration), the CCM should already ignore the service reconciling if `spec.loadBalancerClass` is non-nil. Perhaps, it is implemented elsewhere in a common place for all cloud providers (maybe [here](https://github.com/kubernetes/cloud-provider/blob/2ef5ab1cb2d946aff64918a8e43ae1da1ad5f98c/controllers/service/controller.go#L862-L865)).

I think this is not specific for dualstack support, but it's more of an additional safe-guard in case, right?

pkg/providers/v1/aws_loadbalancer.go
nrb: Yep, makes sense.

pkg/providers/v1/aws.go
nrb: Good catch on the `wantsLoadBalancer` function, wasn't aware of that.

@kmala this was mostly to be explicit about the contract around load balancer classes, but Thuan's pointed out the service controller handling that, so I'll remove this check.

pkg/providers/v1/aws_loadbalancer.go
nrb: Reading through https://github.com/kubernetes/cloud-provider/tree/2ef5ab1cb2d946aff64918a8e43ae1da1ad5f98c/controllers/service, I don't see where it is setting the ipFamilies field; I think it happens when routes are associated with nodes.

I think this code might be wrong if the user provides an IPv6 CIDR to `--service-cluster-ip-range` and then leaves `spec.IPFamilies` unset.

tthvo: Oh, I also saw the following spec description of `service.spec.ipFamilies` :point_down: 

```bash
$ kubectl explain svc.spec.ipFamilies
KIND:       Service
VERSION:    v1
FIELD: ipFamilies <[]string>
DESCRIPTION:
    IPFamilies is a list of IP families (e.g. IPv4, IPv6) assigned to this
    service. This field is usually assigned automatically based on cluster
    configuration and the ipFamilyPolicy field. If this field is specified
    manually, the requested family is available in the cluster, and
    ipFamilyPolicy allows it, it will be used; otherwise creation of the service
    will fail. This field is conditionally mutable: it allows for adding or
    removing a secondary IP family, but it does not allow changing the primary
    IP family of the Service. Valid values are "IPv4" and "IPv6".  This field
    only applies to Services of types ClusterIP, NodePort, and LoadBalancer, and
    does apply to "headless" services. This field will be wiped when updating a
    Service to type ExternalName.
    
    This field may hold a maximum of two entries (dual-stack families, in either
    order).  These families must correspond to the values of the clusterIPs
    field, if specified. Both clusterIPs and ipFamilies are governed by the
    ipFamilyPolicy field.
```

> IPFamilies is a list of IP families (e.g. IPv4, IPv6) assigned to this service. This field is usually assigned automatically based on cluster configuration and the ipFamilyPolicy field. If this field is specified manually, the requested family is available in the cluster, and ipFamilyPolicy allows it, it will be used; otherwise creation of the service will fail.

I guess the `IPFamilies` may be defaulted by some built-in k8s controllers and it will have at least 1 entry before the CCM starts to reconcile it...?

tthvo: > I think this code might be wrong if the user provides an IPv6 CIDR to --service-cluster-ip-range and then leaves spec.IPFamilies unset.

Right... I am not sure how we can handle it without knowing the configured service CIDRs of the cluster :disappointed: My hope is that https://github.com/kubernetes/cloud-provider-aws/pull/1313#discussion_r2856071476 is true. From my local testing with [`kinD`](https://kind.sigs.k8s.io/docs/user/configuration/#networking), the `ipFamilies` are always non-empty after creation :eyes: 

If not, I found CCM azure provider uses the same logic as yours: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/b0192b1b05f449d8b008a4b1519ff9324d527c1f/pkg/provider/azure_utils.go#L302-L315

tthvo: After doing a bit more research, I also found this [issue](https://github.com/kubernetes/kubernetes/issues/131779#issue-3065120072), which pointed me to this [servide default code](https://github.com/kubernetes/kubernetes/blob/761638907a5ebe7fc2e85b2e1a5a43d7356e2cec/pkg/registry/core/service/storage/storage.go#L232-L242) and [ipFamily defaulting code](https://github.com/kubernetes/kubernetes/blob/761638907a5ebe7fc2e85b2e1a5a43d7356e2cec/pkg/registry/core/service/storage/storage.go#L280). This does confirm `ipFamilies` are guaranteed to be non-empty, right?

Though, I believe there **is** a scenario where `ipFamilies` can be empty: **Upgrade from older kubernetes (< v1.20)**. But k8s < v1.20 only supports IPv4 anyway because dual-stack is available starting from v1.20. So, I think we are OK with the current approach?

nrb: Thanks for finding those! Thanks for bringing this up, I'll put a comment in mentioning it for clarity.

pkg/providers/v1/aws_loadbalancer.go
nrb: Ah, yes, we should support `Ipv6Ranges`, will update.

nrb:
Current e2e failure is legit - the provisioned cluster is not set up for IPv6.

```
   STEP: creating LoadBalancer service in Kubernetes @ 03/03/26 17:45:37.338
  I0303 17:45:37.386843 32011 loadbalancer.go:508] Unexpected error: 
      <*errors.errorString | 0xc000321140>: 
      failed to create LoadBalancer Service "lbconfig-test-clb-ds": Service "lbconfig-test-clb-ds" is invalid: spec.ipFamilies[1]: Invalid value: "IPv6": not configured on this cluster
      {
          s: "failed to create LoadBalancer Service \"lbconfig-test-clb-ds\": Service \"lbconfig-test-clb-ds\" is invalid: spec.ipFamilies[1]: Invalid value: \"IPv6\": not configured on this cluster",
      }
  [FAILED] in [It] - /home/prow/go/src/k8s.io/cloud-provider-aws/tests/e2e/loadbalancer.go:508 @ 03/03/26 17:45:37.387
  STEP: Destroying namespace "cloud-provider-aws-593" for this suite. @ 03/03/26 17:45:37.387
  << Timeline 
```

Will work on getting support for that plumbed in to kubetest-ec2.

nrb:
I've opened a PR against kubetest2-ec2 in order to provide an IPv6-capable subnet to CI.

https://github.com/kubernetes-sigs/provider-aws-test-infra/pull/530

PR #1209, opened 2025-07-15
Fix leak managed/owned security group on Service update with BYO SG on CLB

mtulio:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

This PR propose fix on leaked security group (SG) when a Service type-loadBalancer (CLB) is updated adding the BYO SG annotation (`service.beta.kubernetes.io/aws-load-balancer-security-groups`), which replaces all SG added to the Load Balancer without removing linked rules, as well not deleting managed SG (created by controller).

**Which issue(s) this PR fixes**:

Fixes #1208

**Special notes for your reviewer**:

We decided of creating isolated dedicated methods to discover and remove linked rule's SG targeting to:
- enhance code maintenance
- enhance unit tests
- allow to reuse the logic when NLB with SG is supported (future)

The unit tests and documentation(function) comments have been assisted by Cursor AI(model claude-4-sonet): [AIA HAb SeCeNc Hin R v1.0](https://aiattribution.github.io/statements/AIA-HAb-SeCeNc-Hin-R-v1.0)

**Does this PR introduce a user-facing change?**:

```release-note
Fixed security group leak when updating Service loadBalancer Classic Load Balancer (CLB) with `service.beta.kubernetes.io/aws-load-balancer-security-groups` annotation. Controller-managed security groups are now properly cleaned up when switching CLB from managed security groups to user-specified security groups.
```

PR #1326, opened 2026-01-05
feat: Implement support of BYO SG for Service NLB

mtulio:
**What type of PR is this?**

/kind feature

This proposal implements support of Bring Your Own Security Group (BYO SG) annotation to Service type-loadBalancer NLB.

The feature uses the same annotation of BYO SG to Classic Load Balancer, allowing users to create/update a Network Load Balancer with SGs attached to it.

Note: The update will work only when the NLB has been created with managed SG (TODO review this statement).

The BYOSG is an alternative of managed SG for NLBs, when the feature is enabled through cloud-config.

Fixes #

```release-note
Introduce support of Annotation `service.beta.kubernetes.io/aws-load-balancer-security-groups` on the Service for Network Load Balancer, so that users can provide their own security groups.
```


pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -1879,3 +1905,118 @@ func ValidateHealthCheck(s *elbtypes.HealthCheck) error {
...
+// groups by first removing all dependent ingress rules.
+//
+// Parameters:
+//   - ctx: The context for AWS API calls
+//   - sgID: The security group ID to search for in ingress rules
+//
+// Returns:
+//   - map[*ec2types.SecurityGroup]bool: Map of security groups to cluster tag ownership status
+//   - map[*ec2types.SecurityGroup]IPPermissionSet: Map of security groups to their ingress rules that reference sgID
+//   - error: An error if the operation fails
```
mtulio, line 1927: Those functions shares the implementation with https://github.com/kubernetes/cloud-provider-aws/pull/1209/files#diff-70a4ebdf516a65404dc3c3b3ba9925fc62d0f85bd75e272a4f8ddb4b7b8ab852R1885

tests/e2e/aws_helper.go
```diff
@@ -0,0 +1,236 @@
+/*
```
mtulio, line 1: helpers implemented in https://github.com/kubernetes/cloud-provider-aws/pull/1209/files#diff-1d5c293aba66db82ccec7e1851eaa56aebb4f9ac3a0b7be682eaf75fe6034f25 (also waiting for review)

mtulio:
Commits have been cleaned up.

/test all

mtulio:
I am observing two issues in this PR so far:


> [It] [cloud-provider-aws-e2e] loadbalancer NLB with BYO SG - create with BYO from star

1) e2e identity is missing permissions to scenario: create service with BYO SG on NLB:
```
I0106 16:39:58.845514 32245 loadbalancer.go:1082]   [Warning] SyncLoadBalancerFailed/lbconfig-test-nlb-byo-c: Error 
syncing load balancer: failed to ensure load balancer: error ensuring NLB security group rules: error while updating rules 
to security group "sg-0e755bbafb6239e3d": creating ingress rules for security group "sg-0e755bbafb6239e3d": error 
authorizing security group ingress: "operation error EC2: AuthorizeSecurityGroupIngress, https response error 
StatusCode: 403, RequestID: 0f08cf56-95bd-4b10-8ccc-cb6361739970, api error UnauthorizedOperation: 
You are not authorized to perform this operation. 
User: arn:aws:sts::209411653980:assumed-role/aws-cloud-controller-manager.kube-system.sa.test-cluster--iugf5b/1767716035453412230
 is not authorized to perform: ec2:AuthorizeSecurityGroupIngress on resource: arn:aws:ec2:us-west-2:209411653980:security-group/sg-0e755bbafb6239e3d 
because no identity-based policy allows the ec2:AuthorizeSecurityGroupIngress action. 
```

Action item: investigate why permission isn't added to CI since [it is mentioned as pre-req](https://github.com/kubernetes/cloud-provider-aws/blob/master/docs/prerequisites.md?plain=1#L39).

> [It] [cloud-provider-aws-e2e] loadbalancer NLB with BYO SG update - managed to BYO scenario

2) update e2e for BYO SG NLB can't be implemented in existing e2e scenario the test setup does not support managed SG . 

Proposed Action Itens: remove this test from upstream or implement cloud-config patch for managed SG scenario.

docs/nlb_security_groups.md
```diff
@@ -105,6 +106,119 @@ curl -v $LB_DNS
...
+kubectl patch service ${SVC_NAME} -n ${APP_NAMESPACE} --type=json \
+  --patch '[{"op": "add", "path": "/spec/ports/-", "value": {"name":"http8888","port":8888,"protocol":"TCP","targetPort":8080}}]'
+```
+6. Check if security group ingress rules has been updated
+```sh
+$ aws ec2 describe-security-group-rules --filters Name=group-id,Values=$SG_ID | jq .SecurityGroupRules[].ToPort
+80
+8888
+..
+```
```
alebedev87, line 211: I was thinking that unmanaged (BYO) frontend SG rules were not supposed to be managed (unlike managed frontend SG).

docs/service_controller.md
```diff
@@ -16,7 +16,7 @@ The service controller is responsible for watch for service and node object chan
...
 | service.beta.kubernetes.io/aws-load-balancer-extra-security-groups             | Comma-separated list                | -   | ELB | Specifies additional security groups to be added to ELB.    |
-| service.beta.kubernetes.io/aws-load-balancer-security-groups                   | Comma-separated list                | -   | ELB | Specifies the security groups to be added to ELB. Differently from the annotation "service.beta.kubernetes.io/aws-load-balancer-extra-security-groups", this replaces all other security groups previously assigned to the ELB. |
+| service.beta.kubernetes.io/aws-load-balancer-security-groups                   | Comma-separated list                | -   | ELB,NLB | Specifies the security groups to be added to ELB and NLB. For ELB: differently from the annotation "service.beta.kubernetes.io/aws-load-balancer-extra-security-groups", this replaces all other security groups previously assigned to the ELB. For NLB, it is supported only when load balancer is created with managed security group. |
```
alebedev87, line 19: > For NLB, it is supported only when load balancer is created with managed security group

I'm not sure I understand this part of the description. Isn't this annotations aims at setting unmanaged frontend security groups? Or you meant when `NLBSecurityGroupMode` flag is set to `Managed`?

pkg/providers/v1/aws.go
```diff
@@ -2227,55 +2227,85 @@ func (c *Cloud) buildNLBHealthCheckConfiguration(svc *v1.Service) (healthCheckCo
...
+	// For new NLBs, determine security group based on configuration priority
+
+	// Priority 2: BYO Security Group from annotation
+	// Validation has already ensured this is exactly one valid SG ID (starts with "sg-")
+	if sgAnnotation, hasBYOSG := annotations[ServiceAnnotationLoadBalancerSecurityGroups]; hasBYOSG {
+		sgList := getSGListFromAnnotation(sgAnnotation)
+		if len(sgList) == 1 {
+			klog.Infof("Using BYO security group %q for NLB service %q (from annotation)", sgList[0], serviceName)
+			return sgList, nil
+		}
```
alebedev87, line 2276: Why this limitation of only 1 SG?

pkg/providers/v1/aws.go
```diff
@@ -2227,55 +2227,85 @@ func (c *Cloud) buildNLBHealthCheckConfiguration(svc *v1.Service) (healthCheckCo
...
 	}
 
-	// Do nothing when controller is not in NLB SG managed mode, NLBSecurityGroupMode=Managed.
-	isManaged, err := c.cfg.IsNLBSecurityGroupModeManaged()
-	if err != nil {
-		return nil, fmt.Errorf("error checking NLB security group mode: %w", err)
+	// For new NLBs, determine security group based on configuration priority
+
+	// Priority 2: BYO Security Group from annotation
+	// Validation has already ensured this is exactly one valid SG ID (starts with "sg-")
```
alebedev87, line 2270: Do we accept names too?

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -219,6 +220,38 @@ func (c *Cloud) ensureLoadBalancerv2(ctx context.Context, namespacedName types.N
...
+			expected := sets.New(securityGroups...)
+			actual := sets.New(loadBalancer.SecurityGroups...)
+
+			// Update NLB security groups to prevent SG leak when updating from managed to BYO.
+			if !expected.Equal(actual) {
+				// Unsupported operation by AWS: NLB udpate security groups:
+				// Prevent updating security groups for NLB if the load balancer is created without managed security group.
+				if len(loadBalancer.SecurityGroups) == 0 {
+					return nil, fmt.Errorf("cannot update security groups for NLB %q as it is created without managed security group", loadBalancerName)
+				}
```
alebedev87, line 234: Does this condition handle the case of load balancers created before the managed frontedn SG feature was implemented? If so, it means that upgrades to new controller won't enable BYO SG annotation. Can we change this? For instance by enabling `NLBSecurityGroupMode` after the upgrade?

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -1879,3 +1912,118 @@ func ValidateHealthCheck(s *elbtypes.HealthCheck) error {
...
+	for _, sg := range securityGroups {
+		// Verify ownership via cluster tags
+		isOwned, err := c.isOwnedSecurityGroup(ctx, sg)
+		if err != nil {
+			errors = append(errors, fmt.Errorf("error verifying ownership of security group %q: %w", sg, err))
+			continue
+		}
+
+		if !isOwned {
+			klog.V(2).Infof("Skipping non-owned security group %q for load balancer %q", sg, loadBalancerName)
```
alebedev87, line 1989:
```suggestion
			klog.V(2).Infof("Skipping removal of non-owned security group %q for load balancer %q", sg, loadBalancerName)
```

pkg/providers/v1/aws_validations.go
```diff
@@ -54,18 +55,34 @@ func validateServiceAnnotations(v *awsValidationInput) error {
 	isNLB := isNLB(v.annotations)
 
 	// ServiceAnnotationLoadBalancerSecurityGroups
-	// NLB only: ensure the BYO annotations are not supported and return an error.
-	// FIXME: the BYO SG for NLB implementation is blocked by https://github.com/kubernetes/cloud-provider-aws/pull/1209
-	if _, hasBYOAnnotation := v.annotations[ServiceAnnotationLoadBalancerSecurityGroups]; hasBYOAnnotation {
+	// NLB: validate BYO SG constraints (only one SG, must start with "sg-")
```
alebedev87, line 58: Any links or more details to explain why only one is allowed?

pkg/providers/v1/aws_validations.go
```diff
@@ -54,18 +55,34 @@ func validateServiceAnnotations(v *awsValidationInput) error {
...
+	if annotatedSG, hasBYOAnnotation := v.annotations[ServiceAnnotationLoadBalancerSecurityGroups]; hasBYOAnnotation {
 		if isNLB {
-			return fmt.Errorf("BYO security group annotation %q is not supported by NLB", ServiceAnnotationLoadBalancerSecurityGroups)
+			sgList := getSGListFromAnnotation(annotatedSG)
+
+			// NLB supports only one security group
+			if len(sgList) > 1 {
+				return fmt.Errorf("NLB supports only one security group, got %d security groups %v in annotation %q",
+					len(sgList), sgList, ServiceAnnotationLoadBalancerSecurityGroups)
+			}
```
alebedev87, line 67: Interesting, [looking at ALBC, I can see that multiple SGs are supported](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.17/guide/service/annotations/#security-groups).

alebedev87 (commented):
First look, mostly questions..

docs/service_controller.md
mtulio: You are right, technically is possible to create an NLB with BYO SGs without the managed configuration.

Initially this annotation was though to be used as a fallback mechanism to users who wanted to skip the managed LB created by controller, satisfying the NLB restriction of attaching SG only when it was created with a SG (day-2, managed SG).

> SetSecurityGroups: You can't perform this operation on a Network Load Balancer unless you specified a security group for the load balancer when you created it.

I will keep this thread open to collect more thoughts and review the code if this restriction is hard coded in unmanaged mode.

pkg/providers/v1/aws_loadbalancer.go
mtulio: NLBs can't be updated if it was not created with a Security Group:

> [SetSecurityGroups](https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_SetSecurityGroups.html) : You can't perform this operation on a Network Load Balancer unless you specified a security group for the load balancer when you created it.

I can't see a supported path in this situation, if you have something in mind, would you mind elaborate it, please?

mtulio:
thanks you all for review. We going to shift hands on this PR, @mfbonfigli is working on this feature from now on, and I will let him to open a new PR, I will be engaging reviewing it as soon it is available.

Federico, would you mind taking a look at the review comments so far to ensure we are covering it in your changes? Thanks

pkg/providers/v1/aws_loadbalancer.go
mfbonfigli: +1 to Marco, if a NLB was provisioned without security groups, unfortunately any transition to a NLB with a managed or BYO security group requires deletion and recreation of the NLB, which can create downtime and can be dangerous if the user does not plan for it or is not aware of this behavior.  For this reason I would suggest to leave old NLBs without security groups as they are rather than attempting an automatic deletion and recreation of these old NLBs by the controller

pkg/providers/v1/aws.go
mfbonfigli: Good point. We could, in the new PR #1379 I am working on we only still support SG IDs, but for example the AWS Load Balancer Controller supports [both ids and names](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/service/annotations/?h=service.beta.kubernetes.io%2F#security-groups) in this annotation, so I guess we could implement that here as well if we wanted.

pkg/providers/v1/aws.go
mfbonfigli: This limitation is lifted in PR #1379 where more than one BYO SG can be specified in the annotation value as a comma separated list similarly to what AWS LBC does.

docs/service_controller.md
mfbonfigli: In PR #1379 I am allowing to create a NLB directly with one or more BYO SGs associated if the annotation is present from the beginning, skipping the creation of the controller-owned security group. However the `SetSecurityGroups` permission will still be required else the NLB will be stuck with that SG and the controller won't be able to manipulate SGs anymore after NLB creation if the user has a need to change them (by either adding/removing BYO ones or switching to a managed security group)

Issue #939, opened 2024-06-04
Support Region for DescribeInstance Call


atsai1220:
**What would you like to be added**:
- I would like to see the cloud-controller's DescribeInstance query also use the instance's region.

**Why is this needed**:
- This allows us to create workers in another region (not considering storage needs).

**Questions**:
- Are the repercussions for disabling node-life-cycle controller?

**Findings**:
Currently nodes from another region are "not found" by `node-life-cycle` controller and will be promptly deleted from the cluster after joining.

On 1.30.1 logs
```
node_controller.go:240] error syncing 'ip-10-117-161-37.ap-southeast-2.compute.internal': failed to get instance metadata for node ip-10-117-161-37.ap-southeast-2.compute.internal: instance not found, requeuing
node_controller.go:425] Initializing node ip-10-117-161-37.ap-southeast-2.compute.internal with cloud provider
node_controller.go:229] error syncing 'ip-10-117-161-37.ap-southeast-2.compute.internal': failed to get instance metadata for node ip-10-117-161-37.ap-southeast-2.compute.internal: instance not found, requeuing
```

Configuration
```
      containers:
        - args:
            - '--v=2'
            - '--cloud-provider=aws'
            - '--configure-cloud-routes=false'
          image: registry.k8s.io/provider-aws/cloud-controller-manager:v1.30.1
```

/kind feature

atsai1220:
I can see from CloudTrail that it was looking for an instance in the wrong region and receiving an error
```json
{
    "eventVersion": "1.09",
    "userIdentity": {
        "type": "AssumedRole",
        "principalId": ":i-07143150147441be0",
        "arn": "arn:aws:sts:::assumed-role//i-07143150147441be0",
        "accountId": "",
        "accessKeyId": "",
        "sessionContext": {
            "sessionIssuer": {
                "type": "Role",
                "principalId": "",
                "arn": "arn:aws:iam:::role/",
                "accountId": "982008609023",
                "userName": ""
            },
            "attributes": {
                "creationDate": "2024-06-03T17:16:33Z",
                "mfaAuthenticated": "false"
            },
            "ec2RoleDelivery": "2.0"
        }
    },
    "eventTime": "2024-06-03T21:58:18Z",
    "eventSource": "ec2.amazonaws.com",
    "eventName": "DescribeInstances",
    "awsRegion": "us-west-2",
    "sourceIPAddress": "",
[...]

PR #1372, opened 2026-04-02
Ignore nodes outside configured region

benturnkey:
**What type of PR is this?**
/kind feature

The AWS cloud provider now ignores nodes whose `providerID` points to a different AWS region than the controller's configured region when checking instance existence or shutdown state.

Addresses #939 by preventing controllers from interfering with nodes it doesn't own.

```release-note
The AWS cloud provider now ignores nodes whose `providerID` points to a different AWS region than the controller's configured region when checking instance existence or shutdown state.
```


pkg/providers/v1/instances_v2_test.go
```diff
@@ -120,6 +167,24 @@ func TestInstanceExists(t *testing.T) {
...
+func TestInstanceExistsSkipsDifferentRegionProviderID(t *testing.T) {
+	c, mockedEC2API := getCloudWithMockedDescribeInstancesAndAPI(true, ec2types.InstanceStateNameRunning, "i-abc")
+	c.region = "us-west-2"
+
+	result, err := c.InstanceExists(context.TODO(), &v1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "foreign-node",
+		},
+		Spec: v1.NodeSpec{
+			ProviderID: "aws:///us-east-1c/1abc-2def/i-abc",
```
JoelSpeed, line 180: What does `1abc-2def` represent here?

pkg/providers/v1/instances.go
```diff
@@ -57,6 +57,32 @@ func (i InstanceID) awsString() *string {
...
 type KubernetesInstanceID string
 
+// Region extracts the AWS region from the provider ID when it includes an AZ token.
+// It returns an empty string when the provider ID has no zone/region component.
+func (name KubernetesInstanceID) Region() (string, error) {
+	s := string(name)
+
+	if !strings.HasPrefix(s, "aws://") {
+		// Assume a bare aws instance id (i-1234...)
+		// Build a URL with an empty host (AZ)
```
JoelSpeed: If the instance ID is bare, will it ever have a region prefix? When/why would the instance ID be bare?

benturnkey: > If the instance ID is bare, will it ever have a region prefix?
No, it was an attempt to coerce the bare instance into something parsable without erroring.

> When/why would the instance ID be bare?
Just trying to cover any case.

pkg/providers/v1/instances_v2_test.go
benturnkey: Additional custom identifiers. We only ensure region and instance are valid for operation.

JoelSpeed: Are custom identifiers supported by the CCM/Kubelet? How do other parts of the CCM already handle the provider ID, is there existing logic that parses the provider ID that we could re-use?

pkg/providers/v1/instances.go
JoelSpeed: If the instance ID isn't prefixed, and isn't prefixed `aws://`, I would expect we just ignore the node rather than trying to coerce anything no?

benturnkey: I mean, when you put it that way, yeah.

I'll push a change.

pkg/providers/v1/instances_v2_test.go
benturnkey: Additional identifiers could be used by Fargate or Outpost.

https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws_test.go#L3675

JoelSpeed:
@benturnkey There are a lot of commits in this PR which don't look like they are relevant to the actual goal of the PR, what's going on with those other commits?

benturnkey:
> @benturnkey There are a lot of commits in this PR which don't look like they are relevant to the actual goal of the PR, what's going on with those other commits?

Yeah, sorry, I reused the branch for some other testing. Reverted :pray:

PR #1374, opened 2026-04-06
fix: Add explicit HTTP request timeouts to all AWS SDK clients

howard-junec:
Add explicit HTTP request timeouts (30s) to all AWS SDK clients using the SDK's BuildableClient via WithHTTPClient on every LoadDefaultConfig call. Using BuildableClient instead of a bare http.Client preserves the SDK's default transport settings (TLS 1.2, connection pooling, dialer config). This prevents the Go SDK's clock skew overcorrection from breaking all API calls when a single response is slow. Added AST test to fail if any future LoadDefaultConfig is added without WithHTTPClient.

/kind bug

Add explicit HTTP request timeouts (30s) to all AWS SDK clients by configuring
`WithHTTPClient` on every `LoadDefaultConfig` call. Without a timeout, a single
slow API response can trigger the Go AWS SDK v2's clock skew overcorrection,
breaking all subsequent API calls with HTTP 401 errors.

Also adds an AST-based test that scans source files and fails if any
`LoadDefaultConfig` is added without `WithHTTPClient`.

NONE

`tests/e2e/loadbalancer.go` has one `LoadDefaultConfig` without timeout —
intentionally excluded since it's test tooling, not production controller code.
The AST test skips the `tests/` directory accordingly.

```release-note
NONE
```

### DIFF
--- a/cmd/ecr-credential-provider/main.go
+++ b/cmd/ecr-credential-provider/main.go
@@ -21,6 +21,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"net/http"
 	"net/url"
 	"os"
 	"regexp"
@@ -42,6 +43,8 @@ import (
 )
 
 const ecrPublicRegion string = "us-east-1"
+
+var defaultHTTPClient = &http.Client{Timeout: 30 * time.Second}
 
 var ecrPublicHosts []string = []string{"public.ecr.aws", "ecr-public.aws.com"}
 
@@ -74,10 +77,13 @@ func defaultECRProvider(ctx context.Context, region string) (ECR, error) {
 	if region != "" {
 		cfg, err = config.LoadDefaultConfig(ctx,
 			config.WithRegion(region),
+			config.WithHTTPClient(defaultHTTPClient),
 		)
 	} else {
 		klog.Warningf("No region found in the image reference, the default region will be used. Please refer to AWS SDK documentation for configuration purpose.")
-		cfg, err = config.LoadDefaultConfig(ctx)
+		cfg, err = config.LoadDefaultConfig(ctx,
+			config.WithHTTPClient(defaultHTTPClient),
+		)
 	}
 
 	if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/ecr-credential-provider/main.go
```diff
@@ -74,10 +77,13 @@ func defaultECRProvider(ctx context.Context, region string) (ECR, error) {
 	if region != "" {
 		cfg, err = config.LoadDefaultConfig(ctx,
 			config.WithRegion(region),
+			config.WithHTTPClient(defaultHTTPClient),
```
kmala, line 80: Can we just override the the default client's timeout since aws sdk has other default for transport which are good to take https://github.com/aws/aws-sdk-go-v2/blob/main/aws/transport/http/client.go#L44

pkg/providers/v1/aws_sdk.go
```diff
@@ -40,6 +42,12 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// defaultHTTPClient is shared across all AWS SDK clients to enforce an explicit
+// HTTP request timeout and reuse connection pools. Without a timeout, a single
+// slow response can trigger the Go SDK's clock skew overcorrection and break all
+// subsequent API calls (COE-389792).
+var defaultHTTPClient = &http.Client{Timeout: 30 * time.Second}
```
kmala: is 30 sec enough to list all the instances or does pagination take care of that ?

cmd/ecr-credential-provider/main.go
howard-junec: thanks, I will update to use `awshttp.NewBuildableClient().WithTimeout(30 * time.Second)` to preserve the SDK's default transport settings.

pkg/providers/v1/aws_sdk.go
howard-junec: The 30s timeout is per HTTP request, not per operation. DescribeInstances paginates via NextToken, each page is a separate HTTP call with its own 30s window.

PR #1377, opened 2026-04-09
Add AWS API metrics middleware for status codes and error tracking

howard-junec:
/kind feature

Add Deserialize middleware to track AWS API error response status codes as metrics.

Goal: Monitor controller metrics for AWS API errors — track API call errors by status code in CCM.

New metric:
- `cloudprovider_aws_api_response_status_total` — counts error responses by specific status code (400, 401, 500, etc.), with service and operation labels. Only records status codes >= 400.

Middleware is added to `AddMiddleware()` covering EC2, ELB, ELBv2, KMS clients, and separately to `NewStsClient()` for STS.

Unit tests included.

NONE

IMDS is intentionally not covered — it only runs at startup for metadata lookup and does not go through AddMiddleware().

Metric definitions are in `pkg/providers/v1/aws_metrics.go` alongside existing metrics. STS coverage is achieved by passing middleware via `NewStsClient's` new apiOptions parameter.

```release-note
NONE
```


pkg/services/aws_api_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/aws/smithy-go"
+	"github.com/aws/smithy-go/middleware"
+	"github.com/aws/smithy-go/transport/http"
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+var (
+	// AWSAPIResponseStatusTotal counts AWS API responses by status code class (4xx, 5xx).
+	AWSAPIResponseStatusTotal = metrics.NewCounterVec(
```
DanielCKennedy: I don't understand the difference between these 2 metrics. A single metric should be sufficient which counts error counts for each error code

howard-junec: just updated this, thanks!

pkg/services/aws_api_metrics.go
```diff
@@ -0,0 +1,72 @@
...
+	"sync"
+
+	"github.com/aws/smithy-go/middleware"
+	"github.com/aws/smithy-go/transport/http"
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+// AWSAPIResponseStatusTotal counts AWS API responses by status code.
+var AWSAPIResponseStatusTotal = metrics.NewCounterVec(
```
kmala: Why can't we use and extend the existing aws_metrics.go?

kmala: It already has the errors, latency and throttling... I think we should try to expand that

howard-junec: The metric is in `pkg/services/` because `pkg/services/aws_sts.go` also needs to reference it for the STS client middleware. `providers/v1` already imports services (5 places), so if the metric lived in `providers/v1/aws_metrics.go`, `services/aws_sts.go` couldn't import it without creating a circular dependency (services -> providers/v1 -> services).

kmala: That cyclic dependency can be fixed with a small refactor right?

howard-junec: I'll refactor `NewStsClient()` to accept optional API options, so the metric and middleware can live in `providers/v1/aws_metrics.go` alongside the existing metrics. Will push an update.

howard-junec: just updated this, thanks!

pkg/providers/v1/aws.go
```diff
@@ -518,7 +519,11 @@ func init() {
 
 		var creds *stscreds.AssumeRoleProvider
 		if cfg.Global.RoleARN != "" {
-			stsClient, err := services.NewStsClient(ctx, regionName, cfg.Global.RoleARN, cfg.Global.SourceARN)
+			stsClient, err := services.NewStsClient(ctx, regionName, cfg.Global.RoleARN, cfg.Global.SourceARN,
+				func(stack *smithymiddleware.Stack) error {
```
yue9944882, line 523: nit: looks like no need to expose `func(stack *smithymiddleware.Stack) error` as a parameter here? can we move the definition into `NewStsClient`?

howard-junec: I considered moving it in, but `awsAPIMetricsMiddleware()` is defined in `pkg/providers/v1` (alongside the other k8s metrics registrations and the prometheus counter it writes to), while `NewStsClient` is in `pkg/services`. Moving it would require `services` to import `providers/v1`, creating a circular dependency. The variadic `apiOptions` keeps the dependency  direction correct (`providers/v1` -> `services`), same injection pattern the AWS SDK itself uses. Happy to refactor if you see a cleaner way to avoid the cycle.

yue9944882: thanks for clarifying, no need for that refactoring in that case..

yue9944882 (commented):
/lgtm
/approve

PR #1379, opened 2026-04-13
feat: Add BYO security group support for NLB

### LINKED ISSUE #1376: Support for externally managed security groups for NLB

Hi! In my previous issue https://github.com/kubernetes/cloud-provider-aws/issues/1151, support for managed security groups for NLB was added, and we really appreciate it!

But can we expect to be able to attach an existing, externally managed security group to an NLB? Something similar to what’s available in the AWS Load Balancer Controller:
https://kubernetes-sigs.github.io/aws-load-balancer-controller/v3.1/guide/service/annotations/#security-groups

**Why is this needed**:

In some cases, we need to prepare the policy set externally and provide the load balancer with pre-created security groups.

Thanks!

/kind feature

mfbonfigli:
**What type of PR is this?**
/kind feature

This PR adds support for Bring Your Own (BYO) security groups for Network Load Balancers (NLB) when the controller is configured with `NLBSecurityGroupMode = Managed`. AWS Network Load Balancers support specifying Security Groups to restrict traffic flow. In PR #1158 opt-in support for Managed Security Groups was added. This PR builds on top of that to add support for BYO Security Groups via a custom service annotation.

Key changes:
- Implements BYO security group support for NLBs via the `service.beta.kubernetes.io/aws-load-balancer-security-groups` annotation (requires `NLBSecurityGroupMode = Managed` in cloud-config).
- Adds automatic cleanup of stale security group rules when transitioning between managed and BYO security groups.
- Adds unit tests covering BYO SG scenarios for NLBs.
- Adds e2e tests for Service creation with NLB and BYO SG, and transitions from managed > byo and byo > managed
- Changes the test infrastructure (via `run.sh`) to inject the `NlbSecurityGroupMode=Managed` into the cloud config. This allows writing e2e tests for the feature. See the related [kops PR](https://github.com/kubernetes/kops/pull/18211) to add support for this configuration.
- Transition from managed to byo and byo to managed requires `elasticloadbalancing:SetSecurityGroups` IAM permission. Documentation has been updated to reflect this requirement.
- Updates documentation with usage examples.

**Why we need it:**
Users need the ability to attach externally managed security groups to NLBs, similar to the functionality available in the AWS Load Balancer Controller, and similarly to what is already supported for AWS Classic Load Balancers. This enables:
- Pre-configured security group policies prepared externally
- Compliance requirements that mandate specific security group configurations
- Integration with existing security group architectures managed outside the cluster
- Fine-grained control over NLB network access policies

Example of a recent issue requesting this feature is Issue #1376 

Fixes #1376 

- This PR is forked from #1209 which adds utilities for Classic Load Balancers reused here to avoid duplicated work. Ideally #1209 should be merged before this one.
- The BYO security group annotation for NLB only works when `NLBSecurityGroupMode = Managed` is set in cloud-config due to the limitations in AWS NLB which if provisioned with a Security Group should always have one.
- When BYO security groups are specified, the controller does not manage ingress rules (user's responsibility to configure externally for now)
- When transitioning from managed to BYO security groups, the controller automatically cleans up and deletes the old managed security groups.
- When transitioning from BYO to managed security groups, the controller properly deassociates BYO security groups without deleting them (user retains ownership)
- Security group IDs only (format: `sg-*`) are supported by the annotation; security group names are not resolved in the current implementation. A separate issue and PR will be created for that as support for security group names is also needed for BYO Security Groups for Classic Load Balancers and implementing all in a single PR would expand the scope too much.

**Files changed:**
- `pkg/providers/v1/aws.go`: Core BYO SG logic for NLB
- `pkg/providers/v1/aws_loadbalancer.go`: Security group cleanup helpers
- `pkg/providers/v1/aws_test.go`: Unit tests for BYO SG scenarios
- `pkg/providers/v1/aws_validations.go`: Updated validation logic
- `docs/nlb_security_groups.md`: Added BYO SG documentation and test procedures
- `docs/service_controller.md`: Updated annotation table to reflect NLB support
- `hack/e2e/run.sh`: Updated `kops` flags to enable `NlbSecurityGroupMode=Managed` in the controller cloud config.
- `tests/e2e/loadbalancer.go`: Added e2e tests and related utility functions.

```release-note
Added support for user-provided (BYO) security groups for Network Load Balancers via the `service.beta.kubernetes.io/aws-load-balancer-security-groups` annotation. This feature requires `NLBSecurityGroupMode = Managed` in the cloud-config and the `elasticloadbalancing:SetSecurityGroups` IAM permission. See the [NLB Security Groups documentation](https://github.com/kubernetes/cloud-provider-aws/blob/master/docs/nlb_security_groups.md) for usage examples and configuration details.

```


mtulio:
Hey @mfbonfigli, thanks for submitting the PR. 🎉 

I will be reviewing this PR once you signalize me do so :)

/assign @mtulio 

Meanwhile, please rebase the PR to fetch recent changes, as well I am setting the test labels as you mentioned on Slack that have tested successfully in local environment: 

/ok-to-test

pkg/providers/v1/aws.go
```diff
@@ -2231,44 +2226,132 @@ func (c *Cloud) buildNLBHealthCheckConfiguration(svc *v1.Service) (healthCheckCo
...
+	// Managed mode disabled & existing NLB: keep the existing SGs.
+	if !isManaged && lbExists {
+		return existingSGs, nil
 	}
 
-	serviceName := types.NamespacedName{Namespace: svc.Namespace, Name: svc.Name}
+	// Managed mode disabled & new NLB: don't create any SGs.
+	if !isManaged && !lbExists {
+		return &nlbSecurityGroups{}, nil
+	}
```
mtulio: What about the flow: mode disabled & new NLB & BYO SG? Is there any limitation in the controller to allow this?

mfbonfigli: We could add support for that, however I do not recommend because I believe it creates a scenario where the controller behavior would be inconsistent.

 The scenario is **Provisioning a NLB with BYO SG and then removing the annotation, with `NLBSecurityGroupMode` not set**:

In this scenario the NLB is provisioned with the BYO group attached, then the annotation is removed.

Since `NLBSecurityGroupMode` **is not** set to `Managed`, I believe we should not let the controller create a Security Group and attach it to the NLB to replace the BYO one (note that the NLB must always have a SG attached if initially created with one), or else it would behave as if `NLBSecurityGroupMode` was set to `Managed` which would be odd as we didn't enable that feature. 

Given this, then the only options left in this case, both not appropriate in my opinion, would be:

- Option 1: Delete the NLB and recreate it without a security group. This approach is problematic because it produces downtime, and also is inconsistent with the "no downtime" scenario that we would instead have if `NLBSecurityGroupMode` was set to `Managed`, which would just result in a new security group attached to the same NLB without deletion and recreation of the NLB itself.

- Option 2: leave the BYO SG attached, which is also not a good choice as it disregards the desired state after the annotation removal.

For this reason I believe that it would be cleaner if BYO groups were only supported if the controller was configured with `NLBSecurityGroupMode` set to `Managed`, to be intended as "The controller manages security groups for NLBs, which includes not just creation of new SGs but also association and deassociation of BYO ones to NLBs".

Let me know what do you think, open to different opinions.

mtulio: yeah, I think option 1 would be anti-pattern of CCM and we need to avoid that. Option 2 may be complicated too since the user is removing the BYO annotation.

> For this reason I believe that it would be cleaner if BYO groups were only supported if the controller was configured with NLBSecurityGroupMode set to Managed, to be intended as "The controller manages security groups for NLBs, which includes not just creation of new SGs but also association and deassociation of BYO ones to NLBs".

Right, I agree with restricting to managed configuration only, otherwise it may increase the complexity of the feature with additional logic to validation phase to cover the updates - impacting more api calls, etc.

It would be nice to have this decision of not implementing right now with complexity in a comment on top of the restriction (line 2321. :)

mfbonfigli: added comment in last revision

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -219,6 +220,41 @@ func (c *Cloud) ensureLoadBalancerv2(ctx context.Context, namespacedName types.N
...
+				}
+				for _, sg := range securityGroups {
+					if !oldSGSet[sg] {
+						needsSecurityGroupUpdate = true
+						break
+					}
+				}
+			}
+
+			if needsSecurityGroupUpdate && len(securityGroups) > 0 {
```
mtulio: can we use [sets](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets) here? similar check: https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws_loadbalancer.go#L1254-L1258

tests/e2e/loadbalancer.go
```diff
@@ -247,6 +250,246 @@ var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
...
+				cfg.byoSecurityGroupID, err = createSecurityGroup(cfg.ctx, cfg, securityGroupName, fmt.Sprintf("BYO Security Group for NLB e2e test service %s/%s", cfg.svc.Namespace, cfg.svc.Name))
+				framework.ExpectNoError(err, "Failed to create BYO security group")
+				framework.Logf("Created BYO security group: %s", cfg.byoSecurityGroupID)
+
+				DeferCleanup(func(ctx context.Context) {
+					if cfg.byoSecurityGroupID != "" {
+						framework.Logf("Cleaning up BYO security group: %s", cfg.byoSecurityGroupID)
+						framework.Logf("Waiting 30s for ENIs to be detached from security group...")
+						time.Sleep(30 * time.Second)
+						err := deleteSecurityGroup(ctx, cfg.byoSecurityGroupID)
```
mtulio: What about to  use `Eventually` or `PollUntilContextTimeout` to safety wait for dependencies to instead of unpredictable sleep?

tests/e2e/loadbalancer.go
```diff
@@ -247,6 +250,246 @@ var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
...
```
mtulio: same question about the retries above.

tests/e2e/loadbalancer.go
```diff
@@ -247,6 +250,246 @@ var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
...
+				// Step 2: Add BYO SG annotation
+				framework.Logf("Step 2: Adding BYO SG annotation to service")
+				cfg.svc.Annotations[annotationLBSecurityGroups] = cfg.byoSecurityGroupID
+				newSvc, err := cfg.kubeClient.CoreV1().Services(cfg.LBJig.Namespace).Update(cfg.ctx, cfg.svc, metav1.UpdateOptions{})
+				framework.ExpectNoError(err, "Failed to update Kubernetes Service with BYO SG annotation")
+				cfg.svc = newSvc
+				framework.Logf("Updated service with BYO SG annotation: %s=%s", annotationLBSecurityGroups, cfg.byoSecurityGroupID)
+
+				// Wait for controller to process the update
+				time.Sleep(15 * time.Second)
```
mtulio: can we also use eventually here to prevent transient issues that could be generated by Sleep?

tests/e2e/loadbalancer.go
```diff
@@ -247,6 +250,246 @@ var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
...
+				byoSGs, err := getLoadBalancerSecurityGroups(cfg.ctx, lbDNS)
+				framework.ExpectNoError(err, "Failed to get load balancer security groups after BYO annotation")
+				framework.Logf("NLB %s has security groups after BYO annotation: %+v", lbDNS, byoSGs)
+
+				if len(byoSGs) != 1 {
+					framework.Failf("Expected NLB to have exactly 1 security group (BYO SG), got %d: %v", len(byoSGs), byoSGs)
+				}
+				if byoSGs[0] != cfg.byoSecurityGroupID {
+					framework.Failf("Expected NLB to have BYO SG %q, got %q", cfg.byoSecurityGroupID, byoSGs[0])
+				}
```
mtulio: What about to rewrite those errors conditions using the gomega assertions? eg. [Expect](https://onsi.github.io/gomega/#making-assertions)

tests/e2e/loadbalancer.go
```diff
@@ -247,6 +250,246 @@ var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
...
+
+				// Verify managed SG was deleted
+				framework.Logf("Verifying managed SG %s was deleted", managedSGID)
+				_, err = getSecurityGroup(cfg.ctx, managedSGID)
+				if err == nil {
+					framework.Failf("Expected managed SG %s to be deleted, but it still exists", managedSGID)
+				}
+				if !strings.Contains(err.Error(), "InvalidGroup.NotFound") {
+					framework.Failf("Error checking if managed SG was deleted: %v", err)
+				}
```
mtulio: same for those checks, it would be nice to use the gomega framework

tests/e2e/loadbalancer.go
```diff
@@ -247,6 +250,246 @@ var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
...
+				cfg.byoSecurityGroupID, err = createSecurityGroup(cfg.ctx, cfg, securityGroupName, fmt.Sprintf("BYO Security Group for NLB e2e test service %s/%s", cfg.svc.Namespace, cfg.svc.Name))
+				framework.ExpectNoError(err, "Failed to create BYO security group")
+				framework.Logf("Created BYO security group: %s", cfg.byoSecurityGroupID)
+
+				DeferCleanup(func(ctx context.Context) {
+					if cfg.byoSecurityGroupID != "" {
+						framework.Logf("Cleaning up BYO security group: %s", cfg.byoSecurityGroupID)
+						framework.Logf("Waiting 30s for ENIs to be detached from security group...")
+						time.Sleep(30 * time.Second)
+						err := deleteSecurityGroup(ctx, cfg.byoSecurityGroupID)
```
mtulio: same comment of retry pattern above

tests/e2e/loadbalancer.go
```diff
@@ -247,6 +250,246 @@ var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
...
+				byoSGs, err := getLoadBalancerSecurityGroups(cfg.ctx, lbDNS)
+				framework.ExpectNoError(err, "Failed to get load balancer security groups")
+				framework.Logf("NLB %s has security groups: %+v", lbDNS, byoSGs)
+
+				if len(byoSGs) != 1 {
+					framework.Failf("Expected NLB to have exactly 1 security group (BYO SG), got %d: %v", len(byoSGs), byoSGs)
+				}
+				if byoSGs[0] != cfg.byoSecurityGroupID {
+					framework.Failf("Expected NLB to have BYO SG %q, got %q", cfg.byoSecurityGroupID, byoSGs[0])
+				}
```
mtulio: is it possible to use gomega assertions here too?

mtulio (commented):
Thanks for submitting the PR. I have only main questions/suggestions for e2e handling. I see this PR is blocked by

pkg/providers/v1/aws_loadbalancer.go
mfbonfigli: Good catch. updated.

tests/e2e/loadbalancer.go
mfbonfigli: updated to use `gomega.Eventually`

tests/e2e/loadbalancer.go
mfbonfigli: updated to use `gomega` assertions where possible and where it was not inconsistent with the rest of the pre-existing tests (for example some assertions used `framework.ExpectNoError` rather than the equivalent `gomega` assertions. I kept using the existing approach for those.)

tests/e2e/loadbalancer.go
mfbonfigli: good suggestion, I updated the tests to use those approaches

mfbonfigli:
Thanks @mtulio , I should have implemented your feedback, feel free to review it again when you have some time. Thanks!

docs/nlb_security_groups.md
```diff
@@ -4,7 +4,7 @@
...
 
-> Note: The BYO SG (user-provided security groups) annotations (`service.beta.kubernetes.io/aws-load-balancer-security-groups` and `service.beta.kubernetes.io/aws-load-balancer-extra-security-groups`) are valid only for Classic Load Balancers. To learn more about supported annotations by load balancer type, see the [service_controller documentation][doc-ctrl-service].
+> Note: The BYO SG (user-provided security groups) annotation `service.beta.kubernetes.io/aws-load-balancer-security-groups` is supported for both Classic Load Balancers and Network Load Balancers (requires `NLBSecurityGroupMode = Managed` for NLB). When managed mode is disabled for NLB, this annotation is ignored. The annotation `service.beta.kubernetes.io/aws-load-balancer-extra-security-groups` is valid only for Classic Load Balancers. To learn more about supported annotations by load balancer type, see the [service_controller documentation][doc-ctrl-service].
```
mtulio: Maybe this?
```suggestion
> Note 1: The BYO SG (user-provided security groups) annotation `service.beta.kubernetes.io/aws-load-balancer-security-groups` is supported for both Classic Load Balancers and Network Load Balancers (requires `NLBSecurityGroupMode = Managed` for NLB). When managed mode is disabled for NLB, this annotation is ignored. 
> Note 2: The annotation `service.beta.kubernetes.io/aws-load-balancer-extra-security-groups` is valid only for Classic Load Balancers. To learn more about supported annotations by load balancer type, see the [service_controller documentation][doc-ctrl-service].
```

docs/nlb_security_groups.md
```diff
@@ -105,13 +108,88 @@ curl -v $LB_DNS
...
+
+# Should show your BYO security group ID
+```
+
+3. Verify the BYO security group is NOT tagged with cluster ownership:
+```sh
+aws ec2 describe-security-groups --group-ids $BYO_SG_ID \
+  --query 'SecurityGroups[0].Tags' --output json
+
+# Should NOT contain kubernetes.io/cluster/<cluster-name>: owned tag
```
mtulio: Maybe something like this?
```suggestion
# Should NOT contain the key-value pair: kubernetes.io/cluster/<cluster-name>: owned tag
```

pkg/providers/v1/aws.go
```diff
@@ -2318,6 +2427,104 @@ func (c *Cloud) ensureNLBSecurityGroupRules(ctx context.Context, securityGroups
...
+// across other cluster-owned Security Groups and then deleting them if they are cluster-owned.
+//
+// Parameters:
+//   - ctx: The context for the request.
+//   - loadBalancerName: The name of the network load balancer.
+//   - sgsToClean: The security groups to clean up.
+//
+// Returns:
+//   - error: An error if any issue occurs while cleaning up security groups or revoking rules.
+func (c *Cloud) cleanupNLBSecurityGroups(ctx context.Context, loadBalancerName string, sgsToClean *nlbSecurityGroups) error {
```
mtulio: I see building blocks of this function similar [removeOwnedSecurityGroups()](https://github.com/kubernetes/cloud-provider-aws/blob/dafc0a1c2edcfd0f19fc9a0928af094d98f09a69/pkg/providers/v1/aws_loadbalancer.go#L1940), do you think we can reuse/merge logic to increase maintainability.

docs/nlb_security_groups.md
mfbonfigli: sgtm, will update

pkg/providers/v1/aws.go
mfbonfigli: This is a good point. Initially I diverged from CLBs to handle some logic parts differently but there is probably an opportunity to reuse code here and in other parts. Will see if I can reuse more code without changing the behavior.

mfbonfigli: The latest revision has a significant refactor that allowed to reduce code duplication a lot by reusing some approaches of BYO SG for CLBs, please take a look and let me know what you think.

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -1949,29 +2003,11 @@ func (c *Cloud) removeOwnedSecurityGroups(ctx context.Context, loadBalancerName
...
-				GroupId:       sgTarget.GroupId,
-				IpPermissions: sgPerms.List(),
-			}); err != nil {
-				allErrs = append(allErrs, fmt.Errorf("error revoking security group ingress rules from %q: %w", aws.ToString(sgTarget.GroupId), err))
-				continue
-			}
+		// from cluster-tagged security groups.
+		if err := c.revokeSecurityGroupReferences(ctx, sg); err != nil {
+			allErrs = append(allErrs, err)
+			continue
```
mfbonfigli, line 2010: note: here I just extracted the code into a separate more specific function for better readability, the logic did not change.

mtulio (commented):
Thanks for addressing the comments, @mfbonfigli ! 🚀 

/lgtm

@damdo @nrb  Would you mind taking a look, please?

mtulio:
cc @JoelSpeed

JoelSpeed:
We should really have someone from AWS review this, @kmala are you able to take a look?

kmala:
> We should really have someone from AWS review this, @kmala are you able to take a look?

i can take a look either today or by Monday

pkg/providers/v1/aws.go
```diff
@@ -2237,75 +2237,95 @@ func (c *Cloud) buildNLBHealthCheckConfiguration(svc *v1.Service) (healthCheckCo
...
 // Returns:
-//   - []string: A list of security group IDs to be associated with the NLB.
-//   - error: An error if any issue occurs while ensuring the NLB security group.
-func (c *Cloud) ensureNLBSecurityGroup(ctx context.Context, loadBalancerName, clusterName string, svc *v1.Service) ([]string, error) {
-	annotations := svc.Annotations
-	loadBalancer, err := c.describeLoadBalancerv2(ctx, loadBalancerName)
+//   - []string: Security group IDs to attach to the NLB.
+//   - bool: true if SGs are managed (need rules), false if BYO (don't modify).
+//   - error: An error if any issue occurs.
+func (c *Cloud) ensureNLBSecurityGroup(ctx context.Context, clusterName string, svc *v1.Service, existingLB *elbv2types.LoadBalancer, annotations map[string]string) ([]string, bool, error) {
```
kmala: why `annotations` as a param, we already have service object right?

pkg/providers/v1/aws.go
```diff
@@ -2237,75 +2237,95 @@ func (c *Cloud) buildNLBHealthCheckConfiguration(svc *v1.Service) (healthCheckCo
...
 	}
 
-	if loadBalancer != nil {
-		// Existing NLB with security groups.
-		if len(loadBalancer.SecurityGroups) > 0 {
-			return loadBalancer.SecurityGroups, nil
+	// Managed mode disabled: keep existing SGs (if LB exists) or return empty
+	if !isManaged {
+		if existingLB != nil {
+			return existingLB.SecurityGroups, false, nil
```
kmala, line 2271: nit: `existingLB.SecurityGroups` can be nil now right where as previously we are making sure to not return nil. Didn't look at the usage but do you think its better to keep the same behavior?

pkg/providers/v1/aws.go
```diff
@@ -2237,75 +2237,95 @@ func (c *Cloud) buildNLBHealthCheckConfiguration(svc *v1.Service) (healthCheckCo
...
+	additionalTags := getKeyValuePropertiesFromAnnotation(annotations, ServiceAnnotationLoadBalancerAdditionalTags)
+
+	// createSecurityGroup will ensure the creation of the managed SG
+	// or return the existing one if it already exists.
+	securityGroupID, err := c.createSecurityGroup(ctx, sgName, sgDescription, additionalTags)
 	if err != nil {
-		return nil, fmt.Errorf("unable to create security group for NLB: %w", err)
+		return nil, false, fmt.Errorf("unable to create security group for NLB: %w", err)
 	}
-	klog.Infof("Created NLB security group %q for service %q", securityGroupID, serviceName)
```
kmala, line 2286: Can we keep the logs when we create which would be useful for debugging

pkg/providers/v1/aws.go
```diff
@@ -2237,75 +2237,95 @@ func (c *Cloud) buildNLBHealthCheckConfiguration(svc *v1.Service) (healthCheckCo
...
-	securityGroupID := securityGroups[0]
 
+	// Check if managed mode is enabled
+	isManaged, err := c.cfg.IsNLBSecurityGroupModeManaged()
+	if err != nil {
+		return fmt.Errorf("error checking NLB security group mode: %w", err)
+	}
+
+	// Don't manage rules when managed mode is disabled
+	if !isManaged {
```
kmala: This function is only called when `isManaged` , then do we need this check here?

hack/e2e/run.sh
```diff
@@ -128,7 +128,7 @@ if [[ "${UP}" = "yes" ]]; then
       --run-id="${test_run_id}" \
       --cloud-provider=aws \
       --cluster-name="${CLUSTER_NAME}" \
-      --create-args="--dns=none --zones=${ZONES} --node-size=m5.large --control-plane-size=m5.large --set=cluster.spec.cloudControllerManager.cloudProvider=aws --set=cluster.spec.cloudControllerManager.image=${IMAGE_NAME}:${IMAGE_TAG}" \
+      --create-args="--dns=none --zones=${ZONES} --node-size=m5.large --control-plane-size=m5.large --set=cluster.spec.cloudControllerManager.cloudProvider=aws --set=cluster.spec.cloudControllerManager.image=${IMAGE_NAME}:${IMAGE_TAG} --set=cluster.spec.cloudProvider.aws.nlbSecurityGroupMode=Managed" \
```
kmala, line 131: This would mean we are only testing the managed mode correct? do you think it would make sense to have separate test for managed and non managed so we are sure that existing config also works since there is separate code paths based on this field?

mfbonfigli: Good point.  That either requires to:

1. Create two clusters with different configurations, which would make CI times/costs grow.
or
2. Find a way to dynamically manipulate the CCM configuration during the tests (which I think it essentially requires changing the cloud-config and find a way to trigger a reloading of the CCM at runtime) and then roll it back after the tests, plus ensuring that tests do not run in parallel. I am not sure if this is even feasible to be honest, and even if it is I have doubts if it can create stability problems on the test infrastructure (so an increase in flakes).  

For now I went with this approach as it's simpler and in a sense backward compatible as essentially managed mode only adds security groups to NLBs, so its more restrictive in a sense. 

Let me know what do you think and if you want me to explore either of these.

pkg/providers/v1/aws.go
mfbonfigli: that's a great point, indeed I can remove that. Will update.

pkg/providers/v1/aws.go
mfbonfigli: I think we could, but since the variable is a slice, a nil slice behaves pretty much almost the same as an empty slice (except for explicit nil checks) so I don't believe it would create any problem, but I can keep the existing behavior no problem, will update.

pkg/providers/v1/aws.go
mfbonfigli: good point, this is a redundant safety check, I can remove it

mfbonfigli:
@kmala I have updated the code with your feedback. 

The only pending item should be the comment regarding the e2e tests, I'd appreciate if you could share your opinion there. I'm hesitant to introduce complexity into the test setup to handle that

Edit: Discussed at SIG meeting today, will work on a proposal for how to change the cloud config at test execution time and update the PR again.

hack/e2e/run.sh
mfbonfigli: Another thing to factor in is that the new feature requires the `SetSecurityGroups` IAM permission. `kOps` only injects that if the cluster is provisioned with `--set=cluster.spec.cloudProvider.aws.nlbSecurityGroupMode=Managed` ([see here](https://github.com/kubernetes/kops/pull/18211#discussion_r3124005848)), so if we want to test the feature, we must provision the cluster with this flag set in any case.

mfbonfigli:
@kmala Could you take a look at the updated revision? 

As discussed at the latest SIG meeting the new approach consists in running the NLB relevant tests also against a cloud config without `NLBSecurityGroupMode=Managed` set so that we can have confidence no regression in NLB configurations is introduced when the mode is not enabled.

Since these tests edit the cloud config, I'm configuring them as `Serial` and `Ordered` so that (1) they don't risk of running now or in future concurrently with other tests and (2) they manage the changes of cloud config and its rollback at the end using a `BeforeAll` / `AfterAll` clause.

Please let me know what do you think. Thanks!

tests/e2e/cloudconfig.go
```diff
@@ -0,0 +1,386 @@
...
+	if err := m.createConfigMapAndPatchDaemonSet(ctx, cs, ds, volumeIdx, configContent); err != nil {
+		return err
+	}
+
+	// Restart CCM pods and wait for them to be ready
+	return restartCCMPods(ctx, cs, m.restartTimeout)
+}
+
+// restoreCloudConfig restores the original CCM cloud configuration as saved by setCloudConfig.
+// This restores the hostPath volume and deletes the temporary ConfigMap.
```
mtulio: What about this one? Just suggestion to make it clear that the file is being restored to the new format, which is "managed" configuration.
```suggestion
// This restores the hostPath volume and deletes the temporary ConfigMap created by kops.
```

tests/e2e/loadbalancer.go
```diff
@@ -56,354 +59,647 @@ var (
...
+	// Run relevant NLB tests with NLBSecurityGroupMode disabled via cloud config override
+	Context("[nlb-security-group-mode-disabled] loadbalancer", Serial, Ordered, func() {
+		cloudConfigMgr := newCloudConfigManager(withRestartTimeout(3 * time.Minute))
 
-			// Update the service to cluster IP
-			By("cleaning up: converting service to ClusterIP")
-			_, err = e2e.LBJig.UpdateService(ctx, func(s *v1.Service) {
-				s.Spec.Type = v1.ServiceTypeClusterIP
+		BeforeAll(func(ctx context.Context) {
+			// Disable NLB managed security group mode by setting an empty config
```
mtulio: Maybe this to make it clear CCM defaults?
```suggestion
			// Disable NLB managed security group mode by setting an empty config (default CCM configuration)
```

mtulio (commented):
great refact in the e2e tests by grouping test by supported scenarios on LB types, as well running only a subset after patching CCM to use new cloud config without managed configuration set (CCM defaults).

I have just few nit/comments about documenting the approach. Feel free to skip it.

/lgtm

tests/e2e/cloudconfig.go
mfbonfigli: Sure, I would however write it as follows to be more clear:

```suggestion
// This restores the hostPath volume created by kops and deletes the temporary ConfigMap injected.
```

mfbonfigli:
@kmala I saw that @mtulio reviewed again my changes, would you have some time to take a look as well? Thank you!

pkg/providers/v1/aws.go
```diff
@@ -2461,15 +2479,31 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
...
 		)
 		if err != nil {
 			return nil, err
 		}
 
-		// Ensure SG rules only if the LB reconciliator finished successfully.
-		if err := c.ensureNLBSecurityGroupRules(ctx, securityGroups, ec2SourceRanges, v2Mappings); err != nil {
-			return nil, fmt.Errorf("error ensuring NLB security group rules: %w", err)
+		// Ensure SG rules for cluster-owned groups, only if the LB reconciliator finished successfully.
+		if isManagedSg && len(securityGroupIDs) > 0 {
```
kmala, line 2489: This should be based on the BYO security group right? because for other cases CCM should update the rules as its managing it?

pkg/providers/v1/aws.go
```diff
@@ -2461,15 +2479,31 @@ func (c *Cloud) EnsureLoadBalancer(ctx context.Context, clusterName string, apiS
...
+
+		// Cleanup old managed security groups and rules referencing them, if any
+		if existingLB != nil && len(existingLB.SecurityGroups) > 0 {
+			currentSGSet := sets.New(existingLB.SecurityGroups...)
+			expectedSGSet := sets.New(securityGroupIDs...)
+			detachedSGs := currentSGSet.Difference(expectedSGSet).UnsortedList()
+			if len(detachedSGs) > 0 {
+				errs := c.removeOwnedSecurityGroups(ctx, loadBalancerName, detachedSGs)
+				if len(errs) > 0 {
+					klog.Warningf("Error cleaning up detached security groups: %v", errs)
```
kmala, line 2503: This would be a leak right ?

mfbonfigli: Good point.  Yes, if this happens despite the retries it would be leaked. The security groups to delete are determined comparing the current LB state prior to the reconciliation with the desired state. Since the LB desired state has converged at this point and the only remaining item is to delete the old groups, if this fails at the next reconciliation no divergence would be detected by CCM anymore and hence in the current approach it would not have a way of determining which leftover groups remained to be deleted. 

--- 

I actually had evaluated if I could add a reconcile loop find and delete all leftover groups, however the problem is that I could not find a way to reliably identify leftover groups belonging to the service. Unfortunately NLB security groups don't carry a tag with the service name so I can't really identify them with that. 

They do carry the service name on the name, however while the service name is part of the security group name (see `GetSecurityGroupNameForNLB`), only the first `8` characters make into it:
```
fmt.Sprintf("k8s-%.8s-%.8s-%.10s", sanitizedNamespace, sanitizedName, uuid)
```

This means that any prefix-based security group query would risk of retrieving and deleting security groups for other services with a similar name. In other words if I have a service with name `myservice1` and another named `myservice2` in the same namespace they would both get the same name prefix for the security groups, so I would have no way of reliably querying for only the security groups of `myservice1`. 

The current comparison-based cleanup handles the common case correctly, it's only if that specific cleanup call fails despite retries that the leak would occur with no major side effects, so while not ideal for simplicity I am just ignoring the error, but I'm open to suggestions for alternative approaches!

pkg/providers/v1/aws.go
mfbonfigli: correct, if the service has the BYO security group annotation the `isManagedSg` will be false so rules won't be managed as determined in `ensureNLBSecurityGroup`

kmala: But this is false when existing lb's are reconciled also right https://github.com/kubernetes/cloud-provider-aws/pull/1379/changes#diff-97a31d0f193ea0b4e2c538595daf93405351054377458053a5010f6fdb97f4dcR2271 ? in those cases we should be updating the rules correct?

mfbonfigli: it's false for existing lbs essentially in three possible circumstances:

1. if sg managed mode is disabled, because in that case we don't manage security groups at all for NLBs ([ref](https://github.com/kubernetes/cloud-provider-aws/pull/1379/changes#diff-97a31d0f193ea0b4e2c538595daf93405351054377458053a5010f6fdb97f4dcR2265))
2. if sg managed mode is enabled but the existing lb has no security groups, because we can't attach SGs to it ([ref](https://github.com/kubernetes/cloud-provider-aws/pull/1379/changes#diff-97a31d0f193ea0b4e2c538595daf93405351054377458053a5010f6fdb97f4dcR2278))
3. if sg managed mode is enabled but the service has the BYO SG annotation, because we would not be managing rules in that case ([ref](https://github.com/kubernetes/cloud-provider-aws/pull/1379/changes#diff-97a31d0f193ea0b4e2c538595daf93405351054377458053a5010f6fdb97f4dcR2285))

In a nutshell the `isManagedSg` flag will be true for an existing NLB if all these conditions are satisfied:
- NLBSecurityGroupMode=Managed in the cloud config > it means that we have enabled the feature
- The NLB has already at least one security group attached > it means we can actually attach / manage SG groups for the specific NLB
- The service does not have the BYO SG annotation > it means that the user doesn't want to manage the security groups himself

kmala: In the current implementation [[ref](https://github.com/kubernetes/cloud-provider-aws/pull/1379/changes#diff-97a31d0f193ea0b4e2c538595daf93405351054377458053a5010f6fdb97f4dcL2259-L2262)] we are updating the rules of the existing security group for the NLB irrespective of managed or not. I meant that is not the case with the new changesright?

mfbonfigli: Oh I get now what you mean. Yes that's a good catch, the behavior is different. I think I could technically align it with the current behavior, just not sure if that's actually what we want, which is managing security groups without the sg managed mode being turned on, (and worse I think even changing rules if the security group is not cluster owned...) I understand it's a change in behaviour but one could argue it's a bugfix. 

Let me know what do you think!

mtulio: I am unsure if this is a bug as it was intentionally implemented following the same behavior of CLB - considering CLB manages security group by default:
1. [Manage SGs](https://github.com/kubernetes/cloud-provider-aws/blob/746099eda51cfab340f3d61ca3f10664e606c550/pkg/providers/v1/aws.go#L2653) are created by CLB
2. Managed SG Rules [are updated](https://github.com/kubernetes/cloud-provider-aws/blob/746099eda51cfab340f3d61ca3f10664e606c550/pkg/providers/v1/aws.go#L2675-L2677) only when [managed (setupSG)](https://github.com/kubernetes/cloud-provider-aws/blob/746099eda51cfab340f3d61ca3f10664e606c550/pkg/providers/v1/aws.go#L2011-L2017) was created by controller.

Then the managed SG in NLB, indeed it is discovering [SGs from NLBs](https://github.com/kubernetes/cloud-provider-aws/blob/746099eda51cfab340f3d61ca3f10664e606c550/pkg/providers/v1/aws.go#L2272-L2283), and updating(but [skipped in unmanaged](https://github.com/kubernetes/cloud-provider-aws/blob/746099eda51cfab340f3d61ca3f10664e606c550/pkg/providers/v1/aws.go#L2323-L2326)) regardless it is managed or not, but considering:
1) SGs can be created only in managed mode on NLB, as BYOSG annotation is not supported so far.
2) Users can't manually (outside controller) add SGs to NLBs that was not created with one

How that path will be exercised? 🤔 

WRT the approach so far, it looks like the current suggestion (restrict to managed && not update BYOSG rules) will be safe and aligned with existing CLB as well with ALBC BYOSG practice, wdyt?

mfbonfigli: Yes I agree, if I'm not wrong the change in behavior would manifest only in this scenario:

- Existing CCM Runs with NLBSecurityGroupMode=Managed
- User provisions a Service with a NLB, CCM creates the NLB with a managed security group attached
- User stops CCM
- User edits cluster owned NLB security groups (changing rules / attaching / detaching groups)
- User edits CCM cloud config to remove NLBSecurityGroupMode=Managed
- User restarts CCM to have it refresh the configuration

In this scenario where the user tampered with CCM and the cloud infrastructure:
- existing behavior: ccm will manipulate the security groups even if NLBSecurityGroupMode=Managed is disabled and even if the security group now attached is not cluster owned
- new behavior (this PR): CCM won't touch the security groups at all

I personally believe that this PR handles correctly this edge case scenario, and I would not consider this as a breaking change as for what I can think of it only manifests if the user should has tampered with the cluster and cloud configurations in a way that is not supported or documented.

mfbonfigli:
@kmala I have replied to your comments, please take a look when you can. Thanks!

pkg/providers/v1/aws.go
mtulio: > Unfortunately NLB security groups don't carry a tag

Do you think it could be helpful. In the initial proposal it was suggested but removed before final version. I still wonder how we'll do it since the controller is event-drive reconciliation model.

Anyway, considering:
1) `removeOwnedSecurityGroups()` ensures a) rules references are deleted, b) owned SG is deleted without returning premature error with results of those; and
2) `deleteSecurityGroupsWithBackoff()` implements backoff to retry accordingly preventing premature errors due transient or even delays caused by attached SG from LBs, such as `DependencyViolation`

I think this risk of leaking here is mitigated, right?

pkg/providers/v1/aws.go
mtulio: yes, thanks for pointing out. that makes sense to me.

PR #1381, opened 2026-04-14
DNM/test/e2e: export load balancer helpers

mtulio:
**What type of PR is this?**

/kind cleanup
/kind design
/kind failing-test

This PR introduces exports for e2e library to be extended in test frameworks of layered controllers.

Fixes #

```release-note

```


mtulio:
Covnerting to reguar PR just to trigger e2e

mtulio:
Node discovery for hairpinning traffic was failing:

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,277 @@
...
+// // discoverClusterTag discovers the cluster tag from a cluster.
+// // The discover is done by looking up the EC2 instance tags with tag:Name prefix kubernetes.io/cluster.
+// // The EC2 Instance ID is discovered from a cluster node object.
+// // The cluster ID, VPC ID and cluster tag are discovered from the EC2 instance tags.
+// // If is any error is found, the function returns an error.
+// func (h *E2ETestHelperAWS) discoverClusterTag(cs clientset.Interface) error {
+// 	nodes, err := cs.CoreV1().Nodes().List(h.ctx, metav1.ListOptions{})
+// 	if err != nil {
+// 		return fmt.Errorf("failed to list nodes: %v", err)
+// 	}
```
mfbonfigli, line 95: Consider to remove commented out code blocks (here and in other parts of the PR)

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,277 @@
...
+	cfg, err := config.LoadDefaultConfig(ctx)
+	framework.ExpectNoError(err, "unable to load AWS config")
+
+	// Configure custom retryer to handle transient AWS API errors and credential failures.
+	// AWS API limits for ELB are generous (400 TPS for DescribeLoadBalancers),
+	// so aggressive retries are safe and necessary for CI stability.
+	customRetryer := retry.NewStandard(func(o *retry.StandardOptions) {
+		o.MaxAttempts = 10              // Handle IMDS timeouts and transient errors
+		o.MaxBackoff = 30 * time.Second // Cap backoff to avoid excessive wait
+	})
```
mfbonfigli, line 61: Consider storing this inside the struct, as it could be reused e.g. in `GetLoadBalancerFromDNSNameWithRetry` to detect non-retryable errors (see related comment below).

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,277 @@
...
+type E2ETestHelperAWS struct {
+	ctx context.Context
+
+	ec2Client   *ec2.Client
+	elbClient   *elb.Client
+	elbv2Client *elbv2.Client
+}
+
+// NewAWSHelper creates a new AWS helper with configured clients
+func NewAWSHelper(ctx context.Context, cs clientset.Interface) (*E2ETestHelperAWS, error) {
```
mfbonfigli, line 51: `cs clientset.Interface` seems unused, consider to remove it.

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,277 @@
...
+// }
+
+// GetLBTargetCount verifies the number of registered targets for a given LBv2 DNS name matches the expected count.
+// The steps includes:
+// - Get Load Balancer ARN from DNS name extracted from service Status.LoadBalancer.Ingress[0].Hostname
+// - List listeners for the load balancer
+// - Get target groups attached to listeners
+// - Count registered targets in target groups
+// - Verify count matches number of worker nodes
+func (h *E2ETestHelperAWS) GetLBTargetCount(lbDNSName string, expectedTargets int) error {
```
mfbonfigli, line 166: If it's a verification/assertion function I would call it `ExpectLBTargetCount` instead of `GetLBTargetCount`, however I wonder if it's however better to just return the number and let the caller do the assertion outside this method. In that case the name could stay as it is but the logic needs to change to return the targets count.

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to describe listeners: %v", err)
+	}
+
+	// Get target groups attached to listeners
+	targetGroupARNs := map[string]struct{}{}
+	for _, listener := range listenersOut.Listeners {
+		if len(targetGroupARNs) > 0 {
+			break
+		}
```
mfbonfigli, line 187: With this I believe we are only detecting the groups of the first listener with target groups associated, is this intentional? I believe different listeners could have a different number of target groups (maybe there are assumptions we can make in the setup here and this is not an issue, just asking for clarification).

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,277 @@
...
+func (h *E2ETestHelperAWS) GetLoadBalancerFromDNSName(lbDNSName string) (*elbv2types.LoadBalancer, error) {
+	var foundLB *elbv2types.LoadBalancer
+	framework.Logf("describing load balancers with DNS %s", lbDNSName)
+
+	paginator := elbv2.NewDescribeLoadBalancersPaginator(h.elbv2Client, &elbv2.DescribeLoadBalancersInput{})
+	for paginator.HasMorePages() {
+		page, err := paginator.NextPage(h.ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to describe load balancers: %v", err)
+		}
```
mfbonfigli, line 228: I would do error wrapping or even just return the error, this way it would be easier for the caller to check for specific errors, so either:

`return nil, fmt.Errorf("failed to describe load balancers: %w", err)` (note %w instead of %v)

or, more simply,

`return nil, err`

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,277 @@
...
+		if foundLB != nil {
+			break
+		}
+	}
+
+	if foundLB == nil {
+		return nil, fmt.Errorf("no load balancer found with DNS name: %s", lbDNSName)
+	}
+
+	return foundLB, nil
```
mfbonfigli, line 248: would slightly simplify with this syntax which avoids accessing the array with an index. Also we could return early if LB is found. With the changes this block becomes:

```
[...]

    for _, lb := range page.LoadBalancers {
      	if aws.ToString(lb.DNSName) == lbDNSName {
      		framework.Logf("found load balancer with DNS %s", aws.ToString(foundLB.DNSName))
      		return &lb, nil
      	}
    }
}
return nil, fmt.Errorf("no load balancer found with DNS name: %s", lbDNSName)
```

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,277 @@
...
+		var err error
+		foundLB, err = h.GetLoadBalancerFromDNSName(lbDNSName)
+		if err != nil {
+			return false, nil
+		}
+		return true, nil
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to find load balancer %s within timeout: %v", lbDNSName, err)
```
mfbonfigli, line 278: Would use `%w` to wrap the error as `%v` breaks the error chain, unless we have a reason to do so.

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,277 @@
...
+		Steps:    22, // Covers ~9.5 minutes
+		Cap:      30 * time.Second,
+	}
+
+	err := wait.ExponentialBackoffWithContext(ctx, backoff, func(ctx context.Context) (bool, error) {
+		var err error
+		foundLB, err = h.GetLoadBalancerFromDNSName(lbDNSName)
+		if err != nil {
+			return false, nil
+		}
```
mfbonfigli, line 273: Here we should try to treat differently retryable vs non-retryable errors, else we'd risk getting stuck for 9 minutes attempting to retry an error that is not transient (e.g. IAM permission issue). I think one way could be to do something like:

```
if err != nil && h.retryer.IsErrorRetryable(err) { 
    return false, nil
}
if err != nil {
   return true, err
}
return true, nil
```

where `h.retryer` is the retryer stored in the struct (see comment above in the PR)

tests/e2e/helper_e2e.go
```diff
@@ -0,0 +1,859 @@
...
+	}, nil
+}
+
+func (e2e *E2ETestHelper) GetAWSHelper() *E2ETestHelperAWS {
+	return e2e.awsHelper
+}
+
+func (e2e *E2ETestHelper) GetContext() context.Context {
+	return e2e.ctx
+}
```
mfbonfigli, line 100: Is this needed? Not a common pattern to store context inside structs and also to have getters for it.

tests/e2e/helper_e2e.go
```diff
@@ -0,0 +1,859 @@
...
+
+// NewE2ETestHelper creates a new E2ETestHelper instance.
+func NewE2ETestHelper(ctx context.Context, cs clientset.Interface) (*E2ETestHelper, error) {
+	var cancel context.CancelFunc
+	if ctx == nil || ctx == context.Background() {
+		// Create a context with a reasonable timeout for e2e tests when no context is provided
+		// E2E tests can take several minutes for load balancer provisioning and configuration
+		ctx, cancel = context.WithTimeout(context.Background(), 25*time.Minute)
+		_ = cancel // We'll let the test framework handle cleanup
+	}
```
mfbonfigli, line 73: Generally one passes the context to individual requests, rather than storing it in a struct which is not a recommended pattern in go. 

Would it be possible to add it as a parameter to the individual functions that need it instead (`discoverClusterWorkerNode`, `inClusterTestReachableHTTP`, etc)?

tests/e2e/helper_e2e.go
```diff
@@ -0,0 +1,859 @@
...
+func (e2e *E2ETestHelper) inClusterTestReachableHTTP(target string, targetPort int) error {
+	podName := "http-test-pod"
+
+	// Enhanced curl configuration for better resilience
+	// Total timeout calculation: 30 retries * 30s delay + 15min curl max time = ~25 minutes
+	// This aligns with the 25-minute polling timeout below
+	default_max_time := "900" // 15 minutes max for curl operations
+	if os.Getenv("CCM_E2E_OVERRIDE_MAX_TIME") != "" {
+		default_max_time = os.Getenv("CCM_E2E_OVERRIDE_MAX_TIME")
+	}
```
mfbonfigli, line 315: nit, slightly less redundant:

```
defaultMaxTime := "900"
if maxTime := os.Getenv("CCM_E2E_OVERRIDE_MAX_TIME"); maxTime != "" {
  defaultMaxTime = maxTime
}
```

tests/e2e/helper_aws.go
mtulio: I think the later would be better for an API perspective, so that we can just get the counters and callers would assert it. I will make the change. good catch

tests/e2e/helper_aws.go
mtulio: > only detecting the groups of the first listener with target groups associated, is this intentional?

Yes it is, there was logic embedded to this function to satisfy the test it is doing ([ensure node filter](https://github.com/kubernetes/cloud-provider-aws/blob/master/tests/e2e/loadbalancer.go#L120)), but I think as long we are creating a function to perform this discovery (dnsName -> lb -> listener -> target), I think it would be nice to make it generic with parameter to return targets from a given listener and targetPort.

The new version will rename this function to GetLBTargets().

tests/e2e/helper_e2e.go
mtulio: Good catch, makes sense totally. The calls requiring context have been updated.

mtulio:
Updated the worker node lookup function to get compatibility with existing logic:

mtulio:
local tests passing

```
$ ./e2e.test --ginkgo.v --ginkgo.focus="loadbalancer"
....
• [219.703 seconds]

Ran 5 of 8 Specs in 880.678 seconds
SUCCESS! -- 5 Passed | 0 Failed | 0 Pending | 3 Skipped

```

tests/e2e/helper_e2e.go
```diff
@@ -0,0 +1,883 @@
...
+
+	return svc
+}
+
+// cleanup cleans up the e2e resources.
+func (e *E2ETestHelper) cleanup(ctx context.Context) {
+	framework.Logf("Cleaning up e2e resources")
+	// TODO cleanup resources created by the test helper
+}
```
mfbonfigli, line 144: Do we need this or does it need to be implemented?

tests/e2e/loadbalancer.go
```diff
@@ -112,20 +104,26 @@ var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
...
 			},
-			hookPostServiceCreate: func(cfg *e2eTestConfig) {
+			hookPostServiceCreate: func(ctx context.Context, cfg *E2ETestHelper) {
 				framework.Logf("running hook post-service-create to validate the number of targets in the load balancer selected")
 				if len(cfg.svc.Status.LoadBalancer.Ingress) == 0 {
 					framework.Failf("No ingress found in LoadBalancer status for service %s/%s", cfg.svc.Namespace, cfg.svc.Name)
 				}
 				lbDNS := cfg.svc.Status.LoadBalancer.Ingress[0].Hostname
-				framework.ExpectNoError(getLBTargetCount(cfg.ctx, lbDNS, cfg.nodeCount), "AWS LB target count validation failed")
+				// TODO expected lbDNS not empty
```
mfbonfigli, line 121: nit: remove comment or add validation of lbDNS

tests/e2e/loadbalancer.go
```diff
@@ -181,13 +181,13 @@ var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
...
 				}
 
-				elbClient, err := getAWSClientLoadBalancer(e2e.ctx)
-				framework.ExpectNoError(err, "failed to create AWS ELB client")
-
 				// DescribeLoadBalancers API doesn't support filtering by DNS name directly
 				// Use AWS SDK paginator to search through all load balancers
-				foundLB, err := getAWSLoadBalancerFromDNSName(e2e.ctx, elbClient, hostAddr)
-				framework.ExpectNoError(err, "failed to find load balancer with DNS name %s", hostAddr)
+				foundLB, err := e2e.GetAWSHelper().GetLoadBalancerFromDNSNameWithRetry(ctx, hostAddr)
```
mfbonfigli, line 186: `GetLoadBalancerFromDNSNameWithRetry` is marked as deprecated (see comment to the function), if so we should not use it in new code.

tests/e2e/helper_e2e.go
```diff
@@ -0,0 +1,883 @@
...
+					Name:    "curl",
+					Image:   imageutils.GetE2EImage(imageutils.Agnhost),
+					Command: []string{"curl"},
+					Args:    curlArgs,
+					SecurityContext: &v1.SecurityContext{
+						AllowPrivilegeEscalation: aws.Bool(false),
+						Capabilities: &v1.Capabilities{
+							Drop: []v1.Capability{"ALL"},
+						},
+						ReadOnlyRootFilesystem: aws.Bool(true),
```
mfbonfigli, line 371: nit: I don't particularly like using the AWS SDK helpers (`aws.Bool` in this case) outside AWS-related blocks of code, it adds unnecessary coupling.

tests/e2e/helper_e2e.go
```diff
@@ -0,0 +1,883 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
mfbonfigli, line 2: Should be `2026`

tests/e2e/helper_e2e.go
```diff
@@ -0,0 +1,883 @@
...
+// Acknowledgement:
+// Documentation generated by Cursor AI, reviewed by Human.
+// Function generated by Human, reviewed and verbosity increased by Cursor AI.
+func (e2e *E2ETestHelper) inClusterTestReachableHTTP(ctx context.Context, target string, targetPort int) error {
+	podName := "http-test-pod"
+
+	// Enhanced curl configuration for better resilience
+	// Total timeout calculation: 30 retries * 30s delay + 15min curl max time = ~25 minutes
+	// This aligns with the 25-minute polling timeout below
+	defaultMaxTime := "1200"
```
mfbonfigli, line 335: nit: would extract this 1200 magic value to a constant similarly to what was done in helper_aws.go around lines 38-61

tests/e2e/helper_e2e.go
```diff
@@ -0,0 +1,883 @@
...
+func (e2e *E2ETestHelper) GetAWSHelper() *E2ETestHelperAWS {
+	return e2e.awsHelper
+}
+
+func (e2e *E2ETestHelper) GetKubeClient() clientset.Interface {
+	return e2e.kubeClient
+}
+
+func (e2e *E2ETestHelper) GetServicePodPort() int32 {
+	return int32(e2e.cfgPodPort)
```
mfbonfigli, line 92: isn't e2e.cfgPodPort already a `int32`? It used to be `uint16` but the type definition was changed in this PR

tests/e2e/helper_e2e.go
```diff
@@ -0,0 +1,883 @@
...
+// Behavior:
+// - The function creates a client pod with a curl-based container to perform the HTTP request.
+// - The client pod is scheduled on the specified node using node affinity.
+// - Logs are periodically collected during the client pod's execution for troubleshooting.
+// - Events are inspected if the client pod remains in a pending state for too long.
+// - The function waits for the client pod to complete and inspects its exit code to determine success or failure.
+//
+// Acknowledgement:
+// Documentation generated by Cursor AI, reviewed by Human.
+// Function generated by Human, reviewed and verbosity increased by Cursor AI.
```
mfbonfigli, line 328: Is this acknowledgement needed in code?

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,336 @@
...
+
+// GetELBV2Client returns the ELBV2 client.
+func (h *E2ETestHelperAWS) GetELBV2Client() *elbv2.Client {
+	return h.elbv2Client
+}
+
+// GetLBTargets returns the targets for a given LB DNS name, listener port, and target port.
+// This performs a single check without retry. For waiting until targets are registered,
+// use WaitForLBTargets instead.
+func (h *E2ETestHelperAWS) GetLBTargets(ctx context.Context, lbDNSName string, listenerPort, targetPort int32) ([]string, error) {
```
mfbonfigli, line 148: The `targetPort` parameter appears to be unused

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,336 @@
...
+	var targets []string
+	var lastErr error
+
+	framework.Logf("Waiting for LB %s to have at least %d targets (timeout: %v) on port %d", lbDNSName, minTargets, timeout, listenerPort)
+
+	// Use fixed-interval polling for state convergence (Kubernetes standard pattern)
+	// Poll immediately first (true), then every 5 seconds
+	err := wait.PollUntilContextTimeout(ctx, DefaultTargetPollInterval, timeout, true, func(ctx context.Context) (bool, error) {
+		var err error
+		targets, err = h.GetLBTargets(ctx, lbDNSName, listenerPort, targetPort)
```
mfbonfigli, line 217: Since `targetPort` is unused in `GetLBTargets` it could also be removed in this function signature.

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// GetELBV2Client returns the ELBV2 client.
+func (h *E2ETestHelperAWS) GetELBV2Client() *elbv2.Client {
+	return h.elbv2Client
+}
+
+// GetLBTargets returns the targets for a given LB DNS name, listener port, and target port.
+// This performs a single check without retry. For waiting until targets are registered,
+// use WaitForLBTargets instead.
```
mfbonfigli, line 147: It says it does not do any retry, but then it uses `GetLoadBalancerFromDNSNameDefaultTimeout` which does do retries, so the comment is not accurate, would clarify that the lookup does retry or change the code to truly be retry-free.

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,336 @@
...
+		LoadBalancerArn: aws.String(lbARN),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to describe listeners: %v", err)
+	}
+
+	framework.Logf("Found %d listeners for load balancer, checking for listener port %d", len(listenersOut.Listeners), int32(listenerPort))
+	targetGroupARNs := map[string]struct{}{}
+	for _, listener := range listenersOut.Listeners {
+		if aws.ToInt32(listener.Port) == int32(listenerPort) {
```
mfbonfigli, line 165: `listenerPort` is already `int32`, so type cast is redundant

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,336 @@
...
+	return h.elbv2Client
+}
+
+// GetLBTargets returns the targets for a given LB DNS name, listener port, and target port.
+// This performs a single check without retry. For waiting until targets are registered,
+// use WaitForLBTargets instead.
+func (h *E2ETestHelperAWS) GetLBTargets(ctx context.Context, lbDNSName string, listenerPort, targetPort int32) ([]string, error) {
+	foundLB, err := h.GetLoadBalancerFromDNSNameDefaultTimeout(ctx, lbDNSName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get load balancer from DNS name: %v", err)
```
mfbonfigli, line 151: nit (was already present before): it is good practice to error wrap with `%w` instead of `%v`, consider fixing.

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	lbARN := aws.ToString(foundLB.LoadBalancerArn)
+
+	listenersOut, err := h.elbv2Client.DescribeListeners(ctx, &elbv2.DescribeListenersInput{
+		LoadBalancerArn: aws.String(lbARN),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to describe listeners: %v", err)
```
mfbonfigli, line 159: nit (was already present before): it is good practice to error wrap with `%w` instead of `%v`, consider fixing.

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,336 @@
...
+// AWS API limits for ELB are generous (400 TPS for DescribeLoadBalancers),
+// so aggressive retries are safe and necessary for CI stability.
+//
+// For rate-limited environments or shared test accounts, adjust MaxAttempts and MaxBackoff:
+//
+//	helper, err := NewAWSHelperWithOptions(ctx, E2ETestHelperAWSOptions{
+//	    MaxAttempts: 5,
+//	    MaxBackoff: 10 * time.Second,
+//	})
+func NewAWSHelperWithOptions(ctx context.Context, opts E2ETestHelperAWSOptions) (*E2ETestHelperAWS, error) {
```
mfbonfigli, line 110: nit, just an alternative approach to `E2ETestHelperAWSOptions`:

It is quite common in go to use the "Functional Options" pattern to pass optional configurations, and it might work well here as well. For example instead of having a `NewAWSHelperWithOptions` and a `E2ETestHelperAWS` struct, we could have:

```
type E2ETestHelperAWSOpt func(*E2ETestHelperAWS)

func NewAWSHelper(ctx context.Context, opts ...E2ETestHelperAWSOpt) (*E2ETestHelperAWS, error) {
	h := &E2ETestHelperAWS {}
    for _, optFn := range opts {
       optFn(h)
    }

   // rest of code down here
}

func WithMaxAttempts(attempts int) func(*E2ETestHelperAWS) {
  return func(h *E2ETestHelperAWS) {
    h.MaxAttempts = attempts
  }
}


func WithMaxBackoff(maxBackoff time.Duration) func(*E2ETestHelperAWS) {
  return func(h *E2ETestHelperAWS) {
    h.MaxBackoff = maxBackoff
  }
}
```

With this pattern you have only 1 constructor and the user can pass only the configurations he wants to edit. 
For example to get the default one would do:
```
h, err := NewAWSHelper(ctx)
```

And instead to customize only one of the settings would do:

```
h, err := NewAWSHelper(ctx, WithMaxAttempts(5))
```

I think this approach is less code and more flexible.

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,336 @@
...
+
+	// Use fixed-interval polling for state convergence (Kubernetes standard pattern)
+	// Poll immediately first (true), then every 5 seconds
+	err := wait.PollUntilContextTimeout(ctx, DefaultTargetPollInterval, timeout, true, func(ctx context.Context) (bool, error) {
+		var err error
+		targets, err = h.GetLBTargets(ctx, lbDNSName, listenerPort, targetPort)
+		if err != nil {
+			// Log but continue polling - target groups might not be ready yet
+			framework.Logf("error getting LB targets (will retry): %v", err)
+			lastErr = err
```
mfbonfigli, line 221: we could consider isolating and treating differently retryable vs non-retryable errors here too.

tests/e2e/helper_aws.go
```diff
@@ -0,0 +1,336 @@
...
+//
+//	helper, err := NewAWSHelperWithOptions(ctx, E2ETestHelperAWSOptions{
+//	    MaxAttempts: 5,
+//	    MaxBackoff: 10 * time.Second,
+//	})
+func NewAWSHelperWithOptions(ctx context.Context, opts E2ETestHelperAWSOptions) (*E2ETestHelperAWS, error) {
+	opts.ApplyDefaults()
+
+	cfg, err := config.LoadDefaultConfig(ctx)
+	framework.ExpectNoError(err, "unable to load AWS config")
```
mfbonfigli, line 114: nit: Should we do test assertions inside the helpers (here and in other parts of the PR)? Not a strong opinion, just generally I'd do assertions inside the test function.

mtulio:
Interim update, this PR is on hold for now. I will return on this as soon as possible addressing the review comments.

PR #1388, opened 2026-04-23
Fix route-controller by fanning out getInstancesByIDs through the batcher

### LINKED ISSUE #1351: Route reconciliation fails with "expected to receive a single instance only" when multiple instance IDs are passed to describeInstanceBatcher
**What happened**:
Route reconciliation fails with the error "expected to receive a single instance only" when multiple instance IDs are passed to describeInstanceBatcher.

Route reconciliation should handle multiple instance IDs without failing.

1. Use AWS cloud-controller-manager with route-based pod networking.
2. Ensure the route table contains routes pointing to multiple instances.
3. Route reconciliation fails with:
   "Couldn't reconcile node routes: error listing routes: expected to receive a single instance only, found N".

The issue occurs because `getInstancesByIDs` passes multiple instance IDs to `describeInstanceBatcher.DescribeInstances`, while the batcher implementation expects exactly one instance ID per request.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.34.1 && v1.35.0
- Cloud provider or hardware configuration: aws ec2
- OS (e.g. from /etc/os-release): Amazon Linux 9 / RHEL9-based
- Kernel (e.g. `uname -a`): 5.14.0-229.el9.x86_64
- Install tools: Deckhouse
- Others: simple-bridge CNI with route-based pod networking

/kind bug

babyhuey:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it:**

`describeInstanceBatcher.DescribeInstances` requires an input with exactly one instance ID:

https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/describe_instance_batch.go#L56-L58

The batcher's `BatchExecutor` is what aggregates concurrent single-instance inputs back into one `DescribeInstances` API call. But `getInstancesByIDs` was passing the full slice of instance IDs as a single input, which trips that guard and errors out any time more than one instance is looked up.

The most visible consequence is that in clusters using kubenet / route-based pod networking, the route-controller's initial `ListRoutes` call always needs to look up every node and therefore always fails with `expected to receive a single instance only, found N`. No pod-CIDR VPC routes get created and source-dest-check is left enabled on new workers, so pods on new nodes have no connectivity — see #1351.

This PR fans out one goroutine per instance ID, each passing a single-instance input into the batcher. The batcher then coalesces those concurrent calls into one underlying `DescribeInstances` API call, preserving the original batching behavior while honoring the single-instance contract.

**Which issue(s) this PR fixes:**

Fixes #1351

**Special notes for your reviewer:**

- `go build ./...` clean
- `go test ./pkg/providers/v1/...` passes
- The only other caller of `getInstancesByIDs` (`aws.go:3412`) already passes a single-ID slice, so its behavior is unchanged.

```release-note
Fix route-controller and other code paths that resolve multiple instances in one call. `getInstancesByIDs` previously passed the full slice into `describeInstanceBatcher.DescribeInstances`, which rejects inputs with more than one ID, breaking the route-controller in kubenet/route-based clusters.
```

### DIFF
--- a/pkg/providers/v1/describe_instance_batch.go
+++ b/pkg/providers/v1/describe_instance_batch.go
@@ -18,6 +18,7 @@ package aws
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/providers/v1/aws.go
```diff
@@ -3430,22 +3431,40 @@ func (c *Cloud) getInstancesByIDs(ctx context.Context, instanceIDs []string) (ma
 		return instancesByID, nil
 	}
 
-	request := &ec2.DescribeInstancesInput{
-		InstanceIds: instanceIDs,
-	}
-
-	instances, err := c.describeInstanceBatcher.DescribeInstances(ctx, request)
```
kmala, line 3437: I think we should add the functionality in the batcher instead of the client making complex logic

babyhuey:
Thanks @kmala, that makes sense. I moved the fan out into `describeInstanceBatcher.DescribeInstances` itself in the latest push, so multi ID inputs are split into one `batcher.Add` per ID inside the batcher and the caller in `getInstancesByIDs` is back to the simple single call. Added a regression test (`TestDescribeInstanceBatchingWithMultipleInstanceIDs`) covering the multi ID path. Local `go test ./pkg/providers/v1/...`, gofmt, golint, and vet all pass.

rifelpet:
@kmala any chance you can take another look at this?

Many of kops' kubenet e2e jobs are failing because of this bug. For example:

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/e2e-kops-grid-kubenet-al2023-k34/2065547502785925120

and the CCM logs in the job artifacts:

https://storage.googleapis.com/kubernetes-ci-logs/logs/e2e-kops-grid-kubenet-al2023-k34/2065547502785925120/artifacts/cluster-info/kube-system/aws-cloud-controller-manager-8h7tq/aws-cloud-controller-manager.log

```
E0612 21:45:43.643908       1 route_controller.go:118] Couldn't reconcile node routes: error listing routes: expected to receive a single instance only, found 3
E0612 21:45:53.649833       1 route_controller.go:118] Couldn't reconcile node routes: error listing routes: expected to receive a single instance only, found 3
E0612 21:46:03.667509       1 route_controller.go:118] Couldn't reconcile node routes: error listing routes: expected to receive a single instance only, found 3
```

hakman:
Thanks @kmala! Any objections to cherry-picking this down to 1.34?

kmala:
> Thanks @kmala! Any objections to cherry-picking this down to 1.34?

No. Since this is a bug we should be okay to cherry pick

PR #1397, opened 2026-05-04
feat: Add ALPN policy annotation for NLB TLS listeners

### LINKED ISSUE #1309: Provide annotation for ALPN profile on NLB

I would like the ability to specify an ALPN profile on NLBs.  Something to the lines of an annotations like `service.beta.kubernetes.io/aws-load-balancer-alpn-policy`

**Why is this needed**:

Currently if an ALPN policy is desired on a cloud provider provisioned NLB the ALPN policy must be set manually after creation.

/kind feature

alimx07:
**What type of PR is this?**
/kind feature

Adds `service.beta.kubernetes.io/aws-load-balancer-alpn-policy` annotation to configure the ALPN policy on NLB TLS listeners. Valid values are `HTTP1Only`, `HTTP2Only`, `HTTP2Optional`, `HTTP2Preferred`, and `None`. Applied only when `aws-load-balancer-ssl-cert` is also set. Removing the annotation from an existing service explicitly resets the listener to `None`.

Fixes #1309

```release-note
Add `service.beta.kubernetes.io/aws-load-balancer-alpn-policy` annotation to configure the ALPN policy on NLB TLS listeners. Valid values: `HTTP1Only`, `HTTP2Only`, `HTTP2Optional`, `HTTP2Preferred`, `None`. Requires `aws-load-balancer-ssl-cert` to be set.
```


pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -113,6 +113,7 @@ type nlbPortMapping struct {
 
 	SSLCertificateARN string
 	SSLPolicy         string
+	ALPNPolicy        string
```
kmala, line 116: Since the listener supports multiple ALPN policies, we should also support multiple i think?

alimx07: Hi @kmala , While that sounds reasonable, the AWS SDK actually only allows one ALPN policy per listener as they state [here](https://github.com/aws/aws-sdk-go-v2/blob/dc2d13fa6f1db25f1c6d804567e1ecfcdff4f040/service/elasticloadbalancingv2/api_op_CreateListener.go#L59-L76). So I though of keeping it as single string for simplicity. Ofc, keeping it slice with one string will do the job also.

alimx07:
Hi @kmala, If you have time, could you look at this again. Thanks in advance!

pkg/providers/v1/aws_loadbalancer.go
kmala: interesting that its a list but only single value is supported.

PR #1447, opened 2026-05-15
apply annotation tags to NLB & listeners on update

### LINKED ISSUE #1334: Tags are not applied to NLB post creation
**What happened**:

Apply `"service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags"` on a live `NLB`

See the `tags` on the AWS console `UI` 

I seems like `ensureLoadBalancerv2` unlike `ensureLoadBalancer` doesnt have:

```
		{
			// Add additional tags
			klog.V(2).Infof("Creating additional load balancer tags for %s", loadBalancerName)
			tags := getKeyValuePropertiesFromAnnotation(annotations, ServiceAnnotationLoadBalancerAdditionalTags)
			if len(tags) > 0 {
				err := c.addLoadBalancerTags(loadBalancerName, tags)
				if err != nil {
					return nil, fmt.Errorf("unable to create additional load balancer tags: %v", err)
				}
			}
		}
```

so it doesnt apply the tags after the NLB is created

alimx07:
**What type of PR is this?**

/kind bug

service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags was only applied to the NLB at creation time (via `CreateLoadBalancer` input tags). Annotation changes never propagated on later reconciles, and removed tags were never cleaned up.

This reconciles additional tags on every `ensureLoadBalancerv2` update across the load balancer, all surviving listeners, and target groups ,adding/updating to match the annotation and removing keys the controller previously applied but no longer wants.

Two tag keys are involved:

- `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags` : existing user annotation, now propagated on update.
- `kubernetes.io/cloud-controller/managed-tags` : new marker tag recording which keys the controller owns, so removed keys are reconciled away while out-of-band and system tags are preserved.

Fixes #1334 

```release-note
Propagate `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags` to the NLB, its listeners, and target groups on update; previously tags were only applied to the load balancer at creation. Tag removals are now reconciled, and a `kubernetes.io/cloud-controller/managed-tags` marker preserves tags added out-of-band.
```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -86,6 +86,10 @@ const ProviderName = "aws"
 // services. Used currently for ELBs only.
 const TagNameKubernetesService = "kubernetes.io/service-name"
 
+// TagNameKubernetesManagedTags records the additional-tag keys the controller applied from
+// ServiceAnnotationLoadBalancerAdditionalTags, so removed keys can be reconciled away.
+const TagNameKubernetesManagedTags = "kubernetes.io/cloud-controller/managed-tags"
+
 // TagNameSubnetInternalELB is the tag name used on a subnet to designate that
 // it should be used for internal ELBs
 const TagNameSubnetInternalELB = "kubernetes.io/role/internal-elb"
@@ -354,6 +358,8 @@ type ELB interface {
 // ELBV2 is a simple pass-through of AWS' ELBV2 client interface, which allows for testing
 type ELBV2 interface {
 	AddTags(ctx context.Context, input *elbv2.AddTagsInput, optFns ...func(*elbv2.Options)) (*elbv2.AddTagsOutput, error)
+	RemoveTags(ctx context.Context, input *elbv2.RemoveTagsInput, optFns ...func(*elbv2.Options)) (*elbv2.RemoveTagsOutput, error)
+	DescribeTags(ctx context.Context, input *elbv2.DescribeTagsInput, optFns ...func(*elbv2.Options)) (*elbv2.DescribeTagsOutput, error)
 
 	CreateLoadBalancer(ctx context.Context, input *elbv2.CreateLoadBalancerInput, optFns ...func(*elbv2.Options)) (*elbv2.CreateLoadBalancerOutput, error)
 	DescribeLoadBalancers(ctx context.Context, input *elbv2.DescribeLoadBalancersInput, optFns ...func(*elbv2.Options)) (*elbv2.DescribeLoadBalancersOutput, error)
@@ -1217,6 +1223,131 @@ func (c *Cloud) describeLoadBalancerv2(ctx context.Context, name string) (*elbv2
 	}
 
 	return nil, fmt.Errorf("NLB '%s' could not be found", name)
+}
+
+// addLoadBalancerTagsv2 tags a single ELBv2 resource (LB, listener, target group, etc.).
+func (c *Cloud) addLoadBalancerTagsv2(ctx context.Context, resourceARN string, requested map[string]string) error {
+	var tags []elbv2types.Tag
+	for k, v := range requested {
+		tags = append(tags, elbv2types.Tag{
+			Key:   aws.String(k),
+			Value: aws.String(v),
+		})
+	}
+
+	_, err := c.elbv2.AddTags(ctx, &elbv2.AddTagsInput{
+		ResourceArns: []string{resourceARN},
+		Tags:         tags,
+	})
+	if err != nil {
+		return fmt.Errorf("error adding tags NLB resource %s: %w", resourceARN, err)
+	}
+	return nil
+}
+
+// removeLoadBalancerTagsv2 removes a set of tag keys from a single ELBv2 resource.
+func (c *Cloud) removeLoadBalancerTagsv2(ctx context.Context, resourceARN string, keys []string) error {
+	_, err := c.elbv2.RemoveTags(ctx, &elbv2.RemoveTagsInput{
+		ResourceArns: []string{resourceARN},
+		TagKeys:      keys,
+	})
+	if err != nil {
+		return fmt.Errorf("error removing tags from NLB resource %s: %w", resourceARN, err)
+	}
+	return nil
+}
+
+// parseManagedTagKeys parses the comma-separated list stored in the
+// TagNameKubernetesManagedTags marker tag into a set of keys.
+func parseManagedTagKeys(value string) sets.Set[string] {
+	keys := sets.New[string]()
+	for _, k := range strings.Split(value, ",") {
+		if k != "" {
+			keys.Insert(k)
+		}
+	}
+	return keys
+}
+
+// ensureLoadBalancerResourceTagsV2 makes sure every ELBv2 resource in resourceARNs carries the
+// desired tags. Tags are added or updated to match desired.
+func (c *Cloud) ensureLoadBalancerResourceTagsV2(ctx context.Context, resourceARNs []string, desired map[string]string) error {
+	if len(desired) == 0 || len(resourceARNs) == 0 {
+		return nil
+	}
+
+	// Keys the controller currently owns from the annotation.
+	var currentManaged sets.Set[string]
+	if keys, ok := desired[TagNameKubernetesManagedTags]; ok {
+		currentManaged = parseManagedTagKeys(keys)
+	}
+
+	// DescribeTags accepts at most 20 resource ARNs per call.
+	const describeTagsMaxARNs = 20
+
+	for start := 0; start < len(resourceARNs); start += describeTagsMaxARNs {
+		end := start + describeTagsMaxARNs
+		if end > len(resourceARNs) {
+			end = len(resourceARNs)
+		}
+
+		out, err := c.elbv2.DescribeTags(ctx, &elbv2.DescribeTagsInput{
+			ResourceArns: resourceARNs[start:end],
+		})
+		if err != nil {
+			return fmt.Errorf("error describing tags for NLB resources: %w", err)
+		}
+
+		for _, desc := range out.TagDescriptions {
+			actual := make(map[string]string, len(desc.Tags))
+			for _, t := range desc.Tags {
+				actual[aws.ToString(t.Key)] = aws.ToString(t.Value)
+			}
+
+			// Keep only the desired tags that are absent or have a stale value.
+			missing := map[string]string{}
+			for k, v := range desired {
+				if cur, ok := actual[k]; !ok || cur != v {
+					missing[k] = v
+				}
+			}
+			if len(missing) > 0 {
+				if err := c.addLoadBalancerTagsv2(ctx, aws.ToString(desc.ResourceArn), missing); err != nil {
+					return err
+				}
+			}
+
+			// Remove additional-tag keys the controller previously owned but no longer desires.
+			var previousManaged sets.Set[string]
+			if keys, ok := actual[TagNameKubernetesManagedTags]; ok {
+				previousManaged = parseManagedTagKeys(keys)
+			}
+			var toRemove []string
+			for key := range previousManaged {
+				if currentManaged.Has(key) {
+					continue
+				}
+				// Never remove a still-desired tag.
+				if _, ok := desired[key]; ok {
+					continue
+				}
+				toRemove = append(toRemove, key)
+			}
+			// Drop the marker itself once the user has cleared all additional tags, so no
+			// stale marker is left behind.
+			if currentManaged.Len() == 0 {
+				if _, ok := actual[TagNameKubernetesManagedTags]; ok {
+					toRemove = append(toRemove, TagNameKubernetesManagedTags)
+				}
+			}
+			if len(toRemove) > 0 {
+				if err := c.removeLoadBalancerTagsv2(ctx, aws.ToString(desc.ResourceArn), toRemove); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
 }
 
 // Retrieves instance's vpc id from metadata
--- a/pkg/providers/v1/aws_loadbalancer.go
+++ b/pkg/providers/v1/aws_loadbalancer.go
@@ -156,6 +156,15 @@ func (c *Cloud) ensureLoadBalancerv2(ctx context.Context, namespacedName types.N
 
 	// Get additional tags set by the user
 	tags := getKeyValuePropertiesFromAnnotation(annotations, ServiceAnnotationLoadBalancerAdditionalTags)
+
+	// Record which keys the controller owns from the annotation,
+	managedTagKeys := sets.List(sets.KeySet(tags))
+
+	// The marker is only set when the user supplies additional tags;
+	if len(managedTagKeys) > 0 {
+		tags[TagNameKubernetesManagedTags] = strings.Join(managedTagKeys, ",")
+	}
+
 	// Add default tags
 	tags[TagNameKubernetesService] = namespacedName.String()
 	tags = c.tagging.buildTags(ResourceLifecycleOwned, tags)
@@ -272,6 +281,10 @@ func (c *Cloud) ensureLoadBalancerv2(ctx context.Context, namespacedName types.N
 				nodePortTargetGroup[*targetGroup.Port] = &targetGroup
 			}
 
+			// resourceARNs collects the pre-existing resources that survive this reconcile, so
+			// their tags can be reconciled below.
+			resourceARNs := []string{aws.ToString(loadBalancer.LoadBalancerArn)}
+
 			// Handle additions/modifications
 			for _, mapping := range mappings {
 				frontendPort := mapping.FrontendPort
@@ -279,6 +292,8 @@ func (c *Cloud) ensureLoadBalancerv2(ctx context.Context, namespacedName types.N
 				nodePort := mapping.TrafficPort
 				// modifications
 				if listener, ok := actual[frontendPort][frontendProtocol]; ok {
+					// Surviving listener: reconcile its tags.
+					resourceARNs = append(resourceARNs, aws.ToString(listener.ListenerArn))
 					listenerNeedsModification := false
 
 					if listener.Protocol != mapping.FrontendProtocol {
@@ -365,6 +380,8 @@ func (c *Cloud) ensureLoadBalancerv2(ctx context.Context, namespacedName types.N
 							return nil, fmt.Errorf("error deleting old target group: %q", err)
 						}
 					} else {
+						// Surviving target group: reconcile its tags.
+						resourceARNs = append(resourceARNs, aws.ToString(targetGroup.TargetGroupArn))
 						// Run ensureTargetGroup to make sure instances in service are up-to-date
 						_, err = c.ensureTargetGroup(ctx,
 							targetGroup,
@@ -409,6 +426,11 @@ func (c *Cloud) ensureLoadBalancerv2(ctx context.Context, namespacedName types.N
 						dirty = true
 					}
 				}
+			}
+
+			// Reconcile tags on the LB and all surviving listeners and target groups collected.
+			if err := c.ensureLoadBalancerResourceTagsV2(ctx, resourceARNs, tags); err != nil {
+				return nil, err
 			}
 		}
 		if err := c.reconcileLBAttributes(ctx, aws.ToString(loadBalancer.LoadBalancerArn), annotations); err != nil {
--- a/pkg/providers/v1/aws_loadbalancer_test.go
+++ b/pkg/providers/v1/aws_loadbalancer_test.go
@@ -2503,3 +2503,303 @@ func TestCloud_ensureTargetGroupTargets(t *testing.T) {
 		})
 	}
 }
+
+// mockELBV2WithDelete extends MockedFakeELBV2 with a working DeleteListener.
+type mockELBV2WithDelete struct {
+	*MockedFakeELBV2
+}
+
+func (m *mockELBV2WithDelete) DeleteListener(ctx context.Context, input *elbv2.DeleteListenerInput, optFns ...func(*elbv2.Options)) (*elbv2.DeleteListenerOutput, error) {
+	kept := m.Listeners[:0]
+	for _, l := range m.Listeners {
+		if aws.ToString(l.ListenerArn) != aws.ToString(input.ListenerArn) {
+			kept = append(kept, l)
+		}
+	}
+	m.Listeners = kept
+	return &elbv2.DeleteListenerOutput{}, nil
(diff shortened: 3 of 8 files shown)


pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -392,6 +392,23 @@ func (c *Cloud) ensureLoadBalancerv2(ctx context.Context, namespacedName types.N
...
+					if _, ok := frontEndPorts[port][protocol]; ok {
+						err := c.addLoadBalancerTagsv2(ctx, aws.ToString(listener.ListenerArn), tags)
+						if err != nil {
+							return nil, err
+						}
+					}
+				}
+			}
+
+			if err := c.addLoadBalancerTagsv2(ctx, aws.ToString(loadBalancer.LoadBalancerArn), tags); err != nil {
```
mtulio: I think we'll need to implement some guardrails to safeguard user-provided tags through `ServiceAnnotationLoadBalancerAdditionalTags` replace controller-managed/private tags, such as `Name`, `kubernetes.io/cluster/<cluster-id>`, etc (which else?)  in [aws_validations](https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws_validations.go)?

mtulio (commented):
Thanks! Initial review, will check deepen later. It would be nice to have an [e2e case](https://github.com/kubernetes/cloud-provider-aws/blob/master/tests/e2e/loadbalancer.go#L100) for this flow.

pkg/providers/v1/aws_loadbalancer.go
alimx07: Good point, addressed.  I think controller-managed tags are `kubernetes.io/cluster/<cluster-id>`  , `KubernetesCluster` and `kubernetes.io/service-name`.  All three are guarded in the validation now, but I couldn't find a `Name` tag being set anywhere in the NLB path, could you point me to where you're seeing it or other worth guarded tags so I can take a look?

joshuakguo:
preface: trying to understand more about cloud providers, a question just for my own understanding

is the case for tag updates that we only care to propagate new tags on reconcile but not remove removed tags on reconcile?

alimx07:
> preface: trying to understand more about cloud providers, a question just for my own understanding
> 
> 
> 
> is the case for tag updates that we only care to propagate new tags on reconcile but not remove removed tags on reconcile?

Hi @joshuakguo, Actually we should, this is missing right. I also noticed this was not found in the previous version on EnsureLoadBalancer. May need clarification from a maintainer, but what comes to my mind now as the simplest way is doing describeTags then diff and call remove on the extra ones. Thanks for mentioning this.

kmala:
> > preface: trying to understand more about cloud providers, a question just for my own understanding
> > is the case for tag updates that we only care to propagate new tags on reconcile but not remove removed tags on reconcile?
> 
> Hi @joshuakguo, Actually we should, this is missing right. I also noticed this was not found in the previous version on EnsureLoadBalancer. May need clarification from a maintainer, but what comes to my mind now as the simplest way is doing describeTags then diff and call remove on the extra ones. Thanks for mentioning this.

I think the reason is because we don't have a way to know who owns the tags on the LB,  since these are values of annotations.

alimx07:
@kmala , Could we borrow from 3 way merge and introduce new annotation (e.g. something/last-applied-tags) as some sort of ownership we can diff on it?

alimx07:
Hey @kmala @mtulio, If you have time. Any updates here ? 
Thanks in advance !!

pkg/providers/v1/aws_loadbalancer.go
mtulio: thanks for restricting those.
you are correct, `Name` is not added, only LB name is set, which is different. I don't have a strong opinion about restricting `Name` since lookup may also use LB names. cc @kmala

pkg/providers/v1/aws.go
```diff
@@ -1219,6 +1219,27 @@ func (c *Cloud) describeLoadBalancerv2(ctx context.Context, name string) (*elbv2
...
+			Value: aws.String(v),
+		})
+	}
+
+	_, err := c.elbv2.AddTags(ctx, &elbv2.AddTagsInput{
+		ResourceArns: []string{resourceARN},
+		Tags:         tags,
+	})
+	if err != nil {
+		return fmt.Errorf("error adding tags: %v", err)
```
mtulio: Maybe increasing details since the caller are just raising the error?
```suggestion
		return fmt.Errorf("error adding tags NLB resource %s: %v", resourceARN, err)
```

tests/e2e/loadbalancer.go
```diff
@@ -998,3 +1033,43 @@ func gatherEventosOnFailure(ctx context.Context, cs clientset.Interface, namespa
...
+			for _, t := range desc.Tags {
+				actual[aws.ToString(t.Key)] = aws.ToString(t.Value)
+			}
+			for k, v := range expectedTags {
+				if actual[k] != v {
+					framework.Logf("tag %s on %s: want %q got %q", k, aws.ToString(desc.ResourceArn), v, actual[k])
+					return false, nil
+				}
+			}
+		}
```
mtulio, line 1577: can we increase logging here to help while debugging e2e test failures? "trying 1/N, etc"..

pkg/providers/v1/aws_loadbalancer_test.go
```diff
@@ -2503,3 +2503,150 @@ func TestCloud_ensureTargetGroupTargets(t *testing.T) {
...
+	for _, l := range m.Listeners {
+		if aws.ToString(l.ListenerArn) != aws.ToString(input.ListenerArn) {
+			kept = append(kept, l)
+		}
+	}
+	m.Listeners = kept
+	return &elbv2.DeleteListenerOutput{}, nil
+}
+
+func TestEnsureLoadBalancerv2_TaggingLifecycle(t *testing.T) {
```
mtulio: Thanks for setting up those test. Since you are defining a function to test EnsureLoadBalancerv2, I'd suggest:
- use Table-Driven test pattern like other units defined in this file for further maintenance :)
- making this more generically to allow other scenarios that may want to validate the annotations more generically through EnsureLoadBalancerv2 it can be expanded by adding cases, reusing all that pre-setup (mocks ec2, etc)
- Use `t.Context()` instead of `context.TODO()`

pkg/providers/v1/aws.go
```diff
@@ -1219,6 +1219,27 @@ func (c *Cloud) describeLoadBalancerv2(ctx context.Context, name string) (*elbv2
 	return nil, fmt.Errorf("NLB '%s' could not be found", name)
 }
 
+// addLoadBalancerTagsv2 tags a single ELBv2 resource (LB, listener, target group, etc.).
+// https://github.com/aws/aws-sdk-go-v2/blob/dc2d13fa6f1db25f1c6d804567e1ecfcdff4f040/service/elasticloadbalancingv2/api_op_AddTags.go#L14
+func (c *Cloud) addLoadBalancerTagsv2(ctx context.Context, resourceARN string, requested map[string]string) error {
```
mtulio, line 1229: we need to skip api calls when no tags `requested` to added

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -392,6 +392,23 @@ func (c *Cloud) ensureLoadBalancerv2(ctx context.Context, namespacedName types.N
 					}
 				}
 			}
+
+			// Tag all surviving listeners. Newly created listeners are already
+			// tagged at creation time via createListenerV2.
+			for port, protocols := range actual {
+				for protocol, listener := range protocols {
+					if _, ok := frontEndPorts[port][protocol]; ok {
+						err := c.addLoadBalancerTagsv2(ctx, aws.ToString(listener.ListenerArn), tags)
```
mtulio: is it possible to save API calls here from N to 1 AddTag, or less? by:
1. check if needs update tags and
2. send ARN list (LB + listeners), instead each ARN?

tests/e2e/loadbalancer.go
```diff
@@ -998,3 +1033,43 @@ func gatherEventosOnFailure(ctx context.Context, cs clientset.Interface, namespa
...
+	}
+	arns := make([]string, 0, len(out.Listeners))
+	for _, l := range out.Listeners {
+		arns = append(arns, aws.ToString(l.ListenerArn))
+	}
+	return arns, nil
+}
+
+func verifyResourceTags(ctx context.Context, elbClient *elbv2.Client, arns []string, expectedTags map[string]string) error {
+	return wait.PollUntilContextTimeout(ctx, 10*time.Second, 3*time.Minute, true, func(ctx context.Context) (bool, error) {
```
mtulio: it would be nice to document the reasoning of those timers.

pkg/providers/v1/aws_loadbalancer.go
alimx07: it seems that even `AddTags` take a slice , this error comes from the SDK when trying to update more than one.
```bash
An error occurred (ValidationError) when calling the AddTags operation: Only one resource can be tagged at a time
```
So this is limitation from AWS API for now.

But you are right , we could save  API calls by first call `DescribeTags` , then diff with tags and call `AddTags` per listener if needed. It will be either 1 or N+1 calls.

pkg/providers/v1/aws_loadbalancer_test.go
alimx07: Okay @mtulio , Makes sense on the `table-driven` pattern and `t.Context()`, will update both.

Just for clarification on the generic part, the goal is a table-driven `TestEnsureLoadBalancerv2_annotations*` where each case exercises a different annotation and asserts what it produces, So some tests like [this](https://github.com/kubernetes/cloud-provider-aws/blob/746099eda51cfab340f3d61ca3f10664e606c550/pkg/providers/v1/aws_loadbalancer_test.go#L1610), could be just cases included in our new generic one ?

mtulio: > on the generic part, the goal is a table-driven TestEnsureLoadBalancerv2_annotations* where each case exercises a different annotation and asserts what it produces, 

yeah, I think that way will pave a good path for annotations' tests maitainance and increase coverage. thanks

alimx07: @mtulio, I have done it. Would love if you any considerations about the test now?

pkg/providers/v1/aws.go
```diff
@@ -1219,6 +1220,74 @@ func (c *Cloud) describeLoadBalancerv2(ctx context.Context, name string) (*elbv2
...
+	})
+	if err != nil {
+		return fmt.Errorf("error adding tags NLB resource %s: %v", resourceARN, err)
+	}
+	return nil
+}
+
+// ensureLoadBalancerResourceTagsv2 makes sure every ELBv2 resource in arns carries the desired tags.
+// Tags are only added or updated.
+func (c *Cloud) ensureLoadBalancerResourceTagsv2(ctx context.Context, arns []string, desired map[string]string) error {
```
mtulio:
```suggestion
// ensureLoadBalancerResourceTagsV2 makes sure every ELBv2 resource in arns carries the desired tags.
// Tags are only added or updated.
func (c *Cloud) ensureLoadBalancerResourceTagsV2(ctx context.Context, arns []string, desired map[string]string) error {
```

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -392,6 +392,20 @@ func (c *Cloud) ensureLoadBalancerv2(ctx context.Context, namespacedName types.N
 					}
 				}
 			}
+
+			// Reconcile tags on the LB and all surviving listeners. Newly created
```
mtulio: I feel this block is not added in the better place it should be, or at least ensureLoadBalancerResourceTagsv2 is not.
Here is the end of the big block to "sync mappings" inside the existing LB synchronization.

Considering the problem below, I think this could be added the end of the ensureLoadBalancerV2 function using the "reconciliation pattern".

Question 1) tag removal support
Question/issue I am seeing here: how that annotation controls removed tags? given the following scenrio:
t1) user update a svc with tags: env=prod,team=eng
t2) user remove a tag `team=eng`: will `team` tag leaked ?
t3) user remove the tag annotation: will `env` tag leaked?

How we could prevent that "leak"?

Question 2) Tagging TGs
do we need tag the target groups as may be part of the "load balancer resources"?

pkg/providers/v1/aws.go
```diff
@@ -1219,6 +1220,74 @@ func (c *Cloud) describeLoadBalancerv2(ctx context.Context, name string) (*elbv2
...
+			Value: aws.String(v),
+		})
+	}
+
+	_, err := c.elbv2.AddTags(ctx, &elbv2.AddTagsInput{
+		ResourceArns: []string{resourceARN},
+		Tags:         tags,
+	})
+	if err != nil {
+		return fmt.Errorf("error adding tags NLB resource %s: %v", resourceARN, err)
```
mtulio:
```suggestion
		return fmt.Errorf("error adding tags NLB resource %s: %w", resourceARN, err)
```

tests/e2e/loadbalancer.go
```diff
@@ -128,6 +129,40 @@ var _ = Describe("[cloud-provider-aws-e2e] loadbalancer", func() {
...
+
+				framework.Logf("updating annotation to e2e-tag=two")
+				updated, err := cfg.LBJig.UpdateService(cfg.ctx, func(s *v1.Service) {
+					s.Annotations[annotationLBAdditionalTags] = "e2e-tag=two"
+				})
+				framework.ExpectNoError(err)
+				cfg.svc = updated
+
+				framework.Logf("verifying updated tag e2e-tag=two on LB and listeners")
+				framework.ExpectNoError(verifyResourceTags(cfg.ctx, elbClient, arns, map[string]string{"e2e-tag": "two"}))
```
mtulio: I believe you need to refresh `listenerARNs` as you are updating the svc, if the controller eventually recreate the listener you'll be checking old listener (if not delted) which will have old tag values

tests/e2e/loadbalancer.go
```diff
@@ -998,3 +1033,56 @@ func gatherEventosOnFailure(ctx context.Context, cs clientset.Interface, namespa
 	gatherControllerLogs(ctx, cs, namespace, resourceName)
 	gatherServiceStatus(ctx, cs, namespace, resourceName)
 }
+
+func getLBListenerARNs(ctx context.Context, elbClient *elbv2.Client, lbARN string) ([]string, error) {
+	out, err := elbClient.DescribeListeners(ctx, &elbv2.DescribeListenersInput{
+		LoadBalancerArn: aws.String(lbARN),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to describe listeners for LB %s: %v", lbARN, err)
```
mtulio, line 1515:
```suggestion
		return nil, fmt.Errorf("failed to describe listeners for LB %s: %w", lbARN, err)
```

pkg/providers/v1/aws.go
```diff
@@ -1219,6 +1220,74 @@ func (c *Cloud) describeLoadBalancerv2(ctx context.Context, name string) (*elbv2
...
+		end := start + describeTagsMaxARNs
+		if end > len(arns) {
+			end = len(arns)
+		}
+
+		out, err := c.elbv2.DescribeTags(ctx, &elbv2.DescribeTagsInput{
+			ResourceArns: arns[start:end],
+		})
+		if err != nil {
+			return fmt.Errorf("error describing tags for NLB resources: %v", err)
```
mtulio:
```suggestion
			return fmt.Errorf("error describing tags for NLB resources: %w", err)
```

pkg/providers/v1/aws_loadbalancer.go
alimx07: For Tag removal support. the current approach will not handle that and we could say tags will `leak` , which is the same approach for the old [ensureLoadBalancer](https://github.com/kubernetes/cloud-provider-aws/blob/c34d66ed717aaffe3319f149c26e28163e206a3e/pkg/providers/v1/aws_loadbalancer.go#L1389-L1399).  The reason is we do not have a way to know who owns tags on LB (e.g. user , controller, terraform) , so naive `describe + remove` will not work. 

we could borrow from 3 way merge and introduce new controller managed annotation (e.g. `something/last-applied-tags`) as some sort of ownership we can diff on it, if we are okay with that, I could go on with it.

For TGs, tags are obtained through the reconcile loop, so you're right. I missed that and was focused only on Listeners. With the current approach, this should be a minimal amount of work. Nice pushback :)

tests/e2e/loadbalancer.go
alimx07: okay I will do it. 

Just a question for curiosity, in the reconcile loop, listeners only `deleted/created` as ports mapping change. while this test just update some tags, so why recreation could happen here?

pkg/providers/v1/aws_loadbalancer.go
mtulio: > we could borrow from 3 way merge and introduce new controller managed annotation (e.g. something/last-applied-tags) as some sort of ownership we can diff on it, if we are okay with that, I could go on with it.

yep, that's a good idea. Beforehand I would check what is the AWS LBC behavior of existing BYO tags, so we can prevent diverging from there and here in the UX. (I quickly read the [documentation](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/annotations/#resource-tags) and did't get much insights)

Having said that, exploring that idea of managing the BYO tags, it indeed be interesting to have controller managed tag managing which tag is added (through values containing tagKeys?)

For example:

when annotation is **added**:  `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: team=squadA,env=stage,group=tmp`
The resource will have tags - managed by controller:
- kubernetes.io/cloud-provider/managed-tags: team,env,group
- team: squadA
- env: stage
- group: tmp

when annotation is **updated** to  `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: team=squadA,env=prod` (removing `group`, new value for `env`)

the tag reconciliation will detect drift from annotation vs managed tag, and remove stale tag `group`, as well update value of `env`, resulting in resource tags:
- kubernetes.io/cloud-provider/managed-tags: team,env
- team: squadA
- env: prod

when annotation is **removed** from service, the tag reconciliation must detect managed tag exists in the resource && annotation not present, resulting in removing any tag added by controller, as well the managed tag

alimx07: Hi @mtulio, sorry for the late reply.

For AWS LBC, it assumes itself as the source of truth for tags. On each reconcile it removes any tag not in its desired set (annotation tags + `--default-tags` + its own tracking tags), except AWS system tags `(aws:*)` and anything listed in `--external-managed-tags` ([code](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/bf939f8bbbfc5888fb98e6ffe0674ed606915052/pkg/deploy/elbv2/tagging_manager.go#L138-L194), [docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/deploy/configurations/#:~:text=external%2Dmanaged%2Dtags,ignored%20during%20reconciliation)). 


Maybe I am wrong but I went with `managed annotation` for now to a tag propagation feature, it could be a weird UX, but matching LBC here making `CCM` authoritative seems like a breaking change I think.

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -156,6 +156,15 @@ func (c *Cloud) ensureLoadBalancerv2(ctx context.Context, namespacedName types.N
...
 	// Get additional tags set by the user
 	tags := getKeyValuePropertiesFromAnnotation(annotations, ServiceAnnotationLoadBalancerAdditionalTags)
+
+	// Record which keys the controller owns from the annotation,
+	managedTagKeys := sets.List(sets.KeySet(tags))
+
+	// The marker is only set when the user supplies additional tags;
+	if len(managedTagKeys) > 0 {
+		tags[TagNameKubernetesManagedTags] = strings.Join(managedTagKeys, ",")
+	}
```
mfbonfigli, line 166: I see this error in the e2e tests:

```
  I0625 04:37:45.865090   33395 loadbalancer.go:1289] Resource events for "lbconfig-test-nlb-tags":
  I0625 04:37:45.865122   33395 loadbalancer.go:1291]   [Normal] EnsuringLoadBalancer/lbconfig-test-nlb-tags: Ensuring load balancer - 2026-06-25 04:33:57 +0000 UTC
  I0625 04:37:45.865132   33395 loadbalancer.go:1291]   [Warning] SyncLoadBalancerFailed/lbconfig-test-nlb-tags: Error syncing load balancer: failed to ensure load balancer: error creating load balancer: "operation error Elastic Load Balancing v2: CreateLoadBalancer, https response error StatusCode: 400, RequestID: 29d66d90-fd4f-450d-91be-009f914df513, api error ValidationError: 1 validation error detected: Value 'drop-tag,e2e-tag' at 'tags.6.member.value' failed to satisfy constraint: Member must satisfy regular expression pattern: ^([\\p{L}\\p{Z}\\p{N}_.:/=+\\-@]*)$" - 2026-06-25 04:33:57 +0000 UTC
```

Reference: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/cloud-provider-aws/1447/pull-cloud-provider-aws-e2e/2069981875467718656/build-log.txt

Doesn't seem the comma is an accepted value, as also documented here:
https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-tags.html

Could you double check?

alimx07: Thanks for investigating! 

I think I hit a blocker , The separator uniqueness part maybe solvable with encoding/decoding, so that one
isn't really the problem. The actual blocker is the [255-char value cap](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-tags.html#:~:text=Maximum%20value%20length%E2%80%94255%20Unicode%20characters), that makes the `managed-tags` marker unreliable in the first place, since the joined key list can easily go past it (I should double check that before). 

So,  either drop the removal mechanism and just keep update/add as the initial intend or we can store the `managed-tags` state as an annotation on the **Service** itself `(3 way merge style)`. I'm not sure if the second is acceptable design-wise

Any ideas @mtulio?

mtulio: @alimx07 thanks for raising that. It is a good question.

Hey @kmala  @elmiko can you see in other cloud-provider implementations that practice? I think the idea through annotation would be safer and prevent "leaking tags", but would love to hear from your thoughts too before any implementation.

PR #1452, opened 2026-06-20
Fix duplicate NLB target group attribute validation

immanuwell:
What type of PR is this?

/kind bug
/kind failing-test

What this PR does / why we need it:

Reject duplicate entries in `service.beta.kubernetes.io/aws-load-balancer-target-group-attributes`.
Before this, `preserve_client_ip.enabled=true,preserve_client_ip.enabled=false` got collapsed by the generic parser and the last value won, so the duplicate check never fired. Pretty sneaky bug, ngl.

Which issue(s) this PR fixes:

Related to #1214

Special notes for your reviewer:

Repro:
1. Create an NLB Service with `service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: preserve_client_ip.enabled=true,preserve_client_ip.enabled=false`
2. Before this patch, validation accepts it and silently keeps `false`
3. With this patch, validation returns `"preserve_client_ip.enabled" is set twice in the annotation`

Does this PR introduce a user-facing change?:

```release-note
Fixed validation for `service.beta.kubernetes.io/aws-load-balancer-target-group-attributes` so duplicate attributes are rejected instead of silently using the last value.
```

### DIFF
--- a/pkg/providers/v1/aws_validations.go
+++ b/pkg/providers/v1/aws_validations.go
@@ -104,17 +104,12 @@ func validateServiceAnnotations(v *awsValidationInput) error {
 func validateServiceAnnotationTargetGroupAttributes(v *awsValidationInput) error {
 	errPrefix := "error validating target group attributes"
 
-	// Attributes are in format key=value separated by comma.
-	annotationGroupAttributes := getKeyValuePropertiesFromAnnotation(v.annotations, ServiceAnnotationLoadBalancerTargetGroupAttributes)
-	targetGroupAttributes := make(map[string]string, len(annotationGroupAttributes))
+	targetGroupAttributes, err := getTargetGroupAttributesFromAnnotation(v.annotations)
+	if err != nil {
+		return fmt.Errorf("%s: %w", errPrefix, err)
+	}
 
-	for attrKey, attrValue := range annotationGroupAttributes {
-		if _, ok := targetGroupAttributes[attrKey]; ok {
-			return fmt.Errorf("%s: %q is set twice in the annotation", errPrefix, attrKey)
-		}
-		if len(attrValue) == 0 {
-			return fmt.Errorf("%s: attribute value is empty for %q", errPrefix, attrKey)
-		}
+	for attrKey, attrValue := range targetGroupAttributes {
 
 		switch attrKey {
 		case targetGroupAttributePreserveClientIPEnabled:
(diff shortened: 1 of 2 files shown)


pkg/providers/v1/aws_validations_test.go
```diff
@@ -100,14 +100,14 @@ func TestValidateServiceAnnotationTargetGroupAttributes(t *testing.T) {
 			expectedError: "",
 		},
 		{
-			name: "duplicate attribute in annotation (last one wins - no error expected)",
```
kmala, line 103: This suggests that this is expected behavior correct that to overwrite instead of erroring? If we error it would be a change in behavior right?

kmala: cc: @mtulio

mtulio: @kmala I believe erroring will be a better approach to validate, so we can keep it strong preventing duplication instead of accepting the last one. As well preventing sending duplicated attributes to [api](https://github.com/kubernetes/cloud-provider-aws/blob/0e30beb4b1531f818872109d79b8f703b587c7ac/pkg/providers/v1/aws_loadbalancer.go#L600).

pkg/providers/v1/aws_validations.go
```diff
@@ -104,11 +104,25 @@ func validateServiceAnnotations(v *awsValidationInput) error {
 func validateServiceAnnotationTargetGroupAttributes(v *awsValidationInput) error {
 	errPrefix := "error validating target group attributes"
 
-	// Attributes are in format key=value separated by comma.
-	annotationGroupAttributes := getKeyValuePropertiesFromAnnotation(v.annotations, ServiceAnnotationLoadBalancerTargetGroupAttributes)
```
mtulio, line 108: looks like `getKeyValuePropertiesFromAnnotation()` is strictly created to validate tags, which accepts empty values, as that suggestion is not using that function anymore, will be good to implement more restrictive key/value extraction from annotation function with constraints (prevent empty, duplicated, etc) implemented in this  validation `validateServiceAnnotationTargetGroupAttributes()` so that we can test and re-use it accordingly?

pkg/providers/v1/aws_validations_test.go
kmala: makes sense....want to check with you since you made the change originally and if there was any specific reason

pkg/providers/v1/aws_validations.go
immanuwell: fixed in the latest push

pkg/providers/v1/aws_validations_test.go
immanuwell: fixed in the latest push here

immanuwell:
@kmala @mtulio addressed your suggestions, PTAL

pkg/providers/v1/aws_validations.go
```diff
@@ -142,3 +137,39 @@ func validateServiceAnnotationTargetGroupAttributes(v *awsValidationInput) error
...
+	for _, rawAttribute := range strings.Split(strings.TrimSpace(annotationValue), ",") {
+		rawAttribute = strings.TrimSpace(rawAttribute)
+		if rawAttribute == "" {
+			continue
+		}
+
+		attrParts := strings.SplitN(rawAttribute, "=", 2)
+		attrKey := attrParts[0]
+		attrValue := ""
+		if len(attrParts) == 2 {
```
kmala, line 160: this will always be only 2 right since we are using SplitN with 2? this doesn't handle invalid key value pair right? can we reuse the similar code as getKeyValuePropertiesFromAnnotation for splitting?

PR #1473, opened 2026-07-23
Narrow CLB BYO-SG cleanup revoke to the matching UserIdGroupPair

### LINKED ISSUE #1472: CLB BYO-SG cleanup revokes the entire consolidated IpPermission on the node SG, dropping the node-SG self-reference and breaking pod traffic
**What happened**:

When a `Service type=LoadBalancer` backed by a **Classic Load Balancer (CLB)** is transitioned from the controller-managed security group to a user-provided (BYO) security group (annotation `service.beta.kubernetes.io/aws-load-balancer-security-groups`), the managed-SG cleanup revokes ingress rules on the cluster/node security group at the **whole-`IpPermission`** granularity instead of removing only the single `UserIdGroupPair` that references the SG being deleted.

EC2 consolidates ingress rules that share the same protocol/port tuple into one `IpPermission` with a list of `UserIdGroupPairs`. The node SG's all-traffic (`-1`) self-reference (which allows intra-cluster node/pod traffic) is typically stored in the **same** `IpPermission` as the controller's all-traffic LB→node rule. Revoking that whole permission therefore also removes the self-reference, breaking cross-node pod-to-pod connectivity.

Root cause, in [pkg/providers/v1/aws_loadbalancer.go](https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws_loadbalancer.go#L1934):

1. `buildSecurityGroupRuleReferences` inserts the **whole** `IpPermission` when any one of its pairs matches the target SG id — not just the matching pair:
   ```go
   // ~L1934-1936
   for _, pair := range rule.UserIdGroupPairs {
       if pair.GroupId != nil && aws.ToString(pair.GroupId) == sgID {
           groupsLinkedPermissions[&sg].Insert(rule) // inserts the entire IpPermission (all pairs)
       }
   }
   ```
2. `removeOwnedSecurityGroups` passes that set straight to `RevokeSecurityGroupIngress` with no narrowing:
   ```go
   // ~L1970-1973
   if _, err := c.ec2.RevokeSecurityGroupIngress(ctx, &ec2.RevokeSecurityGroupIngressInput{
       GroupId:       sgTarget.GroupId,
       IpPermissions: sgPerms.List(), // whole consolidated permission(s)
   }); ...
   ```
   The existing `IPPermissionSet.Ungroup()` helper (`pkg/providers/v1/sets_ippermissions.go`) is not called on this path, and the revoke is gated only by the *target* SG's cluster tag — not per rule.

This was introduced by #1209 (merged to `master` 2026-05-04), first released in **v1.36.0**. The `removeOwnedSecurityGroups` function does not exist in release-1.30 … release-1.35, so those lines are not affected.

Cleanup should revoke only the `UserIdGroupPair` that references the SG being deleted, leaving all other pairs in the consolidated permission — notably the node-SG self-reference — intact. Cross-node pod traffic should be unaffected by a CLB managed→BYO SG transition.

1. On a v1.36 cluster whose node/cluster SG has the common all-traffic (`-1`) self-reference, create a `Service type=LoadBalancer` (CLB) in default managed-SG mode. The controller creates a managed LB SG and adds an all-traffic (`-1`) rule to the node SG allowing traffic from that managed SG; EC2 consolidates it with the existing `-1` self-reference into one `IpPermission`.
2. Add `service.beta.kubernetes.io/aws-load-balancer-security-groups: <byo-sg-id>` to the existing Service.
3. The controller applies the BYO SG and cleans up the orphaned managed SG. A single `RevokeSecurityGroupIngress` is issued whose payload contains **both** the managed-SG pair and the node-SG self-reference pair (verifiable in CloudTrail).
4. Cross-node pod-to-pod traffic breaks. Re-adding the node-SG `-1` self-reference restores it.

The single-`UserIdGroupPair` revoke pattern already exists elsewhere in the provider and can be reused:
- The CLB **delete** path (`updateInstanceSecurityGroupsForLoadBalancer`, `aws.go ~L3071-3075`) constructs a permission carrying only the single LB-SG pair before revoking.
- The **NLB** path ungroups and filters the revoke set by rule description (`kubernetes.io/rule/nlb/*`) so it only removes controller-owned rules.

Minimal fix — in `buildSecurityGroupRuleReferences`, insert a copy of the rule narrowed to the matching pair (and cleared of `IpRanges`/`Ipv6Ranges`/`PrefixListIds`) instead of the whole rule:
```go
if pair.GroupId != nil && aws.ToString(pair.GroupId) == sgID {
    narrowed := rule
    narrowed.UserIdGroupPairs = []ec2types.UserIdGroupPair{pair}
    narrowed.IpRanges = nil
    narrowed.Ipv6Ranges = nil
    narrowed.PrefixListIds = nil
    groupsLinkedPermissions[&sg].Insert(narrowed)
}
```

**Environment**:
- Kubernetes version (use `kubectl version`): 1.36 (server `v1.36.x`); cloud-provider-aws v1.36.0 / v1.36.1
- Cloud provider or hardware configuration: AWS; in-tree AWS cloud-controller-manager; `Service type=LoadBalancer` with CLB
- OS (e.g. from /etc/os-release): N/A (control-plane controller behavior)
- Kernel (e.g. `uname -a`): N/A
- Install tools: N/A
- Others: Triggered by the CLB managed-SG → BYO-SG (`service.beta.kubernetes.io/aws-load-balancer-security-groups`) transition on an existing Service.

/kind bug

snarkychef:
**What type of PR is this?**

/kind bug

`buildSecurityGroupRuleReferences` inserted the whole EC2 `IpPermission` when any one of its `UserIdGroupPairs` referenced the security group being removed. `removeOwnedSecurityGroups` then passed that consolidated permission straight to `RevokeSecurityGroupIngress`, so a single revoke removed every pair sharing the same protocol/port tuple.

EC2 commonly consolidates the node/cluster SG's all-traffic (`-1`) self-reference and the controller's LB→node rule into one `IpPermission`, so cleaning up the managed SG on a CLB managed→BYO SG transition (via the `service.beta.kubernetes.io/aws-load-balancer-security-groups` annotation) also revoked the self-reference and broke cross-node pod traffic.

This PR narrows the linked permission to only the matching `UserIdGroupPair` (and clears the CIDR / IPv6 / prefix-list dimensions), matching the single-pair revoke granularity already used by the CLB delete and NLB paths, so unrelated pairs such as the node SG self-reference are preserved.

Fixes #1472

- The bug is scoped to the CLB managed→BYO-SG cleanup path added in #1209 (first released in v1.36.0); earlier release branches don't have `removeOwnedSecurityGroups`.
- The fix narrows at the source (`buildSecurityGroupRuleReferences`) rather than calling the existing `Ungroup()` helper, because `Ungroup()` would still need per-`sgID` selection and it currently leaves non-split dimensions populated (shallow copy); clearing `IpRanges`/`Ipv6Ranges`/`PrefixListIds` here avoids that class of over-revoke.
- Added a regression case to `TestCloud_buildSecurityGroupRuleReferences`: a consolidated `-1` permission bundling the node-SG self-reference with the target SG must yield a revoke set containing only the target pair. It fails on the pre-fix code and passes after.

```release-note
Fixed a bug where transitioning a Service-managed Classic Load Balancer to a user-provided security group could revoke unrelated ingress rules (including the node security group's self-reference) on security groups that share a consolidated IP permission, breaking cross-node pod traffic.
```

### DIFF
--- a/pkg/providers/v1/aws_loadbalancer.go
+++ b/pkg/providers/v1/aws_loadbalancer.go
@@ -1933,7 +1933,16 @@ func (c *Cloud) buildSecurityGroupRuleReferences(ctx context.Context, sgID strin
 			if rule.UserIdGroupPairs != nil {
 				for _, pair := range rule.UserIdGroupPairs {
 					if pair.GroupId != nil && aws.ToString(pair.GroupId) == sgID {
-						groupsLinkedPermissions[&sg].Insert(rule)
+						// Revoke only the UserIdGroupPair that references sgID. EC2 consolidates rules
+						// with the same protocol/port into one IpPermission with multiple pairs, so the
+						// matched rule may also carry unrelated pairs (e.g. the node SG self-reference).
+						// Build a new permission with just the matching pair so nothing else is revoked.
+						groupsLinkedPermissions[&sg].Insert(ec2types.IpPermission{
+							IpProtocol:       rule.IpProtocol,
+							FromPort:         rule.FromPort,
+							ToPort:           rule.ToPort,
+							UserIdGroupPairs: []ec2types.UserIdGroupPair{pair},
+						})
 					}
 				}
 			}
--- a/pkg/providers/v1/aws_loadbalancer_test.go
+++ b/pkg/providers/v1/aws_loadbalancer_test.go
@@ -2220,6 +2220,64 @@ func TestCloud_buildSecurityGroupRuleReferences(t *testing.T) {
 			},
 		},
 		{
+			name:          "consolidated permission is narrowed to the matching pair only",
+			targetGroupID: "sg-target",
+			setupMocks: func(mockedEC2 *MockedFakeEC2) {
+				mockedEC2.On("DescribeSecurityGroups", &ec2.DescribeSecurityGroupsInput{
+					Filters: []ec2types.Filter{
+						{
+							Name:   aws.String("ip-permission.group-id"),
+							Values: []string{"sg-target"},
+						},
+					},
+				}).Return([]ec2types.SecurityGroup{
+					{
+						GroupId: aws.String("sg-node"),
+						Tags: []ec2types.Tag{
+							{
+								Key:   aws.String("kubernetes.io/cluster/test-cluster"),
+								Value: aws.String("owned"),
+							},
+						},
+						// A single all-traffic IpPermission that EC2 has consolidated: it
+						// contains both the node SG's own self-reference and the reference to
+						// sg-target. Only the sg-target pair should be selected for revocation;
+						// the self-reference must be preserved.
+						IpPermissions: []ec2types.IpPermission{
+							{
+								IpProtocol: aws.String("-1"),
+								UserIdGroupPairs: []ec2types.UserIdGroupPair{
+									{GroupId: aws.String("sg-node")},
+									{GroupId: aws.String("sg-target")},
+								},
+							},
+						},
+					},
+				}, nil)
+			},
+			expectError:                          false,
+			expectedGroupsWithTagsCount:          1,
+			expectedGroupsLinkedPermissionsCount: 1,
+			additionalAssertions: func(t *testing.T, groupsWithTags map[*ec2types.SecurityGroup]bool, groupsLinkedPermissions map[*ec2types.SecurityGroup]IPPermissionSet) {
+				var foundSG *ec2types.SecurityGroup
+				for sg := range groupsLinkedPermissions {
+					if aws.ToString(sg.GroupId) == "sg-node" {
+						foundSG = sg
+						break
+					}
+				}
+				require.NotNil(t, foundSG)
+
+				linked := groupsLinkedPermissions[foundSG]
+				require.Equal(t, 1, linked.Len())
+				perm := linked.List()[0]
+				// The revoke set must reference ONLY sg-target, never the node SG's
+				// self-reference (which would break intra-cluster traffic if revoked).
+				require.Len(t, perm.UserIdGroupPairs, 1)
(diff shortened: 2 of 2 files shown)


kmala:
@mtulio can you take a look?

pkg/providers/v1/aws_loadbalancer.go
```diff
@@ -1933,7 +1933,19 @@ func (c *Cloud) buildSecurityGroupRuleReferences(ctx context.Context, sgID strin
...
 					if pair.GroupId != nil && aws.ToString(pair.GroupId) == sgID {
-						groupsLinkedPermissions[&sg].Insert(rule)
+						// Only revoke the single UserIdGroupPair that references sgID, not the
+						// whole IpPermission. EC2 consolidates rules sharing the same protocol/port
+						// tuple into one IpPermission with multiple UserIdGroupPairs, so a rule that
+						// references sgID may also carry unrelated pairs (e.g. the node SG's own
+						// self-reference that allows intra-cluster traffic). Inserting the whole
+						// rule here would revoke all of them in a single RevokeSecurityGroupIngress
+						// call. Narrow to the matching pair and drop the CIDR/prefix-list dimensions.
+						linkedRule := rule
```
yue9944882: can we construct a new IpPermission struct instead of shallow-copy and reset unwanted fields? say if there are more properties introduced under this struct in the future, we might unexpectedly revoke that. is that correct?

snarkychef: Good point. I made the change now.

I considered both options and didn't have a preference initially as new struct creation requires some field copying too. But I agree that revoking more than we should is a bigger problem than failing to revoke.

yue9944882 (commented):
/lgtm
/approve

Issue #1461, opened 2026-07-02
Compilation break on k8s 1.37 dependency bump: NodeMonitorPeriod moved from KubeCloudShared to NodeLifecycleController


kubeboiii:
## Summary

When `cloud-provider-aws` upgrades its `k8s.io/*` dependencies from v0.36.0 to v0.37.0 (expected around k8s 1.37 RC, ~Aug 2026), compilation will fail due to a field relocation in the upstream `k8s.io/cloud-provider` library.

## Root cause

Kubernetes PR https://github.com/kubernetes/kubernetes/pull/137964 ("Introduce parallelization for MonitorNodes loop") moved the `NodeMonitorPeriod` config field from `KubeCloudSharedConfiguration` to a new `NodeLifecycleControllerConfiguration` struct. This is a source-incompatible change (compile-time break), but behavior-preserving — the `--node-monitor-period` flag and its 5s default remain unchanged.

## Impact in cloud-provider-aws

One line will fail to compile:

**File:** `pkg/controllers/tagging/tagging_controller_wrapper.go:46`

```go
// Current (breaks against v0.37.0):
completedConfig.ComponentConfig.KubeCloudShared.NodeMonitorPeriod.Duration,

// Fix required:
completedConfig.ComponentConfig.NodeLifecycleController.NodeMonitorPeriod.Duration,
```

No other files reference the moved field.

## Fix

One-line change to update the field path. The tagging controller will continue to receive the same `NodeMonitorPeriod` value; no behavioral change.

## Timing

Per coordination with @keerthan in #sig-cloud-provider:

- The fix must land **with or immediately after** the go.mod bump to v0.37.0
- Expected timing: ~Aug 5, 2026 when v0.37.0-rc.0 is published (maintainers may also bump on alpha/beta)
- I have a prepared branch and will open a PR once the dependency bump is coordinated

**Prepared branch:** https://github.com/kubeboiii/cloud-provider-aws/tree/fix/node-monitor-period-1.37

## References

- Upstream breaking change: https://github.com/kubernetes/kubernetes/pull/137964
- SIG cloud-provider meeting discussion (2026-07-02)
- Affected cloud providers: AWS, GCP, Azure (per @cheftako's analysis)

PR #1474, opened 2026-07-23
fix: invalid/deleted instance handling hardening

### LINKED ISSUE #1471: Node object not deleted if providerID not set
**What happened**:

We observed a NotReady node object not being cleaned up after its backing EC2 instance had been deleted on a 1.34 EKS cluster. The cloud-controller-manager was continually re-queuing the node object with the following error:
```
node_lifecycle_controller.go:180] error checking if node <private-dns> exists: instance not found
```

We would expect that the node object would be terminated by the cloud-controller-manager after confirming that the backing EC2 instance had been deleted by calling DescribeInstance. It looks like the controller will only gracefully handle terminating the node object after the EC2 instance is deleted if the node object has the providerID set. In this case, the controller will lookup the corresponding instanceID from the providerID field in the node object to call DescribeInstance before removing the node object if the instance has been deleted. However, if the EC2 instance is deleted prior to the providerID being set, the controller will continually fail trying to lookup the instance using the privateDNS node name rather than identify that the backing EC2 instance has been terminated.

Create a node object without a providerID pointing to a non-existent node name:
```
kubectl apply -f - <<EOF
apiVersion: v1
kind: Node
metadata:
  name: ip-7-248-227-253.us-west-2.compute.internal
spec:
  providerID: ""
status:
  conditions:
  - type: Ready
    status: "False"
    lastHeartbeatTime: "2026-01-01T00:00:00Z"
    lastTransitionTime: "2026-01-01T00:00:00Z"
EOF
```
This will induce the same "instance not found" error. Compared to creating the node object with the providerID set from a non-existent node:
```
kubectl apply -f - <<EOF
apiVersion: v1
kind: Node
metadata:
  name: ip-7-248-227-256.us-west-2.compute.internal
spec:
  providerID: "aws:///us-west-2a/i-015e7cb9c3ac6a284"
status:
  conditions:
  - type: Ready
    status: "False"
    lastHeartbeatTime: "2020-01-01T00:00:00Z"
    lastTransitionTime: "2020-01-01T00:00:00Z"
EOF
```
In this case, the node object will be deleted as expected.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.34.9-eks-8f14419
- OS (e.g. from /etc/os-release): NA
- Kernel (e.g. `uname -a`): NA
- Install tools: NA
- Others: NA

/kind bug

joshuakguo:
**What type of PR is this?**
/kind bug

adds logic to handle notready nodes with no provider id properly to not requeue indefinitely
- also mirrors the existing logic that is in instances v1

while looking at that issue also found a gap where if we get back from aws that the instance is not found for an id that we described, we should properly delete the node

also added regression tests for the original issue outlined in the issue, the secondary issue that was found, and a test for kwok

Fixes #1471

```release-note
NONE
```

### DIFF
--- a/pkg/providers/v1/aws.go
+++ b/pkg/providers/v1/aws.go
@@ -3455,6 +3455,11 @@ func (c *Cloud) UpdateLoadBalancer(ctx context.Context, clusterName string, serv
 func (c *Cloud) getInstanceByID(ctx context.Context, instanceID string) (*ec2types.Instance, error) {
 	instances, err := c.getInstancesByIDs(ctx, []string{instanceID})
 	if err != nil {
+		// A DescribeInstances by ID for an instance that no longer exists returns
+		// an InvalidInstanceID.NotFound error rather than an empty result.
+		if IsAWSErrorInstanceNotFound(err) {
+			return nil, cloudprovider.InstanceNotFound
+		}
 		return nil, err
 	}
 
--- a/pkg/providers/v1/instances_v2.go
+++ b/pkg/providers/v1/instances_v2.go
@@ -22,6 +22,7 @@ package aws
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"k8s.io/cloud-provider-aws/pkg/providers/v1/variant"
 	"strconv"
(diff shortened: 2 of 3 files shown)


joshuakguo:
kubetest e2e's fail because of 1.37 changes it looks like?

kmala:
yes, we need to change the flags with the 1.37 https://github.com/kubernetes/cloud-provider-aws/issues/1461
@kubeboiii are you going to bump the dependencies and fix as beta is released for 1.37?

joshuakguo:
should we just wait for 1.37.0-rc.0 since it is supposed to be released next week instead of bumping beta version?

kmala:
> should we just wait for 1.37.0-rc.0 since it is supposed to be released next week instead of bumping beta version?

i think its okay to update to rc now and later update to beta .

joshuakguo:
I've gone ahead and raised the PR to bump the dependencies #1476

PR #1479, opened 2026-08-12
Add terminal AWS API error metric (count per logical call, not per attempt)

snarkychef:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it:**

`cloudprovider_aws_api_response_status_total` is recorded by a Deserialize
middleware, which runs *inside* the AWS SDK's retry loop and reads the HTTP
status from `out.RawResponse`. As a result it increments **once per attempt**: an
API call that receives a throttling/5xx response and then succeeds on a retry
still increments the counter (once per failed attempt).

Because the SDK transparently retries and recovers from most throttling and
transient 5xx responses, this makes the metric a poor signal for the actual AWS
API error rate — it is dominated by transient responses that never surfaced as a
real failure to the caller. On busy clusters (frequent EC2 calls during node
initialization, node-lifecycle polling, and load-balancer target syncing) the
inflation is large.

This PR moves the metrics middleware from a Deserialize step to a **Finalize step
inserted before the SDK's `Retry` middleware**, so it wraps the entire retry loop
and runs **once per logical call**, recording the status only of calls that fail
**terminally** (after retries are exhausted). The terminal status is read from the
terminal error via `*awshttp.ResponseError`. A response that is retried and then
succeeds is no longer counted.

The metric name and its `{service, operation, status_code}` labels are unchanged;
only the counting semantics (per-attempt → per-terminal-outcome) and the help
text change.

**Which issue(s) this PR fixes:**

**Special notes for your reviewer:**

- Metric name and labels are unchanged; this is a semantics fix to an ALPHA metric.
- The middleware is now attached via a shared `addAWSAPIMetricsMiddleware` option
  at the two existing call sites (the default SDK clients in `aws_sdk.go` and the
  STS client in `aws.go`).
- Tests in `aws_api_metrics_test.go` were updated for the Finalize-based
  middleware and extended with a retry-loop test that drives the real SDK retryer:
  it asserts that a transient 503 which succeeds on retry is **not** counted, a
  retry-exhausted 503 is counted **exactly once**, and a non-retryable 400 is
  counted once.

```release-note
Fixed `cloudprovider_aws_api_response_status_total` to count AWS API errors once per logical call (after the SDK's retries are exhausted) instead of once per attempt. Previously a response that was retried and then succeeded was still counted, which over-counted transient errors; the metric now reflects terminal failures only. The metric name and labels are unchanged.
```

### DIFF
--- a/pkg/providers/v1/aws_metrics.go
+++ b/pkg/providers/v1/aws_metrics.go
@@ -18,11 +18,12 @@ package aws
 
 import (
 	"context"
+	"errors"
 	"strconv"
 	"sync"
 
+	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
 	"github.com/aws/smithy-go/middleware"
-	smithyhttp "github.com/aws/smithy-go/transport/http"
 	"k8s.io/component-base/metrics"
 	"k8s.io/component-base/metrics/legacyregistry"
 )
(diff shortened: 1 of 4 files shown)


pkg/providers/v1/aws_metrics.go
```diff
@@ -110,3 +131,44 @@ func awsAPIMetricsMiddleware() middleware.DeserializeMiddleware {
 		},
 	)
 }
+
+// awsAPITerminalErrorMiddleware returns a Finalize middleware that records the
+// HTTP status code of AWS API calls that fail terminally. It is inserted before
+// the SDK's "Retry" middleware so it wraps the entire retry loop and runs once
+// per logical call, observing only the terminal error (not per attempt).
+func awsAPITerminalErrorMiddleware() middleware.FinalizeMiddleware {
```
yue9944882: do we have to add a new metrics? i think we can replace the existing one in-place. the original metrics contains too many transient errors that's recovered through retry.

snarkychef: I agree. I would prefer updating existing metric too. I wasn't sure if we should treat this as bug in the already added metric. 

I made the necessary changes to fix the existing metric instead of adding new one. Please take a look.

yue9944882 (commented):
/lgtm
/approve

PR #1480, opened 2026-08-19
Classify terminal AWS API throttling responses

howard-junec:
/kind bug

AWS services can report throttling through HTTP 429 or service-specific API error codes. Some throttling errors, such as EC2 `RequestLimitExceeded`, can use HTTP 503 and are therefore indistinguishable from service-side 5xx failures in `cloudprovider_aws_api_response_status_total`.

This PR adds a bounded `error_type` label while preserving the actual HTTP `status_code`. Terminal errors use `error_type="throttle"` when the response is HTTP 429 or the API error code is included in the AWS SDK's `retry.DefaultThrottleErrorCodes`; all other terminal HTTP errors use `error_type="other"`. Because the middleware wraps the SDK retry loop, errors recovered by a retry remain uncounted.

None.

- The implementation references the SDK's exported throttle-code set instead of maintaining a separate list.
- Keeping `status_code` unchanged preserves its HTTP semantics and allows existing consumers to retain their current behavior while producers and consumers roll out independently.
- This metric is ALPHA. `error_type` has exactly two values: `"throttle"` and `"other"`.

Validation:
- `go test ./pkg/providers/v1 -run 'TestAWSAPIMetricsMiddleware' -count=1`
- `go test ./pkg/providers/v1 -count=1`
- `go vet ./pkg/providers/v1`
- `go build ./...`
- `./hack/verify-gofmt.sh`

```release-note
The cloudprovider_aws_api_response_status_total metric now includes a bounded error_type label that classifies terminal AWS API responses as throttle or other while preserving the HTTP status code.
```

### DIFF
--- a/pkg/providers/v1/aws_metrics.go
+++ b/pkg/providers/v1/aws_metrics.go
@@ -19,10 +19,12 @@ package aws
 import (
 	"context"
 	"errors"
+	"net/http"
 	"strconv"
 	"sync"
 
 	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
+	"github.com/aws/smithy-go"
 	"github.com/aws/smithy-go/middleware"
 	"k8s.io/component-base/metrics"
 	"k8s.io/component-base/metrics/legacyregistry"
@@ -61,10 +63,10 @@ var (
 	awsAPIResponseStatusTotal = metrics.NewCounterVec(
 		&metrics.CounterOpts{
 			Name:           "cloudprovider_aws_api_response_status_total",
-			Help:           "AWS API error response status code counts (terminal, after retries are exhausted)",
+			Help:           "AWS API error response counts by status code and error type (terminal, after retries are exhausted)",
 			StabilityLevel: metrics.ALPHA,
 		},
-		[]string{"service", "operation", "status_code"})
+		[]string{"service", "operation", "status_code", "error_type"})
 )
 
 func recordAWSMetric(actionName string, timeTaken float64, err error) {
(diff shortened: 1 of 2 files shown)


pkg/providers/v1/aws_metrics.go
```diff
@@ -110,10 +112,19 @@ func awsAPIMetricsMiddleware() middleware.FinalizeMiddleware {
...
+					if errors.As(err, &apiErr) && apiErr.ErrorCode() == "RequestLimitExceeded" {
+						errorType = "throttle"
+					}
+				}
+
 				awsAPIResponseStatusTotal.With(metrics.Labels{
 					"service":     middleware.GetServiceID(ctx),
 					"operation":   middleware.GetOperationName(ctx),
 					"status_code": strconv.Itoa(respErr.HTTPStatusCode()),
+					"error_type":  errorType,
```
kmala, line 127: how will this explode the dimension of error type? also do we need to have a new dimension just for the throttle? can we just override status_code ?

howard-junec: So `error_type` is bounded to `""` and `"throttle"` (a hardcoded classification, not the raw AWS API error code). A service/operation only splits into two series when it emits both ordinary 503s and `RequestLimitExceeded` 503s, which is the same series split as using `status_code="throttle"`.

I kept a separate label because `status_code` currently represents the actual HTTP response code. Overriding it would make these responses disappear from existing `status_code="503"` queries, while the separate label preserves the current behavior until consumers are updated. My preference is to preserve the HTTP status, but I can switch if you prefer avoiding the schema change.

PR #1486, opened 2026-09-13
Manage multiple cluster-tagged security groups per node for LB ingress

### LINKED ISSUE #666: Support managing multiple tagged security groups for a node
Add support for managing multiple tagged security groups for a node.

**Why is this needed**:
At the moment we don't support finding and managing multiple tagged SecurityGroups per node (only one), and when we find more than one we error (ref code [here](https://github.com/kubernetes/cloud-provider-aws/blob/c9fbfe2bd46ab843655ec14f47021e4766a9c24b/pkg/providers/v1/aws.go#L4533)).

As such if a node has 2 or more tagged SecurityGroups we would fail to modify them in order to allow inbound traffic from the LB SecurityGroup.

Is there anything preventing us to extend this functionality to more than one tagged SecurityGroup?
If not, would it be possible to consider implementing it?
Thanks

/kind feature

KR-Ravindra:
**What type of PR is this?**

/kind feature

#### Problem

When a worker instance carries more than one security group tagged with the cluster tag, `findSecurityGroupForInstance` returns `Multiple tagged security groups found for instance ...` and the Service load balancer reconcile aborts. No ingress rule is opened on *any* node security group, so type=LoadBalancer traffic to those nodes is blocked. Both the classic ELB path (`updateInstanceSecurityGroupsForLoadBalancer`) and the NLB path (`updateInstanceSecurityGroupsForNLB`) go through this helper.

As discussed on the issue, there is no technical reason for the single-tagged-group restriction; it dates back to the original in-tree implementation.

#### Fix

- `findSecurityGroupForInstance` becomes `findSecurityGroupsForInstance` and returns every cluster-tagged security group attached to the instance (`[]ec2types.GroupIdentifier`).
- Back-compat behaviour is unchanged: with no tagged group, a single untagged group is still used, and multiple untagged groups is still an error.
- Both callers loop over the returned groups, so the load balancer ingress rule is authorized on each tagged node security group.

Fixes #666

#### How tested

Unit tests in `pkg/providers/v1/aws_test.go`:

- `TestUpdateInstanceSecurityGroupsForLoadBalancerMultipleTaggedGroups` drives `updateInstanceSecurityGroupsForLoadBalancer` with the mocked EC2 client for an instance that has two tagged groups and one untagged group, and asserts `AuthorizeSecurityGroupIngress` is issued for exactly the two tagged groups. Before the change it fails with:

  ```
  --- FAIL: TestUpdateInstanceSecurityGroupsForLoadBalancerMultipleTaggedGroups (0.10s)
      Error: Received unexpected error:
             Multiple tagged security groups found for instance i-multi-sg; ensure only the k8s security group is tagged; the tagged groups were sg-node-1(node-1) sg-node-2(node-2)
  ```

- `TestFindSecurityGroupsForInstance` (table) covers: single tagged, multiple tagged (all returned), tagged + untagged (only tagged), single untagged fallback, multiple untagged error, groups without id ignored, no groups.
- The existing `TestFindSecurityGroupForInstance` / `TestFindSecurityGroupForInstanceMultipleTagged` are updated to the new signature and new behaviour.

After the change: `go test ./pkg/providers/v1/ -run 'TestFindSecurityGroup|TestUpdateInstanceSecurityGroupsForLoadBalancerMultipleTaggedGroups'` passes; `gofmt`, `go vet` and `golangci-lint run --new-from-rev=upstream/master` report nothing.

```release-note
The cloud controller manager now opens load balancer ingress on every node security group carrying the cluster tag, instead of failing with "Multiple tagged security groups found" when a node has more than one tagged security group.
```

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra:
Round 1 self-review.

1. \`TestFindSecurityGroupForInstanceMultipleTagged\` now overlaps the "multiple tagged groups are all returned" case in the new table test. Fold it into the table and rename \`TestFindSecurityGroupForInstance\` to the plural form.
2. Description says \`golangci-lint\`; this repo runs \`golint\` via \`make verify-lint\`. Run \`make check\` and \`go test ./pkg/providers/v1/...\` and paste the result in the description.
3. Removal path checked against master: the CLB revoke diffs against \`buildSecurityGroupRuleReferences\` and LB deletion passes \`instances=nil\`; the NLB path iterates every cluster SG. All tagged node SGs are cleaned up symmetrically, so no extra code is needed there.
4. The single-untagged fallback and the multiple-untagged error are unchanged, matching the back-compat note in #666.

Staying draft until 1 and 2 are done.

ivomarino:
Traced the reconciliation path end to end against master, not just the
diff - the revoke side of `updateInstanceSecurityGroupsForLoadBalancer`
already keys `instanceSecurityGroupIds` by group ID rather than by
instance, so it diffs correctly against `actualGroups` no matter how
many security groups a given instance now returns. Confirms point 3 of
your self-review; no extra handling needed there.

One thing I didn't see addressed either in the diff or the self-review:
is there an e2e path that exercises the multi-tagged-SG case against
real AWS, not just the mocked unit tests? Reconciliation logic like
this tends to unit-test cleanly but still surprise you against real
`DescribeSecurityGroups` pagination/eventual consistency once a node
actually carries several cluster-tagged groups. Might be worth a quick
check before this comes out of draft, if there's an existing e2e suite
that covers load balancer security group handling.

KR-Ravindra:
No. tests/e2e/loadbalancer.go covers the load balancer's own security groups (managed and BYO NLB modes); nothing there gives a node a second cluster-tagged group, and the prow e2e cluster runs nodes with one tagged group, so the multi-tagged case is unit-tested only.

The AWS calls are unchanged by this PR: the tagged groups still come from one DescribeSecurityGroups in getTaggedSecurityGroups and the per-instance groups from DescribeInstances. The diff only changes which of the already-fetched groups are kept, so pagination and consistency behave as on master. I'll note the coverage gap in the description when I pick up the round-1 items.

ivomarino:
Thanks, that answers it cleanly - the diff only changes which
already-fetched groups are kept, not how they're fetched, so the
pagination/consistency question I raised doesn't actually apply here.
Makes sense that the multi-tagged case is unit-test-only for now given
the e2e cluster's own node setup.

